PR #4, opened 2016-03-31
use `.get_env()` method to get environment

minrk:
this will inherit .env, .environment logic from base spawner


yuvipanda:
I just did the same too :D

PR #7, opened 2016-06-03
Add explicit dependencies to packaging

elibixby:
Hi @yuvipanda 

Tested this with python 3.4 and at least import works after setup.py from a clean virtualenv when specifying these deps. Looks like it should be everything from the actual file...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,10 @@ from setuptools import setup
 setup(
     name='kubespawner',
     version='0.1',
+    install_requires=[
+        'requests-futures>=0.9.7',
+        'jupyterhub>=0.4.0',
+    ],
     description='JupyterHub Spawner targetting Kubernetes',
     url='http://github.com/yuvipanda/jupyterhub-kubernetes-spawner',
     author='Yuvi Panda',


yuvipanda:
Thank you very much for the patch!

elibixby:
@yuvipanda Quick question. I can't for the life of me figure out how to get the package installed into the correct python for Jupyterhub... running 

```
python setup.py install
```

using the same python as was used to install jupyterhub I still get a module not found error when Jupyterhub loads the config file. Have you gotten this working?

yuvipanda:
I have. Can you paste me your config file? ALso does 'pip freeze' list all
the modules?

On Thu, Jun 2, 2016 at 10:24 PM, elibixby <notifications@github.com> wrote:

> @yuvipanda <https://github.com/yuvipanda> Quick question. I can't for the
> life of me figure out how to get the package installed into the correct
> python for Jupyterhub... running
>
> python setup.py install
>
> using the same python as was used to install jupyterhub I still get a
> module not found error when Jupyterhub loads the config file. Have you
> gotten this working?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/yuvipanda/jupyterhub-kubernetes-spawner/pull/7#issuecomment-223493449>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB23ufOV9XWpH3mf9JEeEwSuwgmrc39ks5qH7qLgaJpZM4ItGYK>
> .
>



-- 
Yuvi Panda T
http://yuvi.in/blog

elibixby:
Nevermind, got it working I think.

PR #11, opened 2016-11-11
Volumes

SaladRaider:
Summary
----------------
Supports dynamically provisioning pvc and pv with storage class for single users pods. Users can now have persistent storage on their single user notebooks.

How To Test
----------------
1. Add a StorageClass to the manifest.yaml, maybe one named eg. `gce-standard-storage`
2. Configure the following variables in jupyterhub_config.py, for example:
```
c.KubeSpawner.pvc_name_template = 'claim-{username}-{userid}'
c.KubeSpawner.storage_class = 'gce-standard-storage'
c.KubeSpawner.access_modes = ['ReadWriteOnce']
c.KubeSpawner.storage = '10Gi'

c.KubeSpawner.volumes = [
    {
        'name': 'volume-{username}-{userid}',
        'persistentVolumeClaim': {
            'claimName': 'claim-{username}-{userid}'
        }
    }
]
    {
        'mountPath': '/home',
        'name': 'volume-{username}-{userid}'
    }
]
```
3. Build the new jupyterhub image and apply the new manifest.
4. Log into jupyterhub as an arbitrary user, say `peter`
5. Start `peter`'s single user server
6. Create a file in `home`, or wherever you mount the volume
7. Go to control panel and shut down the server. Ensure that the pod is terminated by `watch kubectl get pods`
8. Start the server again and navigate to `home` to see if the file is still there.
9. Repeat this step for a different user starting at step 4.
10. Repeat steps 4-8 with yet another user but while both of the other users' pods are down, so nothing is mounted to their corresponding volume claims.


kubespawner/spawner.py
```diff
@@ -274,6 +291,32 @@ def _hub_connect_port_default(self):
         """
     )
 
+    storage = Unicode(
```
yuvipanda: Call it user_storage_capacity for clarity?

Also let's default this to None (and set `allow_none=True` to make that possible), and not do anything if it is None?

kubespawner/spawner.py
```diff
@@ -274,6 +291,32 @@ def _hub_connect_port_default(self):
...
+        "1Gi",
+        config=True,
+        help="""
+        The ammount of storage space to request from the volume that the pvc will
+        mount to.
+        """
+    )
+
+    storage_class = Unicode(
```
yuvipanda: user_storage_class? Same note re: None. In general use None to mean 'ignore this setting' rather than reusing ''.

kubespawner/spawner.py
```diff
@@ -274,6 +291,32 @@ def _hub_connect_port_default(self):
...
+        "",
+        config=True,
+        help="""
+        The storage class that the pvc will use. If left blank, the kubespawner will not
+        create a pvc for the pod.
+        """
+    )
+
+    access_modes = List(
```
yuvipanda: user_storage_access_modes? Also perhaps use a sane default here (RWX?) so that people don't have to touch it if they don't have to? 

In general, either properties should be required, in which case set their default to None and refuse to work if they aren't set, or be optional and have sane defaults. So this one looks like it should have sane defaults.

yuvipanda: The docs should also be self explanatory, and define the options. Can be just a short blurb + a link to the k8s docs. Can you expand help= for all fo the three properties?

kubespawner/spawner.py
```diff
@@ -291,29 +291,68 @@ def _hub_connect_port_default(self):
...
-        "1Gi",
+    user_storage_capacity = Unicode(
+        None,
         config=True,
+        allow_none=True,
         help="""
         The ammount of storage space to request from the volume that the pvc will
-        mount to.
+        mount to. This ammount will be the ammount of storage space the user has
+        to work with on their notebook. If left blank, the kubespawner will not
```
yuvipanda: Should be 'If set to None'

kubespawner/spawner.py
```diff
@@ -291,29 +291,68 @@ def _hub_connect_port_default(self):
...
         help="""
-        List of access modes for pvc.
+        List of access modes the user has for the pvc.
+
+        The access modes are:
+            The access modes are:
+                ReadWriteOnce – the volume can be mounted as read-write by a single node
+                ReadOnlyMany – the volume can be mounted read-only by many nodes
+                ReadWriteMany – the volume can be mounted as read-write by many nodes
+            In the CLI, the access modes are abbreviated to:
```
yuvipanda: This isn't relevant, since we aren't using the CLI here.

kubespawner/spawner.py
```diff
@@ -472,7 +511,7 @@ def poll(self):
 
     @gen.coroutine
     def start(self):
-        if self.storage_class:
+        if self.user_storage_class && self.user_storage_capacity:
```
yuvipanda: 'and' not && :) Should also explicitly use a 'is not None' check instead. And an or instead of and probably.

SaladRaider: yuvi, if we use `or` instead of `and`, like `self.user_storage_class is not None or self.user_storage_capacity is not None`, the pvc will be created if either the storage class or storage capacity is set. But if one of them is set, and the other is not, then the pvc will still try to be created. Is this what we want?

yuvipanda: Damn, you're totally right, it should be and! Sorry for the confusion

yuvipanda:
I'm going to merge this now, and will try to spend some time tomorrow or day after testing some more. thanks, and congrats on finishing it up :)

21stio:
That's how to configure it currently
```
c.KubeSpawner.pvc_name_template = 'jupyterhub-notebook-{username}'
c.KubeSpawner.user_storage_pvc_ensure = True #creates persistent volume claim
c.KubeSpawner.user_storage_capacity = '10Gi'
c.KubeSpawner.user_storage_access_modes = ['ReadWriteOnce']
c.KubeSpawner.user_storage_class = 'staging'
```

PR #12, opened 2016-11-14
Correct some links and reflow README text

willingc:
@yuvipanda This is largely a clean-up PR. I may take a swipe at sphinx docs soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,49 @@
-# jupyterhub-kubernetes-spawner #
+# jupyterhub-kubernetes-spawner (kubespawner)#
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
Awesome! TYVM :)

I've gone back and forth between manual linebreaks and not, and got lazy and used non-manual linebreaks here - mostly because it made editing after the fact easier. Are there tools that do automatic reflow?

willingc:
You are too quick. No worries on line breaks. Easy enough to reflow. Atom reflows. Though a bit wonky with bullets. Going to send another small change.

PR #22, opened 2017-02-16
Delete pod if it already exists before attempting to start it again

yuvipanda:
This will kill old pods that JupyterHub has lost track of or haven't started in time, and kill them before attempting to start a new pod.


yuvipanda:
Tested locally, seems to work!

I tested this by creating a bogus pod that has a name that JupyterHub will want to use, but is stuck in a non-running state (easiest way is to use a false image name, which gets it stuck in ErrImagePull). Then try starting the pod - it should kill the bogus existing pod and start a new one.

PR #27, opened 2017-02-20
Make singleuser_uid more configurable

### LINKED ISSUE #26: Allow setting run_as_uid and fs_gid from functions
Right now, they have to be the same static string for all users. Instead, allow it to be a callable that takes a particular set of parameters!

yuvipanda:
Fixes #26 

With this, and a theoretical authenticator that sets user.id to be a POSIX uid of the user, you can use the following line in your jupyterhub_config.py to run users as their uids (important for NFS file access, mostly):

```python
c.KubeSpawner.singleuser_uid = lambda spawner: spawner.user.id
```

The callable can also be a coroutine, so you can actually perform a remote call (to LDAP or whatever) to find the uid as which the user should be running, and just return that.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -11,10 +11,10 @@ from urllib.parse import urlparse, urlunparse
 
 from tornado import gen
 from tornado.httpclient import AsyncHTTPClient, HTTPError
-from traitlets import Unicode, List, Integer, Float
+from traitlets import Unicode, List, Integer, Union
 from jupyterhub.spawner import Spawner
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
@minrk is fine with exposing spawner instances to jupyterhub_config like this, so merging!

antigenius0910:
Sorry @yuvipanda can you tell me where does your request merged to? I am having the same problem need to setup c.KubeSpawner.singleuser_uid as LDAP uid for NFS file access as well. Trying to figure out here how do I achieve this goal. Thanks

PR #34, opened 2017-03-19
Allow setting c.Spawner.cmd to whatever the user wants

### LINKED ISSUE #31: Specify container command
Is it possible to specify the container command via config? To use the docker stacks, command must be set to `/usr/local/bin/singleuser.sh`, but I can't see how to do it here. In DockerSpawner, it's passed to `docker.create`

yuvipanda:
Fixes #31 

Need to verify that this doesn't cause any breaking changes (other than the env vars now gone)


yuvipanda:
Seems ok on local testing!

PR #39, opened 2017-04-09
Change default docker image to jupyterhub/singleuser

bskaggs:
The default "[jupyter/singleuser](https://hub.docker.com/r/jupyter/singleuser/)" image moved last year to "[jupyterhub/singleuser](https://hub.docker.com/r/jupyterhub/singleuser/)", and is no longer being updated at the old name currently used by kubespawner. This pull request also updates the default cmd to `None`, ensuring that the container's default command is used by default.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -87,7 +87,7 @@ class KubeSpawner(Spawner):
     ).tag(config=True)
 
     cmd = Command(
-        ['jupyterhub-singleuser'],
+        None,
         allow_none=True,
         minlen=0,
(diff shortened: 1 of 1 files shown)


willingc:
Thanks @bskaggs.

@yuvipanda Just an FYI I opened a [PR in JupyterHub docs](https://github.com/jupyterhub/jupyterhub/pull/1072) about where the different Docker images are located on DockerHub. This kubespawner looks good to me but you're the expert on this spawner so I'll leave the formal review to you. :-)

PR #40, opened 2017-04-13
Add instructions for MacOS

danielfrg:
Added a couple of comments on how to run the dev environment this successfully on MacOS/OS X.

Added one small trait to be able to set the `asynchttpclient` type since there is a known issue with Curl on MacOS with minikube certificate type.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -36,8 +36,11 @@ at this time, but should work for OS X too.
 7. Now the containers running on kubernetes can connect to your host, via the IP address for `vboxnet4`
    interface. You can find this IP with: 
    ```
+   LINUX:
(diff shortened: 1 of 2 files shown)


yuvipanda:
w00t, ty for the patch :)

danielfrg:
I didn't knew all that was possible so thank you!

I might have some time to work on this and I was thinking on updating some of the code to use the now more official kubernetes python client (https://github.com/kubernetes-incubator/client-python) is there some interest on that?

willingc:
Nice contribution @danielfrg. There's definitely interest in contributors helping out with kubernetes and JupyterHub. I would ping @yuvipanda, myself, or @minrk on the jupyterhub gitter channel or open an issue with any suggestions. Thanks :cake:

Issue #6, opened 2016-05-23
Allow specifying node-selectors for pods


yuvipanda:
To support heterogenous kubernetes clusters.

Issue #1, opened 2016-03-08
Add a default health check for the pods


yuvipanda:
So that they get properly killed when the underlying hosts die.

consideRatio:
I'm closing this as a bit too old to care about at this point. If we want this still, we should document it to be more concrete.

PR #41, opened 2017-04-14
Use official Kubernetes python client

danielfrg:
This PR changes the `objects.py`to use the official [python kubernetes client](https://github.com/kubernetes-incubator/client-python). Hopefully this makes k8s specs creation a little easier by adding a pure python dependency.

This also solves https://github.com/jupyterhub/kubespawner/issues/6 by adding a simple `nodeSelector` to the Pod spec.

I also tried to solve https://github.com/jupyterhub/kubespawner/issues/1 but using `/user/{userid}` as a path for liveliness return a 404 because of Hub redirects or something (not 100% sure) any other ideas on how to check status of the Notebook server?

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -13,8 +13,6 @@ from kubernetes.client.models.v1_container import V1Container
 from kubernetes.client.models.v1_container_port import V1ContainerPort
 from kubernetes.client.models.v1_env_var import V1EnvVar
 from kubernetes.client.models.v1_resource_requirements import V1ResourceRequirements
-from kubernetes.client.models.v1_volume import V1Volume
-from kubernetes.client.models.v1_volume_mount import V1VolumeMount
 
 from kubernetes.client.models.v1_persistent_volume_claim import V1PersistentVolumeClaim
 from kubernetes.client.models.v1_persistent_volume_claim_spec import V1PersistentVolumeClaimSpec
@@ -124,7 +122,6 @@ def make_pod_spec(
         pod.spec.image_pull_secrets.append(image_secret)
 
     pod.spec.containers = []
-
     notebook_container = V1Container()
     notebook_container.name = "notebook"
     notebook_container.image = image_spec
@@ -140,30 +137,10 @@ def make_pod_spec(
     notebook_container.resources = V1ResourceRequirements()
     notebook_container.resources.requests = {"cpu": cpu_guarantee, "memory": mem_guarantee}
     notebook_container.resources.limits = {"cpu": cpu_limit, "memory": mem_limit}
-
-    # Add a hack to ensure that no service accounts are mounted in spawned pods
-    # This makes sure that we don"t accidentally give access to the whole
-    # kubernetes API to the users in the spawned pods.
-    # See https://github.com/kubernetes/kubernetes/issues/16779#issuecomment-157460294
-    hack_volume = V1Volume()
-    hack_volume.name =  "no-api-access-please"
-    hack_volume.empty_dir = {}
-
-    hack_volume_mount = V1VolumeMount()
-    hack_volume_mount.name = "no-api-access-please"
-    hack_volume_mount.mount_path = "/var/run/secrets/kubernetes.io/serviceaccount"
-    hack_volume_mount.read_only = True
-
-    notebook_container.volume_mounts = [hack_volume_mount]
-    for volume_mount in volume_mounts:
-        notebook_container.volume_mounts.append(volume_mount)
-
+    notebook_container.volume_mounts = volume_mounts
     pod.spec.containers.append(notebook_container)
 
-    pod.spec.volumes = [hack_volume]
-    for volume in volumes:
-        pod.spec.volumes.append(volume)
-
+    pod.spec.volumes = volumes
     return api_client.sanitize_for_serialization(pod)
 
 
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -15,8 +15,6 @@ from kubernetes.client.models.v1_env_var import V1EnvVar
 from kubernetes.client.models.v1_resource_requirements import V1ResourceRequirements
 from kubernetes.client.models.v1_volume import V1Volume
 from kubernetes.client.models.v1_volume_mount import V1VolumeMount
-# from kubernetes.client.models.v1_probe import V1Probe
-# from kubernetes.client.models.v1_http_get_action import V1HTTPGetAction
 
 from kubernetes.client.models.v1_persistent_volume_claim import V1PersistentVolumeClaim
 from kubernetes.client.models.v1_persistent_volume_claim_spec import V1PersistentVolumeClaimSpec
@@ -33,10 +31,9 @@ def make_pod_spec(
     run_as_uid,
     fs_gid,
     env,
-    node_selector,
+    working_dir,
     volumes,
     volume_mounts,
-    working_dir,
     labels,
     cpu_limit,
     cpu_guarantee,
@@ -75,8 +72,6 @@ def make_pod_spec(
         read / write to the volumes mounted.
       - env:
         Dictionary of environment variables.
-      - node_selector:
-        Dictionary to match nodes where to launch the Pods
       - volumes:
         List of dictionaries containing the volumes of various types this pod
         will be using. See k8s documentation about volumes on how to specify
@@ -111,12 +106,8 @@ def make_pod_spec(
 
     pod.metadata = V1ObjectMeta()
     pod.metadata.name = name
-    pod.metadata.labels = {}
-    pod.metadata.labels.update({"name": name})
-    for key, value in labels.items():
-        pod.metadata.labels.update({key: value})
+    pod.metadata.labels = labels.copy()
     
-
     pod.spec = V1PodSpec()
 
     security_context = V1PodSecurityContext()
@@ -131,9 +122,6 @@ def make_pod_spec(
         image_secret = V1LocalObjectReference()
         image_secret.name = image_pull_secret
         pod.spec.image_pull_secrets.append(image_secret)
-      
-    if node_selector:
-        pod.spec.node_selector = node_selector
 
     pod.spec.containers = []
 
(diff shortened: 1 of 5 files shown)


yuvipanda:
Thank you for working on this! This looks great, am reviewing it now.

kubespawner/objects.py
```diff
@@ -75,63 +103,98 @@ def make_pod_spec(
...
+    api_client = ApiClient()
+
+    pod = V1Pod()
+    pod.kind = "Pod"
+    pod.api_version = "v1"
+
+    pod.metadata = V1ObjectMeta()
+    pod.metadata.name = name
+    pod.metadata.labels = {}
+    pod.metadata.labels.update({"name": name})
```
yuvipanda: why not just have this be one line?

kubespawner/objects.py
```diff
@@ -11,13 +33,15 @@ def make_pod_spec(
     run_as_uid,
     fs_gid,
     env,
+    node_selector,
```
yuvipanda: Can you split out the additional features into separate commits (or even a separate PR) so it's easier to review? I'll leave commnets about them here too tho

kubespawner/objects.py
yuvipanda:
```python
pod.metadata.labels = labels.copy()
pod.metadata.labels['name'] = name
```

seems clearer and possibly faster?

kubespawner/objects.py
```diff
@@ -75,63 +103,98 @@ def make_pod_spec(
...
+        env_ = V1EnvVar()
+        env_.name = key
+        env_.value = value
+        notebook_container.env.append(env_)
+    notebook_container.args = cmd
+    notebook_container.image_pull_policy = image_pull_policy
+    notebook_container.resources = V1ResourceRequirements()
+    notebook_container.resources.requests = {"cpu": cpu_guarantee, "memory": mem_guarantee}
+    notebook_container.resources.limits = {"cpu": cpu_limit, "memory": mem_limit}
+    # notebook_container.liveness_probe = V1Probe()
```
yuvipanda: Let's get the commented code out? Also would be nice if it were a separate commit :)

kubespawner/spawner.py
```diff
@@ -109,6 +108,26 @@ def _namespace_default(self):
...
+
+        Default is None and means it will be launched in any available Node.
+        
+        For example to match the Nodes that have a label of `disktype: ssd` use:
+            `{"disktype": "ssd"}`
+        """
+    )
+
+    working_dir = Unicode(
+        "/home",
```
yuvipanda: I think the default should be None, which should make it use whatever is specified in the Docker image (which is the current default). Should probably be called singleuser_working_dir as well (I've been trying to do that for all new properties)

kubespawner/spawner.py
```diff
@@ -109,6 +108,26 @@ def _namespace_default(self):
         """
     ).tag(config=True)
 
+
+    node_selector = Dict(
```
yuvipanda: Can this be a separate commit / PR? We need to be careful to support both simple property based selectors and more complex ones...

kubespawner/objects.py
danielfrg: Yes, thats better. Done

kubespawner/objects.py
danielfrg: I have removed the `node_selector` things I will make another PR for that.

kubespawner/objects.py
danielfrg: Yes, i wanted to have it there temporally in case you wanted it on this PR. It's gone.

kubespawner/spawner.py
danielfrg: Ah yes, I agree that its a better default. Also renamed the property to `singleuser_working_dir`.

danielfrg:
I have fixed the tests and removed the `node_selector` logic and I will make a separate PR for that Issue.

You can probably Squash and Merge this one if you want to keep history clean about changes.

kubespawner/spawner.py
```diff
@@ -109,6 +108,13 @@ def _namespace_default(self):
         """
     ).tag(config=True)
 
+    singleuser_working_dir = Unicode(
+        help="""
```
yuvipanda: set the default explicitly to None and also allow_none?

yuvipanda:
(I'll take a more detailed look + test on Monday!)

kubespawner/spawner.py
danielfrg: Done

kubespawner/objects.py
```diff
@@ -75,63 +98,86 @@ def make_pod_spec(
...
+    notebook_container.image = image_spec
+    notebook_container.working_dir = working_dir
+    notebook_container.ports = []
+    port_ = V1ContainerPort()
+    port_.name = "notebook-port"
+    port_.container_port = port
+    notebook_container.ports.append(port_)
+    notebook_container.env = []
+    for key, value in env.items():
+        env_ = V1EnvVar()
```
yuvipanda: This could just be:

```python
notebook_container.env = [V1EnvVar(k, v) for k, v in env.items()]
```

no? https://github.com/kubernetes-incubator/client-python/blob/84bf6cc9dc206cb3314a5b4d2fa1a9c2fa2c5137/kubernetes/client/models/v1_env_var.py#L24 says it takes the properties as params to init

kubespawner/spawner.py
```diff
@@ -433,6 +442,15 @@ def _hub_connect_port_default(self):
         """
     )
 
+    user_storage_match_labels = Dict(
+        help="""
```
yuvipanda: What's the default for this? I'm trying to understand what exactly this feature does - does it allow binding to an existing PVC that isn't created by this? Under what circumstances would I use this?

kubespawner/spawner.py
```diff
@@ -469,19 +487,6 @@ def get_pod_manifest(self):
         """
         Make a pod manifest that will spawn current user's notebook pod.
         """
-        # Add a hack to ensure that no service accounts are mounted in spawned pods
-        # This makes sure that we don't accidentally give access to the whole
-        # kubernetes API to the users in the spawned pods.
-        # See https://github.com/kubernetes/kubernetes/issues/16779#issuecomment-157460294
-        hack_volumes = [{
```
yuvipanda: I think this belongs here, and *not* in the make_pod_spec. We could add an option to allow service account access in here when it's here, and this also keeps the make_pod_spec agnostic about what we do. Someone has actually forked this to make this disabling optional, and that's easier to do when this is here.

kubespawner/spawner.py
```diff
@@ -496,38 +501,39 @@ def get_pod_manifest(self):
...
-            self.mem_guarantee,
+            name=self.pod_name,
+            image_spec=self.singleuser_image_spec,
+            image_pull_policy=self.singleuser_image_pull_policy,
+            image_pull_secret=self.singleuser_image_pull_secrets,
+            port=self.port,
+            cmd=real_cmd,
+            run_as_uid=singleuser_uid,
+            fs_gid=singleuser_fs_gid,
+            env=self.get_env(),
```
yuvipanda: Passing named args is much cleaner, thank you for moving it!

kubespawner/objects.py
```diff
@@ -75,63 +98,86 @@ def make_pod_spec(
...
-        }
-    }
+        pod.spec.image_pull_secrets = []
+        image_secret = V1LocalObjectReference()
+        image_secret.name = image_pull_secret
+        pod.spec.image_pull_secrets.append(image_secret)
+
+    pod.spec.containers = []
+
+    notebook_container = V1Container()
```
yuvipanda: We could also construct notebook_container this way (passing params to __init__) since that supports it too (https://github.com/kubernetes-incubator/client-python/blob/84bf6cc9dc206cb3314a5b4d2fa1a9c2fa2c5137/kubernetes/client/models/v1_container.py). I'm not super attached to it here, though - because there's a lot more params.

kubespawner/spawner.py
danielfrg: So by default its nothing so the default behavior should be the same.

No, it still creates a PVC spec as it was doing before. This new property basically it just passes the `matchLabels` field from the PVC spec so its possible to filter between existing PV. Lets say you have multiple PV you can select which one you want.

From: https://kubernetes.io/docs/concepts/storage/persistent-volumes/ - PersistentVolumeClaims section.

yuvipanda: Ah, so is it for the case where the PV is not dynamically created by using a StorageClass?

kubespawner/objects.py
danielfrg: Yes I agree, personally I don't like to pass 3+ params to a function.

kubespawner/spawner.py
danielfrg: 👍

kubespawner/objects.py
danielfrg: Yes, thats good. Done

danielfrg:
This changes I have only tested on minikube but I still yet to see something in k8s that doesn't work from minukube to GCP.

kubespawner/spawner.py
danielfrg: Yes I agree with keeping `make_pod_spec` independent of the hack i just wanted to keep the `kubernetes_client` code in one place. I put it back now.

kubespawner/spawner.py
yuvipanda: ok! I think  we should set the default explicitly, and explain why you should use this and when in the help. The current help text isn't sufficient, IMO.

Also in the future, it'd be nice to have these as a separate PR, rather than as part of one :) Makes it easier to review.

I'll do some testing and try to get this merged tomorrow!

yuvipanda: The more I think about it, the more I think we should take it out now. I think way to include this would be:

1. Write a 'LabelSelector' trait type, that can be used to express either matchLabels or matchExpressions
2. Use that for this and for nodeSelectors

I think that's the cleanest and most future compatible way of doing this. what do you think?

yuvipanda: It should probably also support expanding {username} and friends in the label values, to provide more targeted matching. Let's definitely get it out of this PR?

danielfrg: Ok, its gone now. Yes, it was for that case where you already had existing PVs. This is the way I have always used PV + PVC, more manual than the automatic creation.

yuvipanda:
Thank you for the patch!

yuvipanda:
I'm going to test this now with https://github.com/jupyterhub/helm-chart

yuvipanda:
I filed https://github.com/jupyterhub/kubespawner/issues/43 with further options on how to move forward getting rid of our custom code in favor of the official client.

PR #44, opened 2017-04-24
Add option to specify nodeSelector of the Pods

danielfrg:
This spins off from: https://github.com/jupyterhub/kubespawner/pull/41 and fixes https://github.com/jupyterhub/kubespawner/issues/6

Based on feedback there I added a new traitlet `LabelSelector` that can be reused to match multiple type of selectors in Kubernetes. For this PR its used to specify the `node_selector` in the Pod spec but it can be reused to match other things like volumes in a PVC.

I moved the `Callable` trailet to the `trailets.py` similar to what JupyterHub core does.


danielfrg:
@yuvipanda could you take a quick look to this one?

ktong:
Also need this feature:)

yuvipanda:
Oops! I'll take a close look tomorrow!

yuvipanda:
ok, I've changed this up a little (undoing one of my earlier suggestions!) and am going to merge it now! Thank you for the patch, and so sorry for the delayed merging and reviewing!

PR #47, opened 2017-05-08
Add node selector

opskumu:
In some cases, the pod need to be eligible to run on a node with nodeSelector option.


opskumu:
duplicated https://github.com/jupyterhub/kubespawner/pull/44

PR #48, opened 2017-05-08
Support lifecycle hooks when spawning single-user Notebook server.

ktong:
Lifecycle hook can be used to preparing or cleanup some resources (e.g. OAuth token, service accounts).

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -36,6 +36,7 @@ def make_pod_spec(
     cpu_guarantee,
     mem_limit,
     mem_guarantee,
+    lifecycle_hooks,
 ):
     """
     Make a k8s pod specification for running a user notebook.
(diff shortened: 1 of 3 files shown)


kubespawner/spawner.py
```diff
@@ -444,6 +444,31 @@ def _hub_connect_port_default(self):
         """
     )
 
+    lifecycle = Dict(
```
yuvipanda: Can you rename this to singleuser_lifecycle_hooks so it's clearer what it is?

kubespawner/objects.py
```diff
@@ -36,6 +36,7 @@ def make_pod_spec(
     cpu_guarantee,
     mem_limit,
     mem_guarantee,
+    lifecycle,
```
yuvipanda: rename this to lifecycle_hooks too for clarity?

ktong:
Names changed:)

yuvipanda:
w00t! thank you for the patch :)

PR #49, opened 2017-05-16
support initContainers

ktong:
To enable some configuration which notebook user has no permission. One example is disable accessing to metadata service from notebook server.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -37,6 +37,7 @@ def make_pod_spec(
     mem_limit,
     mem_guarantee,
     lifecycle_hooks,
+    init_containers,
 ):
     """
     Make a k8s pod specification for running a user notebook.
@@ -97,6 +98,8 @@ def make_pod_spec(
         are allowed
       - lifecycle_hooks:
         Dictionary of lifecycle hooks
+      - init_containers:
+        List of initialization containers belonging to the pod.
     """
     api_client = ApiClient()
 
@@ -154,6 +157,7 @@ def make_pod_spec(
     notebook_container.volume_mounts = volume_mounts
     pod.spec.containers.append(notebook_container)
 
+    pod.spec.init_containers = init_containers
     pod.spec.volumes = volumes
     return api_client.sanitize_for_serialization(pod)
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -469,6 +469,33 @@ class KubeSpawner(Spawner):
         """
     )
 
+    singleuser_init_containers = List(
+        None,
+        config=True,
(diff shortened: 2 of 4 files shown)


kubespawner/spawner.py
```diff
@@ -469,6 +469,33 @@ def _hub_connect_port_default(self):
...
+          command: ["iptables", "-A", "OUTPUT", "-p", "tcp", "--dport", "80", "-d", "169.254.169.254", "-j", "DROP"]
+          securityContext:
+            capabilities:
+              add:
+              - NET_ADMIN         
+
+        See https://kubernetes.io/docs/concepts/workloads/pods/init-containers/ for more
+        info on what init containers are and why you might want to use them!
+        
+        To user this feature, Kubernetes version must greate than 1.6.
```
yuvipanda: must be greater

ktong:
changed. 
There are two changes need mentioned here:
1. kubeneters upgrade to 2.*
2. securityContext move from pod to container, otherwise initContainer can not find the right user to run.

kubespawner/objects.py
```diff
@@ -110,13 +113,6 @@ def make_pod_spec(
 
     pod.spec = V1PodSpec()
 
-    security_context = V1PodSecurityContext()
```
yuvipanda: Can you leave this here too? No harm in having them both, and I'd rather err on the side of overspecifying than underspecifying in this context.

ktong: Do you mean attach a empty V1PodSecurityContext() to pod.spec?

For fs_group and run_as_user in security, it's better to set in container otherwise initContainers will be failed due to no run_as_user found

ktong:
add blank security context into pod

kubespawner/objects.py
yuvipanda: ooooh, I see what you're saying...

Can init containers have their own security contexts?

yuvipanda: Apparently they can - I'd recommend we leave the default on, and have people set an explicit securityContext on the init container if it's using an userID that isn't available in the pod (or vice versa)

ktong:
Yes, every container in containers and initContainers can have their own securityContext

yuvipanda:
Ok in that case I suggest:

1. pod securityContext matches the notebook one
2. People can explicitly set a securityContext for initContainers if they want.

Right now, it looks like initcontainers run as root by default, which I don't think should happen. 

How does that sound?

ktong:
will try

ktong:
changed based on your suggestion.

yuvipanda:
I've rebased and merged this! thank you for the patch!

PR #53, opened 2017-05-31
add KubeTemplateSpawner 

ktong:
To support more flexible configuration for single-user notebook even though configuration does not support in KubeSpawner yet.

### DIFF
--- a/kubespawner/__init__.py
+++ b/kubespawner/__init__.py
@@ -13,6 +13,6 @@ We export KubeSpawner specifically here. This simplifies import for users.
 Users can simply import kubespawner.KubeSpawner in their applications
 instead of the more verbose import kubespawner.spawner.KubeSpawner.
 """
-from kubespawner.spawner import KubeSpawner
+from kubespawner.spawner import KubeSpawner, KubeTemplateSpawner
 
-__all__ = [KubeSpawner]
+__all__ = [KubeSpawner, KubeTemplateSpawner]
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -22,9 +22,9 @@ from kubespawner.utils import request_maker, k8s_url
 from kubespawner.objects import make_pod_spec, make_pvc_spec
 
 
-class KubeSpawner(Spawner):
+class KubeBaseSpawner(Spawner):
     """
-    Implement a JupyterHub spawner to spawn pods in a Kubernetes Cluster.
+    Implementation base of a JupyterHub spawner to spawn pods in a Kubernetes Cluster.
     """
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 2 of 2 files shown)


yuvipanda:
Heya! This is great :D But does it need to be a separate spawner? Can you just not add the hook to KubeSpawner itself?

ktong:
First, KubeSpawner could not cover all configurations supported by K8s. Moreover, K8s adds more configuration in each release. Template Spawner enables using any configurations supported by K8s without waiting for change in KubeSpawner.

For example, I used envFrom from 1.6 to set environment variables in notebook pod. It could be set in jupyterhub and pass it to notebook via env_keep. But it's more straightforward by setting in notebook directly.

ktong:
Moreover, writting yaml template is much easy and straightforward than set individual parameters.

yuvipanda:
You can add extra python code in your jupyterhub_config.py file, including setting up a function that can override things in the spawner and customize it directly. So in your jupyterhub_config.py file, you can have:


```python
from kubespawner import KubeSpawner

class CustomSpawner(KubeSpawner):
    def augment_pod(self, pod):
          # Modify pod object
          return pod

c.JupyterHub.spawner_class = CustomSpawner
```

yuvipanda:
This is going to be much more extensible and simpler than using a template language.

ktong:
It needs change docker image, which is separated with K8s yaml file. By the way, below is  jupyterhub_config.py I am using:
```
import os
import yaml


def _load_from_env(prefix):
    """ Load configuration from environment variables which start with prefix.

    :param prefix: The prefix to match name of environment variables
    :type prefix: str
    :return: Environment variables which name starts with prefix
    :rtype: dict
    """
    prefix_upper = prefix.upper()
    prefix_length = len(prefix) + 1
    return {k[prefix_length:].lower(): yaml.safe_load(v)
           for k, v in os.environ.items()
           if k.upper().startswith(prefix_upper + "_")}


def _load_config(config):
    config_string = config if isinstance(config, str) else config.get_value(None)
    prefix = _get_prefix(config_string) if config_string else ""

    if not prefix:
        return

    env = _load_from_env(prefix)
    if env:
        c[prefix].update(env)


def _get_prefix(config):
    try:
        return config[config.rindex('.') + 1:]
    except ValueError:
        return config


# Load config from environment variables
for config in ["Application", "JupyterHub"]:
    _load_config(config)
_load_config(c.JupyterHub.authenticator_class)
_load_config(c.JupyterHub.spawner_class)

# Add service to monitor and cull idle single-user servers
cull_env = _load_from_env("CULL")
if cull_env.get("enabled", True):
    cull_timeout = cull_env.get("timeout", 3600)
    cull_every = cull_env.get("every", 600)
    c.JupyterHub.services.append(
        {
            'name': 'cull-idle',
            'admin': True,
            'command': 'python3 cull_idle_servers.py --timeout={} --cull_every={}'
                .format(cull_timeout, cull_every).split()
        }
    )
```

With jupyterhub_config.py, K8s is able to customize deployment of docker image with envFrom

ktong:
Because we need write yaml for either K8s or helm, writing template is more natural.

ktong:
replaced by #91

PR #58, opened 2017-06-13
Tickets/dm 10913

athornton:


yuvipanda:
Note that https://github.com/jupyterhub/jupyterhub/issues/1117 should help with getting github tokens and other stuff into user containers.

PR #63, opened 2017-07-09
Use reflector pattern to speed up polls

yuvipanda:
We need to keep state within JupyterHub about the current running
status of all pods. We were currently doing this by making one
API request per pod every `poll_interval` - this is quite expensive,
and makes jupyterhub's polling scale linearly with number of
total users on the hub. Ideally we want this to be as close to
constant time as possible.

With this commit, we'll instead start an extra thread that'll
perform a 'list and watch' operation on the types of resources
we are interested in, and make sure that it is kept up to date
as atomically as possible (although there might be some lag sometime).
This makes poll be a simple synchronous dictionary access, rather
than a much more complex network poll.

This should make everything much faster - polling was one of the
biggest consumers of network time without this patch, and that
should significantly reduce with this patch.

We also take this opportunity to encode better the conditions
in which the pod is no longer considered running - particularly
when it is in 'Terminating' stage. This was a little bit flaky
before, but is ok now!

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -166,6 +166,15 @@ class PodReflector(SingletonConfigurable):
         self.watch_thread.daemon = True
         self.watch_thread.start()
 
+class SingletonExecutor(SingletonConfigurable, ThreadPoolExecutor):
+    """
+    Simple wrapper to ThreadPoolExecutor that is also a singleton.
+
+    We want one ThreadPool that is used by all the spawners, rather
+    than one ThreadPool per spawner!
+    """
+    pass
+
 class KubeSpawner(Spawner):
     """
     Implement a JupyterHub spawner to spawn pods in a Kubernetes Cluster.
@@ -174,7 +183,7 @@ class KubeSpawner(Spawner):
         super().__init__(*args, **kwargs)
         # By now, all the traitlets have been set, so we can use them to compute
         # other attributes
-        self.executor = ThreadPoolExecutor(max_workers=self.k8s_api_threadpool_workers)
+        self.executor = SingletonExecutor.instance(max_workers=self.k8s_api_threadpool_workers)
 
         # FIXME: Is this a memory leak? Will there be other side effects?
         self.pod_reflector = PodReflector.instance(parent=self, namespace=self.namespace)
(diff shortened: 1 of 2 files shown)


kubespawner/spawner.py
```diff
@@ -36,12 +184,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # By now, all the traitlets have been set, so we can use them to compute
         # other attributes
-        self.executor = ThreadPoolExecutor(max_workers=self.k8s_api_threadpool_workers)
+        self.executor = SingletonExecutor.instance(max_workers=self.k8s_api_threadpool_workers)
+
+        # FIXME: Is this a memory leak? Will there be other side effects?
```
minrk: what do you mean by memory leak here?

kubespawner/spawner.py
```diff
@@ -27,6 +31,150 @@
...
+                w = watch.Watch()
+                for ev in w.stream(
+                        self.api.list_namespaced_pod,
+                        self.namespace,
+                        label_selector=self.label_selector
+                ):
+                    cur_delay = 0.1
+                    pod = ev['object']
+                    if ev['type'] == 'DELETED':
+                        try:
```
yuvipanda: Ah, cool. I assume pop is threadsafe too?

kubespawner/spawner.py
```diff
@@ -27,6 +31,150 @@
...
+                    else:
+                        # This is an atomic operation on the dictionary!
+                        self.pods[pod.metadata.name] = pod
+            except:
+                cur_delay = cur_delay * 2
+                self.log.exception("Error when watching pods, retrying in %ss", cur_delay)
+                time.sleep(cur_delay)
+                if cur_delay > 30:
+                    self.log.fatal('Unable to connect to APIServer')
+                    sys.exit(1)
```
yuvipanda: +1

kubespawner/spawner.py
yuvipanda: Clarified comment - I was worried the thread that's doing the watching will live forever, but that's kinda what we wanna do anyway so it is ok.

kubespawner/spawner.py
yuvipanda: Done!

yuvipanda:
Did some cleanup, going to merge this now!

PR #64, opened 2017-07-21
Add a Kubernetes Ingress based Proxy implementation

yuvipanda:
You need an ingress controller set up to be able to use it. Simplest is the traefik one, which you can install with:

```
helm install stable/traefik --namespace=kube-system --name=ingress
```


kubespawner/proxy.py
```diff
@@ -0,0 +1,210 @@
...
+        if os.path.exists(ns_path):
+            with open(ns_path) as f:
+                return f.read().strip()
+        return 'default'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # other attributes
+        self.executor = ThreadPoolExecutor(max_workers=24)
```
minrk: Are you sure you want to run this in 24 threads? It's probably appropriate to make this configurable if it should be any value other than one.

When we talk to the docker API, we use one thread, because it can't actually handle all those requests being concurrent anyway, so the key is that it's off the main thread, but the requests aren't really concurrent with each other.

kubespawner/spawner.py
```diff
@@ -782,13 +797,43 @@ def start(self):
...
+            lambda: self.is_pod_running(self.pod_reflector.pods.get(self.pod_name, None)),
+            'pod/%s did not start in %s seconds!' % (self.pod_name, self.start_timeout),
+            timeout=self.start_timeout
+        )
+
+        pod = self.pod_reflector.pods[self.pod_name]
+        self.pod_reflector.add_deletion_callback(pod.metadata.name, self.stop_notify)
         return (pod.status.pod_ip, self.port)
 
+    def start_polling(self):
```
minrk: I think you can already set `poll_interval = 0` to disable polling without overriding the methods.

kubespawner/proxy.py
yuvipanda: Yeah, I agree - we should make this configurable. I need to remove all the hard coded numbers and make all of 'em configurable.

The kubernetes API master is highly concurrent, and benchmarks often hit it at hundreds/thousands of req/s without issues. We should be ok with a Threadpool I think!

In general we need to think about the various sizes that limit the number of concurrent actions we could be performing. Threadpool sizes are one of those!

kubespawner/spawner.py
yuvipanda: Ah, ok! I'll do that!

kubespawner/proxy.py
yuvipanda: So I spent more time thinkign about this and playing.

I think at least for KubeSpawner, these should be a multiple of the maximum number of spawns we can have pending at a time. That'll allow the spawns to go without bottlenecking on waiting on the threadpool. It also allows admins one single knob to tweak - the maximum concurrent spawns.

yuvipanda:
I've rebased this, and I think this needs a lot of careful review and testing. Specifically, we need to have solid answers for the following questions:

1. Will this cause us to miss events? Is just using resourceversion good enough?
2. If we miss events, will we recover appropriately?

We also need to do a lot of documenting, and decide what's the default mode for z2jh.

yuvipanda:
I've cleaned this up further, so this *only* adds the kubeproxy implementation. I have removed all the bits about making poll go away and be replaced with pure notifications. I'm not entirely sure if that is a good idea yet, so let's just do this one!

yuvipanda:
Note that this probably needs work to support named servers.

kubespawner/proxy.py
yuvipanda: I've made this 3x concurrent_spawn_limit now

yuvipanda:
Actually no, the proxy doesn't have to know about named servers at all. Yay!

kubespawner/spawner.py
```diff
@@ -49,7 +54,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # By now, all the traitlets have been set, so we can use them to compute
         # other attributes
-        self.executor = SingletonExecutor.instance(max_workers=self.k8s_api_threadpool_workers)
+        # FIXME: Can't actually use the .instance here and in IngressReflector, traitlets will throw
```
yuvipanda: @minrk can you suggest a clean way to do this? I want the executor to be a singleton since we want all spawners to share the same threadpool.

minrk: The singleton classes are mainly useful for registering global (process-wide) instances that aren't attached to anything, like IPython itself, or the current Application. If it's a class-global, a normal class attribute ought to work better:

```python
# class-level
executor = None

def __init__(...):
    ....
    if self.executor is None:
        # instantiate executor at the class level
        self.__class__.executor = ThreadPoolExecutor(max_workers...)
    # self.executor from now on is shared by all instances
```

yuvipanda: @minrk I've removed all places that was using trailets for singletons now!

yuvipanda:
I need to test how this behaves on JupyterHub restarts a bit more, and also probably a few stress tests

kubespawner/proxy.py
```diff
@@ -0,0 +1,244 @@
...
+        ns_path = '/var/run/secrets/kubernetes.io/serviceaccount/namespace'
+        if os.path.exists(ns_path):
+            with open(ns_path) as f:
+                return f.read().strip()
+        return 'default'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # We want 3x concurrent spawn limit as our threadpool. This means that if
```
yuvipanda: How do you feel about this logic, @minrk? It vaguely feels right to me, but I can't make a strong case for it yet.

kubespawner/proxy.py
```diff
@@ -0,0 +1,244 @@
...
+        self.extension_api = client.ExtensionsV1beta1Api()
+
+    @run_on_executor
+    def asynchronize(self, method, *args, **kwargs):
+        return method(*args, **kwargs)
+
+    def safe_name_for_routespec(self, routespec):
+        safe_chars = set(string.ascii_lowercase + string.digits)
+        safe_name = generate_hashed_slug(
+            'jupyter-' + escapism.escape(routespec, safe=safe_chars, escape_char='-') + '-route'
```
yuvipanda: Should we make this configurable? We could also pre-replace all '/' with -, since otherwise the objects look kinda unweildy, like 'jupyter--2fuser-2ftester-2f-route'.

kubespawner/proxy.py
yuvipanda: We never actually will have all 3 of these requests going at once, since we wait for each object to show up in our reflector first. Maybe we don't have to do that (and can just rely on the redirects=n looping to fix it), since that will give us a big performance boost.

Perhaps the better thing to do is to figure out how to put multiple objects atomically in the same http request. That seems like the 'right' thing to do.

kubespawner/proxy.py
minrk: Making escape_char configurable is probably a good idea. I noticed in binderhub that since we are creating users with `-` in their names, they are getting pods called `jupyter-2dfoo-2d`, etc. This escaping `-` is technically necessary to rigorously avoid collisions, but not necessary in the binderhub case in particular, where we can be sure that there won't be collisions.

yuvipanda: +1, I'll make that happen here and on the spawner too.

kubespawner/reflector.py
```diff
@@ -29,12 +28,46 @@ class PodReflector(SingletonConfigurable):
...
+        help="""
+        Human readable name for kind of object we're watching for.
+
+        Used for diagnostic messages.
+        """
+    )
+
+    list_method_name = Unicode(
+        None,
+        allow_none=True,
```
minrk: From the usage below, `list_method_name` is *not* allowed to be None, as `getattr(obj, None)` will always fail. Probably best to leave this without setting None or allow_none.

kubespawner/reflector.py
```diff
@@ -29,12 +28,46 @@ class PodReflector(SingletonConfigurable):
...
+        of the form list_namespaced_<resource> - for example, `list_namespaced_pod` will
+        give you a PodReflector.
+
+        This must be set by a subclass.
+        """
+    )
+
+    api_group_name = Unicode(
+        'CoreV1Api',
+        allow_none=False,
```
minrk: No need to specify `allow_none=False`. In general, I don't think we should use `allow_none` on Unicode traits unless there's a specific, odd case for empty strings meaning something specific.

kubespawner/proxy.py
minrk: Maybe use 1x instead of 3x? It seems okay when hitting capacity to allow a single spawn's requests to serialize. It doesn't seem important that all possible requests are in flight at once.

kubespawner/spawner.py
```diff
@@ -11,45 +11,61 @@
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # By now, all the traitlets have been set, so we can use them to compute
         # other attributes
-        self.executor = SingletonExecutor.instance(max_workers=self.k8s_api_threadpool_workers)
+        if self.__class__.executor is None:
+            self.__class__.executor = ThreadPoolExecutor(
+                max_workers=self.k8s_api_threadpool_workers
+            )
+        self.executor = self.__class__.executor
```
minrk: This extra assignment is not necessary. `self.executor` resolves to the class attribute already.

kubespawner/spawner.py
```diff
@@ -58,8 +74,12 @@ def on_reflector_failure():
...
         # This will start watching in __init__, so it'll start the first
         # time any spawner object is created. Not ideal but works!
-        self.pod_reflector = PodReflector.instance(parent=self, namespace=self.namespace,
-            on_failure=on_reflector_failure)
+        if self.__class__.pod_reflector is None:
+            self.__class__.pod_reflector = PodReflector(
+                parent=self, namespace=self.namespace,
+                on_failure=on_reflector_failure
+            )
+        self.pod_reflector = self.__class__.pod_reflector
```
minrk: same here

kubespawner/proxy.py
```diff
@@ -0,0 +1,244 @@
...
+
+    @property
+    def endpoints(self):
+        return self.resources
+
+class KubeIngressProxy(Proxy):
+    namespace = Unicode(
+        config=True,
+        help="""
+        Kubernetes namespace to spawn ingresses for single-user servers in.
```
willingc: Perhaps: `Kubernetes namespace, specified by ingress, to spawn single-user servers in.`

kubespawner/proxy.py
willingc: I agree with @minrk on 1x. It will be simpler to test initially and can be increased in a future refactor if needed.

kubespawner/proxy.py
```diff
@@ -0,0 +1,244 @@
...
+
+        delete_ingress = self.asynchronize(
+                self.extension_api.delete_namespaced_ingress,
+                name=safe_name,
+                namespace=self.namespace,
+                body=delete_options,
+                grace_period_seconds=0
+        )
+
+        # This seems like cleanest way to parallelize all three of these while
```
willingc: Question for my own understanding: does order of deletion matter at all (i.e is there any dependence on each other)?

kubespawner/proxy.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+    @gen.coroutine
+    def get_all_routes(self):
+        # copy everything, because iterating over this directly is not threadsafe
+        # FIXME: is this performance intensive? It could be! Measure?
+        # FIXME: Validate that this shallow copy *is* thread safe
+        ingress_copy = dict(self.ingress_reflector.ingresses)
+        routes = {
+            i.metadata.annotations['hub.jupyter.org/proxy-routespec']:
```
willingc: What is `i`? Each route? Original before copy? Define in a comment or docstring. Thanks 👍

kubespawner/reflector.py
```diff
@@ -1,20 +1,19 @@
...
+from traitlets.config import LoggingConfigurable
 from traitlets import Any, Dict, Unicode
 from kubernetes import client, config, watch
 from tornado.ioloop import IOLoop
 
-class PodReflector(SingletonConfigurable):
+class NamespacedResourceReflector(LoggingConfigurable):
     """
-    Local up-to-date copy of a set of kubernetes pods
+    Local up-to-date copy of a set of kubernetes resources.
```
willingc: Is this more accurate:

Base class for a local up-to-date copy of a type of kubernetes resource

kubespawner/reflector.py
```diff
@@ -48,7 +81,7 @@ def __init__(self, *args, **kwargs):
             config.load_incluster_config()
         except config.ConfigException:
             config.load_kube_config()
-        self.api = client.CoreV1Api()
+        self.api = getattr(client, self.api_group_name)()
 
         # FIXME: Protect against malicious labels?
```
willingc: Do we also need to protect against malicious targets (earlier in this PR)?

kubespawner/spawner.py
```diff
@@ -11,45 +11,61 @@
...
+        return self.resources
 
 class KubeSpawner(Spawner):
     """
     Implement a JupyterHub spawner to spawn pods in a Kubernetes Cluster.
     """
+
+    # We want to have one threadpool executor that is shared across all spawner objects
+    # This is initialized by the first spawner that is created
+    executor = None
```
willingc: Question on naming: Would it be better to name `shared_threadpool_executor`?

Either way we should move this import comment into the general class docstring.

kubespawner/spawner.py
```diff
@@ -727,7 +747,8 @@ def is_pod_running(self, pod):
         pod must be a dictionary representing a Pod kubernetes API object.
         """
         # FIXME: Validate if this is really the best way
-        is_running = pod.status.phase == 'Running' and \
+        is_running = pod is not None and \
```
willingc: This should be converted to the Pythonic `( )` instead of `\`

kubespawner/utils.py
```diff
@@ -1,132 +1,28 @@
...
-    )
+    # if the length of the slug is shorter than limit - hash_length, we just
+    # use it directly. Good enough!
+    if len(slug) < (limit - hash_length):
+        return slug
 
-    def make_request(url, **kwargs):
-        """
-        Make & return a HTTPRequest object suited to making requests to a Kubernetes cluster
+    # If not, we pick the first limit - hash_length chars from slug & hash the rest.
```
willingc: Put what approach we are using in the docstring if > 63 char

kubespawner/proxy.py
yuvipanda: Each single-user server has an ingress associated with it, and both of these things belong to a namespace. This is just for creating ingress objects that corrsepond to spawned servers. I'm confused by the alternate suggestion and am unsure what it means :)

kubespawner/spawner.py
yuvipanda: This just follows Tornado's convention for executors: http://www.tornadoweb.org/en/stable/concurrent.html#tornado.concurrent.run_on_executor

kubespawner/proxy.py
yuvipanda: yup! Added a line here!

kubespawner/proxy.py
yuvipanda: I've renamed it to 'ingress' which I think clarifies what they are (ingress objects). lmk if you think an additional doc line is still warranted!

kubespawner/utils.py
yuvipanda: Done.

kubespawner/proxy.py
yuvipanda: Yeah, I agree! Upon further thought, this will also never use more than 1x, since we serialize these creates anyway. I'll update!

kubespawner/proxy.py
yuvipanda: It does! I've added a comment!

kubespawner/proxy.py
yuvipanda: I've renamed it to 'ingress' now, since these are all ingress objects. Let me know if that's good enough!

kubespawner/reflector.py
yuvipanda: I don't think so right now, since these can only be specified by users editing jupyterhub_config.py files. I think we should take a pass in a separate PR doing more validation here, but for now I think this is fine.

kubespawner/reflector.py
yuvipanda: Yeah, I think this is one of the cases where code I've written (default to None for default) has been different from code you've written (default to "", not allow None). Since it looks like your style is more prevelant in the Jupyter community & I can't make a compelling case for my style, I'll switch to yours from now!

kubespawner/reflector.py
yuvipanda: Yup! updated!

kubespawner/utils.py
yuvipanda: Done

kubespawner/proxy.py
willingc: +1 to renaming. no doc line needed.

kubespawner/proxy.py
willingc: Is this accurate? What's confusing me is I thought ingresses are essentially rules.

Kubernetes namespace for both single users servers to be spawned and their ingresses.

yuvipanda: Everything in kubernetes is an object that represents something. In this case, Ingresses are objects that represent HTTP routing rules.

This only sets the namespace for the ingresses to be created, though. The namespace for singleuser servers is specified in KubeSpawner.namespace, rather than KubeProxy.namespace.

yuvipanda: How about:

"Kubernetes namespace to create Ingress objects in.

This corresponds to KubeSpawner.namespace, which sets the namespace single-user servers
are spawned in. For most use cases, these two should be the same
"

willingc: Love the suggestion. Thanks for the explanation @yuvipanda.

yuvipanda:
I definitely want to run a 5k user stress test before merging this though.

minrk:
@yuvipanda feel free to merge if/when you get that stress test done.

yuvipanda:
Am going to merge this now to prevent more merge conflicts, and do a perf test on this after.

gsemet:
Hello. Is there a documentation on how to use Traefik as ingress for jupyterhub? We heavily use Traefik on our kubernetes cluster, it rocks a lot ! So if we can somehow "remove" the "proxy" pod and let Traefik do the routing job for started servers, it would be very nice!

willingc:
Hi @gsemet, We think Traefik rocks too :-)

I'm not aware of any documentation that currently exists for using Traefik. @yuvipanda @minrk are you aware of anyone using Traefik in production?

gsemet:
Just to be clear: we already have Traefik in front of our jupyterhub. Just their is still this proxy pod that appears the single point of failure.

PR #67, opened 2017-07-27
Add Sphinx docs

willingc:
[Rendering of WIP docs](http://test-kubespawner.readthedocs.io/en/doc-skeleton/utils.html)


yuvipanda:
This looks like a great start, @willingc!

willingc:
@yuvipanda I've taken the docs out of WIP adding docs for all of the functions and a subset of the README as an overview. 

To get Travis to not error on install of the spawner and test runs, I made some changes to the `.travis.yml` and `setup.py` as well as adding a `dev-requirements.txt` for testing. Feel free to delete if you can get Travis to run without these changes.

yuvipanda:
@willingc awesome! Do you know if this will also pick out the help text for each traitlet? I guess that's 'user documentation' that we should try to autogenerate...

willingc:
As written, it does not. Let's merge this and do the traitlets in a separate PR. To clarify, when you say help text for traitlet, do you mean the help text for each attribute in the classes in `reflector.py` and `spawner.py`?

yuvipanda:
Thanks @willingc!

Yup, that's what I meant. It's what allows users to configure kubespawner however they want, and should be our 'reference documentation' I guess.

willingc:
Will see if I can get to this on Monday. Thanks for clarifying.

PR #69, opened 2017-08-04
Add flag that runs the spawned container with a privileged security context

keithcallenberg:
When using packages like s3fs to mount S3 buckets, it is necessary to run containers in privileged security context. This flag makes that possible.


yuvipanda:
Thanks for the patch, @keithcallenberg! Do you think you can add a test for this?

keithcallenberg:
Sure, I will do so today, or tomorrow at the latest.

Keith

keithcallenberg:
Third commit's the charm!

PR #70, opened 2017-08-06
Autodoc traits

willingc:
Here you go @yuvipanda :-)


willingc:
Here's the built and rendered version: http://test-kubespawner.readthedocs.io/en/autodoc-traits/index.html

yuvipanda:
This is *awesome*. We should do this for JupyterHub too!

willingc:
Thanks! I did it for BinderHub last night. JupyterHub is already done ;-)

PR #81, opened 2017-09-15
use resource_version to avoid missing events

minrk:
We have observed a race missing DELETE events sent before the stream is connected:

1. request deletion of pod X
2. list_and_update shows pod X is running
3. pod X is deleted
4. watch stream is connected

which would result in pod X never being notified that it stopped until the next `list_and_update` call. 

Kubernetes API has a `resource_version` field for exactly this purpose - to hook up a list & watch pair ([ref](http://borismattijssen.github.io/articles/kubernetes-informers-controllers-reflectors-stores).

closes https://github.com/jupyterhub/jupyterhub/issues/1420

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -65,6 +65,8 @@ class PodReflector(SingletonConfigurable):
         )
         # This is an atomic operation on the dictionary!
         self.pods = {p.metadata.name: p for p in initial_pods.items}
+        # return the resource version so we can hook up a watch
+        return initial_pods.metadata.resource_version
 
     def _watch_and_update(self):
         """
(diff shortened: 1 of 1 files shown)


minrk:
@yuvipanda this may have prevented the latest binder outage (https://github.com/jupyterhub/mybinder.org-deploy/pull/23), but I suspect it is not the only way for the reflector to miss events.

yuvipanda:
Yeah, I'm not entirely sure if this is the right thing to do, either. http://borismattijssen.github.io/articles/kubernetes-informers-controllers-reflectors-stores explains what kubernetes internally does (periodic refreshes), so we should follow the same thing they are doing...

minrk:
@yuvipanda thanks for the link, it has exactly the answer we need. Updated to use resource_version to ensure list & watch are in sync.

yuvipanda:
w00t, I like this better!

PR #83, opened 2017-09-19
Handle named-servers functionality enabled by jupyterhub 0.8x

Analect:
@yuvipanda @minrk 
This PR isn't working as yet (hence WIP), but adding here for some guidance, as I'm probably doing this in an awkward way.

I extended the `_extend_user_properties` [here](https://github.com/Analect/kubespawner/blob/5fce05901a5127a9bd74829aab4215133b1b48d3/kubespawner/spawner.py#L585-L609), hoping to set a `servername` variable in the case where a named-server has been initialised.

The objective, as per [here](https://github.com/Analect/kubespawner/commit/eeba9e55f00b97ddca34f8381604599c56b29482) was to then use that `servername` to extend 'pod_name_template' and 'pvc_name_template' where appropriate, but I'm clearly doing it in syntactically incorrect way!  ... 

Could either of you set me straight on best achieving this.  Thanks.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -150,7 +150,7 @@ class KubeSpawner(Spawner):
     ).tag(config=True)
 
     pod_name_template = Unicode(
-        'jupyter-{username}' + '-{servername}' if servername is not None,
+        'jupyter-{username}' + '-{servername}' if getattr(self, 'name', None) is not None,
         config=True,
         help="""
         Template to use to form the name of user's pods.
@@ -176,7 +176,7 @@ class KubeSpawner(Spawner):
     )
 
     pvc_name_template = Unicode(
-        'claim-{username}' + '-{servername}' if servername is not None,
+        'claim-{username}' + '-{servername}' if getattr(self, 'name', None) is not None,
         config=True,
         help="""
         Template to use to form the name of user's pvc.
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -608,12 +608,12 @@ class KubeSpawner(Spawner):
     def _expand_user_properties(self, template):
         # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
+
         # Set servername based on whether named-server initialised
-        temp_name = getattr(self, 'name', '')
-        if temp_name:
-            server_name = '-' + temp_name
+        if self.name:
+            servername = '-{}'.format(self.name)
         else:
-            server_name = ''
+            servername = ''
 
         legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
         safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
@@ -621,7 +621,7 @@ class KubeSpawner(Spawner):
             userid=self.user.id,
             username=safe_username,
             legacy_escape_username=legacy_escaped_username,
-            servername=server_name
+            servername=servername
             )
 
     def _expand_all(self, src):
@@ -663,10 +663,9 @@ class KubeSpawner(Spawner):
             'hub.jupyter.org/username': escapism.escape(self.user.name)
         }
 
-        # check if a named-server servername has been set and if so, extend pod labels.
-        temp_servername = getattr(self, 'name', None)
-        if temp_servername:
-            labels['hub.jupyter.org/servername']=temp_servername
+        if self.name:
+            # FIXME: Make sure this is dns safe?
+            labels['hub.jupyter.org/servername'] = self.name
 
         labels.update(self._expand_all(self.singleuser_extra_labels))
 
@@ -708,9 +707,9 @@ class KubeSpawner(Spawner):
         }
 
         # check if a named-server servername has been set and if so, extend pvc labels.
-        temp_servername = getattr(self, 'name', None)
-        if temp_servername:
-            labels['hub.jupyter.org/servername']=temp_servername
+        if self.name:
+            # FIXME: make sure this is DNS safe?
+            labels['hub.jupyter.org/servername'] = self.name
 
         labels.update(self._expand_all(self.user_storage_extra_labels))
         return make_pvc(
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -586,9 +586,9 @@ class KubeSpawner(Spawner):
         # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
         # Set servername based on whether named-server initialised
-        name = getattr(self, name, '')
-        if name:
-            servername = '-' + name
+        temp_name = getattr(self, name, '')
+        if temp_name:
+            servername = '-' + temp_name
         else:
             servername = ''
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -150,7 +150,7 @@ class KubeSpawner(Spawner):
     ).tag(config=True)
 
     pod_name_template = Unicode(
-        'jupyter-{username}' + '-{servername}' if getattr(self, 'name', None) is not None,
+        'jupyter-{username}{servername}',
         config=True,
         help="""
         Template to use to form the name of user's pods.
@@ -176,7 +176,7 @@ class KubeSpawner(Spawner):
     )
 
     pvc_name_template = Unicode(
-        'claim-{username}' + '-{servername}' if getattr(self, 'name', None) is not None,
+        'claim-{username}{servername}',
         config=True,
         help="""
         Template to use to form the name of user's pvc.
@@ -585,28 +585,21 @@ class KubeSpawner(Spawner):
     def _expand_user_properties(self, template):
         # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
-        # Test if a named-server exists
-        if getattr(self, 'name', None) is None:
-            legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
-            safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
-            return template.format(
-                userid=self.user.id,
-                username=safe_username,
-                legacy_escape_username=legacy_escaped_username
-                )
-        # If a named-server exists, then extend the properties    
+        # Set servername based on whether named-server initialised
+        name = getattr(self, name, '')
+        if name:
+            servername = '-' + name
         else:
-            legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
-            safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
-            legacy_escaped_servername = ''.join([s if s in safe_chars else '-' for s in self.name.lower()])
-            safe_servername = escapism.escape(self.name, safe=safe_chars, escape_char='-').lower()
-            return template.format(
-                userid=self.user.id,
-                username=safe_username,
-                legacy_escape_username=legacy_escaped_username,
-                servername=safe_servername,
-                legacy_escape_servername=legacy_escaped_servername
-                )
+            servername = ''
+
+        legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
+        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
+        return template.format(
+            userid=self.user.id,
+            username=safe_username,
+            legacy_escape_username=legacy_escaped_username,
+            servername=servername
+            )
 
     def _expand_all(self, src):
         if isinstance(src, list):
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -586,7 +586,7 @@ class KubeSpawner(Spawner):
         # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
         # Set servername based on whether named-server initialised
-        temp_name = getattr(self, name, '')
+        temp_name = getattr(self, 'name', '')
         if temp_name:
             servername = '-' + temp_name
         else:
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -594,8 +594,12 @@ class KubeSpawner(Spawner):
 
         legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
         safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
-        d = {'username': safe_username, 'servername': server_name}
-        return self.pod_name_template.format(**d)
+        return template.format(
+            userid=self.user.id,
+            username=safe_username,
+            legacy_escape_username=legacy_escaped_username,
+            servername=server_name
+            )
 
     def _expand_all(self, src):
         if isinstance(src, list):
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -861,16 +861,3 @@ class KubeSpawner(Spawner):
                 args[i] = '--hub-api-url="%s"' % (self.accessible_hub_api_url)
                 break
         return args
-
-    def get_env(self):
-        # HACK: This is deprecated, and should be removed soon.
-        # We set these to be compatible with DockerSpawner and earlie KubeSpawner
-        env = super(KubeSpawner, self).get_env()
-        env.update({
-            'JPY_USER': self.user.name,
-            'JPY_COOKIE_NAME': self.server.cookie_name,
-            'JPY_BASE_URL': self.server.base_url,
-            'JPY_HUB_PREFIX': self.hub.server.base_url,
-            'JPY_HUB_API_URL': self.accessible_hub_api_url
-        })
-        return env
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -150,7 +150,7 @@ class KubeSpawner(Spawner):
     ).tag(config=True)
 
     pod_name_template = Unicode(
-        'jupyter-{username}',
+        'jupyter-{username}' + '-{servername}' if servername is not None,
         config=True,
         help="""
         Template to use to form the name of user's pods.
@@ -176,7 +176,7 @@ class KubeSpawner(Spawner):
     )
 
     pvc_name_template = Unicode(
-        'claim-{username}',
+        'claim-{username}' + '-{servername}' if servername is not None,
         config=True,
         help="""
         Template to use to form the name of user's pvc.
@@ -583,15 +583,30 @@ class KubeSpawner(Spawner):
     )
 
     def _expand_user_properties(self, template):
-        # Make sure username matches the restrictions for DNS labels
+        # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
-        legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
-        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
-        return template.format(
-            userid=self.user.id,
-            username=safe_username,
-            legacy_escape_username=legacy_escaped_username
-        )
+        # Test if a named-server exists
+        if getattr(self, 'name', None) is None:
+            legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
+            safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
+            return template.format(
+                userid=self.user.id,
+                username=safe_username,
+                legacy_escape_username=legacy_escaped_username
+                )
+        # If a named-server exists, then extend the properties    
+        else:
+            legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
+            safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
+            legacy_escaped_servername = ''.join([s if s in safe_chars else '-' for s in self.name.lower()])
+            safe_servername = escapism.escape(self.name, safe=safe_chars, escape_char='-').lower()
+            return template.format(
+                userid=self.user.id,
+                username=safe_username,
+                legacy_escape_username=legacy_escaped_username,
+                servername=safe_servername,
+                legacy_escape_servername=legacy_escaped_servername
+                )
 
     def _expand_all(self, src):
         if isinstance(src, list):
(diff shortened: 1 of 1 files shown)


kubespawner/spawner.py
```diff
@@ -150,7 +150,7 @@ def _namespace_default(self):
     ).tag(config=True)
 
     pod_name_template = Unicode(
-        'jupyter-{username}',
+        'jupyter-{username}' + '-{servername}' if getattr(self, 'name', None) is not None,
```
minrk: I think this can be '{username}{servername}' and add the `-` *in* `servername` when the template is applied.

kubespawner/spawner.py
```diff
@@ -583,15 +583,30 @@ def _hub_connect_port_default(self):
...
+            legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
+            safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
+            legacy_escaped_servername = ''.join([s if s in safe_chars else '-' for s in self.name.lower()])
+            safe_servername = escapism.escape(self.name, safe=safe_chars, escape_char='-').lower()
+            return template.format(
+                userid=self.user.id,
+                username=safe_username,
+                legacy_escape_username=legacy_escaped_username,
+                servername=safe_servername,
+                legacy_escape_servername=legacy_escaped_servername
```
minrk: Probably don't need `legacy_escape_servername` since this wasn't present before.

kubespawner/spawner.py
```diff
@@ -583,15 +583,30 @@ def _hub_connect_port_default(self):
     )
 
     def _expand_user_properties(self, template):
-        # Make sure username matches the restrictions for DNS labels
+        # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
-        legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
-        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
```
minrk: I think we can just add:

```python
name = getattr(self, name, '')
if name:
    servername = '-' + name
else:
    servername = ''

format(
  ...
  servername=servername,
)
```

instead of branching at the higher level.

kubespawner/spawner.py
```diff
@@ -850,8 +877,8 @@ def get_env(self):
         env = super(KubeSpawner, self).get_env()
         env.update({
             'JPY_USER': self.user.name,
-            'JPY_COOKIE_NAME': self.user.server.cookie_name,
-            'JPY_BASE_URL': self.user.server.base_url,
+            'JPY_COOKIE_NAME': self.server.cookie_name,
+            'JPY_BASE_URL': self.server.base_url,
```
minrk: I think this entire `.get_env()` method can be deleted. It's been deprecated for a while and is not needed by recent images.

Analect:
@minrk 
I ended up simplifying things by trying to assign `temp_name = getattr(self, name, '')`, however I'm still hitting an error in trying to get the hub running.  I've positioned that assignment within the `_expand_user_properties` function, thinking that I can still access the 'name' from jupyterhub spawner in there, but perhaps that's not the case.  Can you see where I might be going wrong?

```
[I 2017-09-19 15:34:17.816 JupyterHub app:837] Loading cookie_secret from env[JPY_COOKIE_SECRET]
[I 2017-09-19 15:34:17.858 JupyterHub app:986] Not using whitelist. Any authenticated user will be allowed.
[I 2017-09-19 15:34:17.939 JupyterHub reflector:94] watching for pods with label selector heritage=jupyterhub,component=singleuser-server in namespace default
[E 2017-09-19 15:34:17.939 JupyterHub app:1626]
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/app.py", line 1623, in launch_instance_async
        yield self.initialize(argv)
      File "/usr/lib/python3.5/types.py", line 179, in throw
        return self.__wrapped.throw(tp, *rest)
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/app.py", line 1385, in initialize
        yield self.init_spawners()
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/app.py", line 1213, in init_spawners
        self.users[orm_user.id] = user = User(orm_user, self.tornado_settings)
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/user.py", line 178, in __init__
        self.spawners[name] = self._new_spawner(name)
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/user.py", line 208, in _new_spawner
        spawner = spawner_class(**spawn_kwargs)
      File "/usr/local/lib/python3.5/dist-packages/kubespawner/spawner.py", line 61, in __init__
      File "/usr/local/lib/python3.5/dist-packages/kubespawner/spawner.py", line 589, in _expand_user_properties
        temp_name = getattr(self, name, '')
    NameError: name 'name' is not defined
```

Analect:
OK.  I managed to get things running by adjusting the `_expand_user_properties` to function in a similar way to the implementation over on dockerspawner [here](https://github.com/jupyterhub/dockerspawner/blob/00fec063f99c1b1d7333c9f3f6799844af07e8f6/dockerspawner/dockerspawner.py#L376-L380).

```
def _expand_user_properties(self, template):
        # Make sure username and servername match the restrictions for DNS labels
        safe_chars = set(string.ascii_lowercase + string.digits)
        # Set servername based on whether named-server initialised
        temp_name = getattr(self, 'name', '')
        if temp_name:
            server_name = '-' + temp_name
        else:
            server_name = ''

        legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
        d = {'username': safe_username, 'servername': server_name}
        return self.pod_name_template.format(**d)
```
I'm now able to start various named-servers for `testuser` ... test1, test2, test3 ... resulting in pods `jupyter-testuser-test1`, `jupyter-testuser-test2` etc ... calling the hub API in this form:
`curl -v -X POST -H "Authorization: token 1f9aa148exxxxxx" "http://<my_jupyterhub_ip>/hub/api/users/testuser/servers/test1"`

The notebook server is navigable at `http://<my_jupyterhub_ip>/user/testuser/test1/tree?`

... and stop them with:
`curl -v -X DELETE -H "Authorization: token 1f9aa148exxxxxx" "http://<my_jupyterhub_ip>/hub/api/users/testuser/servers/test1"`

I'm also able to stop/start a default server with:
`curl -v -X POST -H "Authorization: token 1f9aa148exxxxxx" "http://<my_jupyterhub_ip>/hub/api/users/testuser/server"`

... resulting in a pod `jupyter-testuser`

For now, I switched off storage (ie. creation of PVCs) ... but that should work OK if I can somehow `return self.pvc_name_template.format(**d)` ... as part of the same `_expand_user_properties` function.  Can one return two items in the same line?

Also, I'm not making the `servername` safe in any way (ie. removing characters that kubernetes may not like), which may be deemed necessary.

kubespawner/spawner.py
```diff
@@ -583,15 +583,19 @@ def _hub_connect_port_default(self):
...
+            server_name = ''
+
         legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
         safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
-        return template.format(
-            userid=self.user.id,
-            username=safe_username,
-            legacy_escape_username=legacy_escaped_username
-        )
+        d = {'username': safe_username, 'servername': server_name}
```
minrk: Instead of rewriting this dict, you can just add `servername=server_name` to the original dict here. As it is now, a few keys have been lost and it's hardcoded `self.pod_name_template` instead of `template`

Analect: I tried the other approach every which way and couldn't get it to work.

Analect:
@minrk @yuvipanda 
Based on the commits above, with `storage: type: none` set in the config.yaml of the helm chart, starting/stopping named-servers and a default server is fine.

However, introducing handling of storage per spawned server complicates things.
It seems that [this](https://github.com/Analect/kubespawner/blob/b576b90cfae5171fd466553f04de99d07bdf0c4e/kubespawner/spawner.py#L597-L602) return of dictionary including `servername` to `template`, doesn't get picked up in the [`pvc_name_template`](https://github.com/Analect/kubespawner/blob/b576b90cfae5171fd466553f04de99d07bdf0c4e/kubespawner/spawner.py#L178) that is then used by [`pvc_name`](https://github.com/Analect/kubespawner/blob/b576b90cfae5171fd466553f04de99d07bdf0c4e/kubespawner/spawner.py#L62) as part of the [`get_pvc_manifest`](https://github.com/Analect/kubespawner/blob/b576b90cfae5171fd466553f04de99d07bdf0c4e/kubespawner/spawner.py#L682-L701).

I was then trying to see if these pvc names could be set directly in jupyterhub_config.py itself, as per below, but it's not clear to me how  I pull in the named-server `servername` in this instance.  It seems that most values for this config files are either derived from environment variables or derived from values in the `config.yaml` on a helm-chart, but in the case of named-servers, they are going to have to be derived real-time, when an API call is made to spawn a server.  Below ... I'm obviously testing for the existence of `servername` ... but I need to somehow access that value ... but I don't think I can use something like `getattr(self, 'name', '')`. since `self` is not available within the `jupyterhub_config.py` file.

```
storage_type = get_config('singleuser.storage.type')
if storage_type == 'dynamic':
    # test if provisioning for a named-server
    if servername:
        c.KubeSpawner.pvc_name_template = 'claim-{username}{servername}'
    else:
        c.KubeSpawner.pvc_name_template = 'claim-{username}'
    c.KubeSpawner.user_storage_pvc_ensure = True
    c.KubeSpawner.user_storage_class = get_config('singleuser.storage.class')
    c.KubeSpawner.user_storage_access_modes = ['ReadWriteOnce']
    c.KubeSpawner.user_storage_capacity = get_config('singleuser.storage.capacity')
    # Add volumes to singleuser pods
    # test if provisioning for a named-server
    if servername:
        c.KubeSpawner.volumes = [
            {
                'name': 'volume-{username}{servername}',
                'persistentVolumeClaim': {
                    'claimName': 'claim-{username}{servername}'
                }
            }
        ]
            {
                'mountPath': get_config('singleuser.storage.home_mount_path'),
                'name': 'volume-{username}{servername}'
            }
        ]
    else:
        c.KubeSpawner.volumes = [
            {
                'name': 'volume-{username}',
                'persistentVolumeClaim': {
                    'claimName': 'claim-{username}'
                }
            }
        ]
            {
                'mountPath': get_config('singleuser.storage.home_mount_path'),
                'name': 'volume-{username}'
            }
        ]
```
Can either of you suggest a course of action?

kubespawner/spawner.py
yuvipanda: That should probably be a different PR though. Agree otherwise.

yuvipanda:
Are you sure you can't use {servername} in pvc_template? According to https://github.com/Analect/kubespawner/blob/b576b90cfae5171fd466553f04de99d07bdf0c4e/kubespawner/spawner.py#L62 it looks like it should be expanded.

Analect:
OK, it seems that the default `jupyterhub_config.py` in zero-to-jupyterhub-k8s will need to change for this to work.  Specifically, these [settings](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/images/hub/jupyterhub_config.py#L72-L73) are over-riding the ability for the kubespawner to set the `pvc_name_template` properly to encompass a `servername`.

If I change the `jupyterhub_config.py` in the helm chart to this (below), then everything seems to work OK ... ie. it creates the appropriate pvc (and requisite pv) as `claim-<username>-<servername>` ... or just `claim-<username>` in the case of a default (ie. non named-server), which is the expected behaviour.

Note, I'm not adding a `-` between `{username}{servername}` below, as we are handling this by including the dash ahead of the servername, in the case of a named-server [here](https://github.com/Analect/kubespawner/blob/handle-named-servers/kubespawner/spawner.py#L591).

```
storage_type = get_config('singleuser.storage.type')
if storage_type == 'dynamic':
    c.KubeSpawner.pvc_name_template = 'claim-{username}{servername}'
    c.KubeSpawner.user_storage_pvc_ensure = True
    c.KubeSpawner.user_storage_class = get_config('singleuser.storage.class')
    c.KubeSpawner.user_storage_access_modes = ['ReadWriteOnce']
    c.KubeSpawner.user_storage_capacity = get_config('singleuser.storage.capacity')

    # Add volumes to singleuser pods
    c.KubeSpawner.volumes = [
        {
            'name': 'volume-{username}{servername}',
            'persistentVolumeClaim': {
                'claimName': 'claim-{username}{servername}'
            }
        }
    ]
        {
            'mountPath': get_config('singleuser.storage.home_mount_path'),
            'name': 'volume-{username}{servername}'
        }
    ]
```
Let me know if you are happy for me to create a pull request against the zero-to-jupyterhub-k8s repo to reflect the changes above ... and if so, is `master` of that repo now the one that is expected to work against v0.8 of the hub?

yuvipanda:
This will also make kubespawner require 0.8, so I'd want to do a release before we merge this.

minrk:
> This will also make kubespawner require 0.8

Does it? The server-name support looks to be added in an 0.7-safe way (`getattr(self, 'name', '')`).

yuvipanda:
Nope, running this on 0.7.2 gives me:

```
[E 2017-10-02 10:47:05.197 JupyterHub web:1590] Uncaught exception POST /hub/login?next= (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8000', method='POST', uri='/hub/login?next=', version='HTTP/1.1', remote_ip='127.0.0.1', headers={'X-Forwarded-For': '127.0.0.1', 'Connection': 'close', 'Content-Length': '25', 'X-Forwarded-Proto': 'http', 'Accept-Encoding': 'gzip, deflate', 'X-Forwarded-Host': 'localhost:8000', 'Cookie': '_ga=GA1.1.2130754625.1498797034; username-localhost-8888="2|1:0|10:1505332071|23:username-localhost-8888|44:NDg0NTRlYjUxZTA2NDI3M2I3ZTU3OTllODFiYzdjY2U=|69d5c447c0c62a1d31d7196e4f33347ed3d4ae9df423e0b9abf0fdd319bff71e"; _xsrf=2|04848945|bd6f24b22ace937275398cd7b43e95ec|1505180487; username-localhost-8889="2|1:0|10:1505259546|23:username-localhost-8889|44:ZGM2MjE3OGYxZjYzNDI0MmJkYWNjYjY0YTU4NDcyNjI=|9185bff62e0633c3d93976d236977356ffee2fea2e1dbbddcc2bc513dbc65e27"', 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64; rv:57.0) Gecko/20100101 Firefox/57.0', 'Upgrade-Insecure-Requests': '1', 'Referer': 'http://localhost:8000/hub/login', 'Dnt': '1', 'X-Forwarded-Port': '8000', 'Accept-Language': 'en-US,en;q=0.5', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8', 'Content-Type': 'application/x-www-form-urlencoded', 'Host': 'localhost:8000'})
      File "/home/yuvipanda/code/kubespawner/lib/python3.5/site-packages/tornado/web.py", line 1511, in _execute
        result = yield result
      File "/home/yuvipanda/code/kubespawner/lib/python3.5/site-packages/jupyterhub/handlers/login.py", line 84, in post
        yield self.spawn_single_user(user)
      File "/home/yuvipanda/code/kubespawner/lib/python3.5/site-packages/jupyterhub/handlers/base.py", line 328, in spawn_single_user
        yield gen.with_timeout(timedelta(seconds=self.slow_spawn_timeout), f)
      File "/home/yuvipanda/code/kubespawner/lib/python3.5/site-packages/jupyterhub/user.py", line 261, in spawn
        raise e
      File "/home/yuvipanda/code/kubespawner/lib/python3.5/site-packages/jupyterhub/user.py", line 229, in spawn
        ip_port = yield gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
      File "/home/yuvipanda/code/kubespawner/kubespawner/spawner.py", line 793, in start
      File "/home/yuvipanda/code/kubespawner/kubespawner/spawner.py", line 674, in get_pod_manifest
        env=self.get_env(),
      File "/home/yuvipanda/code/kubespawner/kubespawner/spawner.py", line 871, in get_env
        'JPY_COOKIE_NAME': self.server.cookie_name,
    AttributeError: 'KubeSpawner' object has no attribute 'server'
```

Analect:
@yuvipanda @minrk 
I think [this](https://github.com/jupyterhub/kubespawner/pull/83/commits/a7cadf382327a62b899e503f7c5fdab1e5672114) change that I needed to make (remove the `user` part) might be the cause of the incompatibility between 0.7.2 and 0.8.x.

minrk:
Ah, right. Removing that `.get_env` code altogether would fix the 0.7-0.8 compatibility, and it should indeed be removed. That argues, for me, for removing it in this PR. It's not related to JupyterHub version, but instead related to supporting old builds of jupyter/docker-stacks.

Analect:
Are you happy for me to remove this whole section then?

https://github.com/Analect/kubespawner/blob/06ec33cefca72ad9d0e11768c578a72cf36e967d/kubespawner/spawner.py#L865-L876

yuvipanda:
Yup, that sounds good @Analect. Can you also write comments next to the getattrs saying 'remove when no longer supporting 0.7' so we can easily grep for them? otherwise this is good to go!

Analect:
@yuvipanda @minrk 
Ahead of me making changes to this PR to remove the `get_env`, I updated a separate branch with these commits.
First to comment-out `get_env` ... 
https://github.com/Analect/kubespawner/commit/cf1f2ce5412e2a5c6b49cf06abf36e1717135dba
... then to remove other references to `env`
https://github.com/Analect/kubespawner/commit/e5d22157e27919cd43acb42438448c6b563968bf

The net effect is that a new named-server is now erroring on spawn with:

`JUPYTERHUB_API_TOKEN env is required to run jupyterhub-singleuser. Did you launch it manually?`

 ... so this obviously breaks something.

Are those commits above not the appropriate approach?

yuvipanda:
@Analect you shouldn't remove references to env - just remove the get_env() override in kubespawner. The base spawner instance still sets environment variables that are required.

Analect:
@yuvipanda  ... it's not fully clear to me ... perhaps I can ask another way:

Should I just get rid of `env = super(KubeSpawner, self).get_env()` 
within [this block](https://github.com/Analect/kubespawner/blob/handle-named-servers/kubespawner/spawner.py#L865-L876) below and leave the rest? .... or is the `env.update` dictionary that needs to get removed?

```
def get_env(self):
        # HACK: This is deprecated, and should be removed soon.
        # We set these to be compatible with DockerSpawner and earlie KubeSpawner
        env = super(KubeSpawner, self).get_env()
        env.update({
            'JPY_USER': self.user.name,
            'JPY_COOKIE_NAME': self.server.cookie_name,
            'JPY_BASE_URL': self.server.base_url,
            'JPY_HUB_PREFIX': self.hub.server.base_url,
            'JPY_HUB_API_URL': self.accessible_hub_api_url
        })
        return env
```
What about [`env=self.get_env()`](https://github.com/Analect/kubespawner/blob/handle-named-servers/kubespawner/spawner.py#L674) within the `make_pod` block?  Should that remain intact?

yuvipanda:
Just get rid of the whole 'get_env' function in KubeSpawner. Since it inherits from 'Spawner', which still has the 'get_env', all references to get_env will work. It'll just stop being customized by kubespawner.

Analect:
@yuvipanda ... with the latest commit to remove the `get_env` function, everything tests well against jupyterhub 0.8.0, however, I can't get things working with `zero-to-jupyterhub` against 0.7.2 of the hub.  

In the image for the 0.7.2 hub, I've tried with the `v0.6.0` tagged version of kubespawner installed as well as using master as per this PR#83.  In both cases, I'm getting errors from the hub that look like this: https://gist.github.com/Analect/e4115afe064e16d6cd7bbea471322da1

As you know, this PR is intended to work in conjunction with this [other PR](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/206) in the `zero-to-jupyterhub` repo that extends the naming of server pods to include `named-server` names.  Within the `jupyterhub_config.py` in the `zero-to-jupyterhub` repo, there are some settings ref `hub_connect_ip`, `hub_connect_port` and `concurrent_spawn_limit` that are causing some problems with running these named-server PRs against version 0.7.2 of the server.  I'm not altogether sure if this is as a consequence of removing the `get_env` function in kubespawner .. I don't think so.

The objective here is to enhance these repos to run against 0.8.0 of the hub with the named-server capability built-out ... but for this setup to also work against 0.7.2 of the hub where just default servers will still work.

I'm a bit stumped in terms of how to address this outstanding problem.

minrk:
That error looks like it could be the wrong version of a database. Since you are testing multiple versions of jupyterhub, try deleting the hub database or testing with in-memory sqlite to make sure that it's not getting confused:

```yaml
hub:
  db:
    type: sqlite-memory
```

Analect:
@minrk thanks ... that got me a bit further ... by using the sqlite-memory .. since with the set-up on kubernetes ... a persistent-volume `hub-db-dir` gets created (but not unique to a hub version) ... and presumably my earlier problems related to chopping between v0.8 and v0.7 of the hub ...  but now this:
```
[W 2017-10-04 22:52:48.530 JupyterHub configurable:168] Config option `concurrent_spawn_limit` not recognized by `JupyterHub`.
[W 2017-10-04 22:52:48.531 JupyterHub configurable:168] Config option `hub_connect_port` not recognized by `JupyterHub`.  Did you mean `hub_port`?
[W 2017-10-04 22:52:48.717 JupyterHub configurable:168] Config option `hub_connect_ip` not recognized by `JupyterHub`.  Did you mean `hub_ip`?
[I 2017-10-04 22:52:48.719 JupyterHub app:720] Loading cookie_secret from env[JPY_COOKIE_SECRET]
[I 2017-10-04 22:52:48.826 JupyterHub app:892] Not using whitelist. Any authenticated user will be allowed.
[I 2017-10-04 22:52:48.880 JupyterHub reflector:94] watching for pods with label selector component=singleuser-server,heritage=jupyterhub in namespace default
[I 2017-10-04 22:52:48.883 JupyterHub app:1453] Hub API listening on http://0.0.0.0:8081/hub/
[E 2017-10-04 22:52:48.888 JupyterHub app:1139] Proxy appears to be running at http://10.7.254.164:80/, but I can't access it (HTTP 599: Failed to connect to 127.0.0.1 port 81: Connection refused)
    Is something else using http://10.7.254.164:80/?
```
As part of me getting the named-server stuff working against v0.8x of the hub, one of the settings required in jupyterhub_config.py was to set `c.ConfigurableHTTPProxy.should_start = False` ... I'm just wondering if that might be contributing to this problem in some way in accessing the proxy above ... something I've not seen up until now.

yuvipanda:
As of #87 master of kubespawner now requires 0.8 of JupyterHub, so we can simplify this.

yuvipanda:
I merged master into this, and removed JupyterHub 0.7 compatibility shims. Normal single user testing seems to work, let me try to create named servers...

yuvipanda:
Alright, I can confirm this works! Thank you for your patience, @Analect! I'm going to merge this now!

Analect:
Thanks @yuvipanda ... could you also merge the related https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/206

PR #84, opened 2017-09-21
Allow specifying service accounts for launched pods

### LINKED ISSUE #75: Add support for setting singleuser service account
Currently we disable all service account mounting from inside the pod.

We should allow people to:

1. Specify that the default serviceaccount needs to be mounted
2. Specify that an arbitrary service account should be created and then mounted.

The default would still be no sa, but would be something you can override. This would be great for people who wanna launch helm charts from inside the notebook.

yuvipanda:
Still launching with hack that prevents service account mounting
by default

Fixes #75

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -7,6 +7,8 @@ from kubernetes.client.models.v1_pod_spec import V1PodSpec
 from kubernetes.client.models.v1_object_meta import V1ObjectMeta
 from kubernetes.client.models.v1_pod_security_context import V1PodSecurityContext
 from kubernetes.client.models.v1_local_object_reference import V1LocalObjectReference
+from kubernetes.client.models.v1_volume import V1Volume
+from kubernetes.client.models.v1_volume_mount import V1VolumeMount
 
 from kubernetes.client.models.v1_container import V1Container
 from kubernetes.client.models.v1_security_context import V1SecurityContext
@@ -40,6 +42,7 @@ def make_pod(
     mem_guarantee,
     lifecycle_hooks,
     init_containers,
+    service_account,
 ):
     """
     Make a k8s pod specification for running a user notebook.
@@ -106,6 +109,8 @@ def make_pod(
         Dictionary of lifecycle hooks
       - init_containers:
         List of initialization containers belonging to the pod.
+      - service_account:
+        Service account to mount on the pod. None disables mounting
     """
 
     pod = V1Pod()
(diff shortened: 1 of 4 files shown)


kubespawner/objects.py
```diff
@@ -150,6 +155,28 @@ def make_pod(
...
+        # This makes sure that we don"t accidentally give access to the whole
+        # kubernetes API to the users in the spawned pods.
+        # Note: We don't simply use `automountServiceAccountToken` here since we wanna be compatible
+        # with older kubernetes versions too for now.
+        hack_volume = V1Volume()
+        hack_volume.name =  "no-api-access-please"
+        hack_volume.empty_dir = {}
+        hack_volumes = [hack_volume]
+
+        hack_volume_mount = V1VolumeMount()
```
foxish: This way of locking down access is rather odd. Why would the user want to disable API access this way? Ideally, I'd expect that the pod takes the default service account, and the user uses RBAC to lock down that service account.

yuvipanda:
This was written pre-RBAC, and this was the only way to prevent service
account tokens with full cluster access from being available to users. You
can set a serviceAccountName of None now, but this emptydir mounting is
still necessary for older k8s versions.

The hub pod is usually in the same namespace as the spawned pods, so
depending on rbac to be 'properly' set up complicates things a little bit.
I do agree that is the better long term solution however.

foxish:
Hmm, if we want just 1.6+ support, we can use the `automountServiceAccountToken: false` field that was added. (https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/)

foxish:
@yuvipanda, is there a defined support window for kubespawner?

kubespawner/objects.py
vishh: +1. In some cases having access to specific namespaces is useful to trigger other Operators for example.

yuvipanda:
@foxish there isn't a defined support window, but my general incliniation is to make sure we support at least the last one year worth of releases, and not break backwards compatability without a strong compelling reason.

kubespawner/objects.py
yuvipanda: Right, I agree. This patch lets you disable this hack and set `c.KubeSpawner.singleuser_service_account = "default"` to have the service account be mounted. Using `automountServiceAccountToken` would make kubespawner be incompatible with pre-1.6 kubernetes versions, and since:

1. This hack is functionally still equivalent
2. There is nothing else that forces us to be 1.6+

I'd rather stick with this for now. However, if / when either of those conditions are false, am happy to bump required version up to 1.6 and get rid of this hack.

Issue #85, opened 2017-09-29
periodic reflector health check


minrk:
We've seen a few cases where the reflector started missing events. We should probably have a periodic re-sync to make sure that the reflector is alive and well and in-sync with Kubernetes. While #81 fixes one known race condition, I suspect that there are other ways for events to be missed.

yuvipanda:
@minrk how do you think we should make this happen now? Also do you think we still need to do this despite the resourceversion patch you landed?

minrk:
I think this might still be worthwhile, but I have a lot more confidence after the resourceversion change, so I would call this a low priority until we see another potential symptom of missed events.

As we've seen with `Proxy.check_routes`, periodic sanity checks in an async world are nontrivial.

minrk:
#174 adds timeouts, which ought to be good enough

Issue #68, opened 2017-07-28
Kubespawner fails to spawn singleuser POD


jbmarcille:
I have been testing the kubespawner for a few weeks now and recently switched to the current master branch. Since then, I am unable to spawn the singleuser POD while this worked with a previous version of the kubespawner with Jupyterhub 0.7.2 (bare metal cluster with Kubernetes 1.5.2).

The working kubespawner version used: **_httpclient_** + **_k8s_url_**
while the new version uses the kubernetes python-client **_CoreV1Api_** directly.

The logs showed that the claim request actually works:

`header: Content-Type header: Date header: Transfer-Encoding send: b'POST /api/v1/namespaces/notebook/persistentvolumeclaims HTTP/1.1\r\nHost: 10.10.10.1\r\nAccept-Encoding: identity\r\nContent-Length: 368\r\nAccept: application/json\r\nauthorization: bearer eyJhbGciOiJSUzI1NiIsInR5cCI6Ik...\r\nContent-Type: application/json\r\nUser-Agent: Swagger-Codegen/1.0.0-snapshot/python\r\n\r\n'
send: b'{"spec": {"resources": {"requests": {"storage": "1Gi"}}, "accessModes": ["ReadWriteMany"]}, "apiVersion": "v1", "metadata": {"labels": {"heritage": "jupyterhub", "hub.jupyter.org/username": "toto", "app": "jupyterhub"}, "name": "pvcjupyterhub-singleuser", "annotations": {"volume.beta.kubernetes.io/storage-class": "glusterfs"}}, "kind": "PersistentVolumeClaim"}'
reply: 'HTTP/1.1 409 Conflict\r\n'`

The code I added to retrieve the jupyterhub accessible IP from the service name also works:

`header: Content-Type header: Date header: Content-Length send: b'GET /api/v1/namespaces/notebook/services/jupyterhub?exact=True&export=False HTTP/1.1\r\nHost: 10.10.10.1\r\nAccept-Encoding: identity\r\nAccept: application/json\r\nauthorization: bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9...\r\nContent-Type: application/json\r\nUser-Agent: Swagger-Codegen/1.0.0-snapshot/python\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'`

The request to kubernetes to create the POD times out:

[...]

Issue #65, opened 2017-07-24
"stop my server" results in hung docker instance if user clicks on "my server" before it is done


zogren:
We are testing this on gcloud:

If user goes to 
/hub/home
and clicks "stop my server" and waits they eventually see "my server" and when they click on that it does work
BUT
if the user clicks away and clicks on my server while server is stopping, they get into a state where kubectl says the image with the given username exists and thus a new one can not be started but JupyterHub seems to think none exist so "start my server" shows up on /hub/home and when user clicks on it  they get in an infinite redirect loop between
/hub/user/<userid>
and
/user/<userid>
The hub logs does not show any errors but does show the redirect loop of 302 responses

yuvipanda:
Heya! Thank you for reporting it!

I think this should be mostly fixed in the #64 and JupyterHub 0.8. I'll verify and close after it is!

zogren:
I see a related issue when stress testing where if spinning up a single user instance takes too long things get out of sync and it ends up in same state (instance exists but hub does not think it does so user cant log in)
Do you know if fix in #64 also helps fix this?

zogren:
Switched to your branch an still get issue. It seems to happen if DB entry for node disappears or never gets created but docker instance with that name did get created. Fix could be to delete on create if it already exists or something like that?

zogren:
Guess I should wait till you merge. When I try to use your branch I get 
"The 'spawner_class' trait of <jupyterhub.app.JupyterHub object at 0x7f5f61eab940> instance must be a type, but 'kubespawner.KubeSpawner' could not be imported"
but it looks like I also need to add something new to my helm chart?

yuvipanda:
I think we fixed this in https://github.com/jupyterhub/jupyterhub/pull/1356.

yuvipanda:
And it seems to work with master.

minrk:
Closing as this may be fixed by #174 (or other fixes since then)

PR #88, opened 2017-10-12
Add timeout handling to spawner and reflector (WIP)

gweis:
Hi,

I had some issues with the reflector thread hanging after a while. It stops receiving watch events from the k8s apiserver due to a proxy/load balancer in front of the k8s apiserver. (the first commit addresses this)

The 2nd commit is a works for me solution. The problem was that if jupyterhub decides that it waited long enough for the spawner to start/stop, then it just exits with an exception, but the actual kubespawner instance never leaves the while loop. There are probably better ways to handle this, and certainly better exceptions to throw. So I am rather looking for some suggestions to improve this.

Also, I didn't clean up no longer needed methods. e.g. _list_and_update is no longer used, as watch without a resource_version returns the full current state anyway.

Regards,

Gerhard

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -1,12 +1,18 @@
+# specifically use concurrent.futures for threadsafety
+# asyncio Futures cannot be used across threads
+from concurrent.futures import Future
+
 import time
 import threading
 
 from traitlets.config import LoggingConfigurable
 from traitlets import Any, Dict, Unicode, Int
-from kubernetes import client, config, watch
+from kubernetes import config, watch
+from tornado.ioloop import IOLoop
 # This is kinda implementation specific, but we need to know it
 from urllib3.exceptions import ReadTimeoutError
 
+from .clients import shared_client
 
 class NamespacedResourceReflector(LoggingConfigurable):
     """
@@ -101,10 +107,12 @@ class NamespacedResourceReflector(LoggingConfigurable):
             config.load_incluster_config()
         except config.ConfigException:
             config.load_kube_config()
-        self.api = getattr(client, self.api_group_name)()
+        self.api = shared_client(self.api_group_name)
 
         # FIXME: Protect against malicious labels?
         self.label_selector = ','.join(['{}={}'.format(k, v) for k, v in self.labels.items()])
+
+        self.first_load_future = Future()
 
         self.start()
 
@@ -161,8 +169,14 @@ class NamespacedResourceReflector(LoggingConfigurable):
             self.log.info("watching for %s with label selector %s in namespace %s", self.kind, self.label_selector, self.namespace)
             w = watch.Watch()
             try:
+                resource_version = self._list_and_update()
+                if not self.first_load_future.done():
+                    # signal that we've loaded our initial data
+                    self.first_load_future.set_result(None)
+
                 # in case of timeout_seconds, the w.stream just exits (no exception thrown)
                 # -> we stop the watcher and start a new one
+
                 for ev in w.stream(
                         getattr(self.api, self.list_method_name),
                         **watch_args
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,5 @@ script:
 
 matrix:
   fast_finish: true
+  allow_failures:
+    - python: nightly
--- a/SETUP.md
+++ b/SETUP.md
@@ -9,63 +9,52 @@ the container and redeploy.
 There is an easier way, with minikube and some networking tricks. Only tested on Linux
 at this time, but should work for OS X too.
 
-1. Install [minikube](http://kubernetes.io/docs/getting-started-guides/minikube/). Use the
-   [VirtualBox](https://virtualbox.org) provider. This will set up a kubernetes cluster inside
-   a VM on your machine. It'll also setup `kubectl` on your host machine to interact with
-   the kubernetes cluster, and a `~/.kube/config` file with credentials for connecting to this
-   cluster.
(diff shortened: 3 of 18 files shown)


gweis:
These patches may address issues #85, #68 and may help with #65

yuvipanda:
Thank you very much for the patch, @gweis!!! 

I'm currently trying to land #64, which refactors some related code. Do you mind if I land that first before merging this?

gweis:
@yuvipanda. 

Yes I saw that ingress proxy code :). I have no problems with landing that one first. 

I created this pull request, to show my solution to some of the issues here and I was also thinking that some of the code in this patch may apply in general for the various "kubernetes resource reflectors".

gweis:
@yuvipanda.

congrats in finishing the kube proxy implementation. I have rebased this pull request. Would you mind reviewing it? (when time permits). 

So far it is working for me, but I think it would be good to clean it up, and/or squash it.

tothandras:
@yuvipanda This would be so great! Any updates on this PR?

kubespawner/reflector.py
```diff
@@ -140,20 +173,34 @@ def _watch_and_update(self):
...
+                        # only set resources_version if new resource_version is not None
+                        watch_args['resource_version'] = resource.metadata.resource_version
+                        self.log.info('Watch Event: {} - {}'.format(ev['type'], resource.metadata.name))
                     if ev['type'] == 'DELETED':
                         # This is an atomic delete operation on the dictionary!
                         self.resources.pop(resource.metadata.name, None)
                     else:
                         # This is an atomic operation on the dictionary!
                         self.resources[resource.metadata.name] = resource
+                self.log.debug('Watch Stream timeout')
```
minrk: Probably good to use `.warning` here and below, since these are misbehaviors

kubespawner/reflector.py
```diff
@@ -140,20 +173,34 @@ def _watch_and_update(self):
...
                 ):
                     cur_delay = 0.1
                     resource = ev['object']
+                    if not resource.metadata.resource_version:
+                        # reset resources_version if we don't have one
+                        del watch_args['resource_version']
+                    elif (watch_args.get('resource_version') != resource.metadata.resource_version):
+                        # only set resources_version if new resource_version is not None
+                        watch_args['resource_version'] = resource.metadata.resource_version
+                        self.log.info('Watch Event: {} - {}'.format(ev['type'], resource.metadata.name))
```
minrk: this log probably shouldn't be sensitive to whether resource version is set. Maybe it's at the wrong indentation level?

kubespawner/reflector.py
```diff
@@ -132,6 +155,16 @@ def _watch_and_update(self):
         update' cycle on them), we should be ok!
         """
         cur_delay = 0.1
+        watch_args = {
+            'namespace': self.namespace,
+            'label_selector': self.label_selector,
+        }
+        if self.request_timeout:
+            watch_args['_request_timeout'] = self.request_timeout
```
minrk: What do we lose by not using this private argument? Will it never timeout, or will it just not be configurable by us? If it's not configurable, I'd rather respect the private API and leave it out.

kubespawner/reflector.py
```diff
@@ -140,20 +173,34 @@ def _watch_and_update(self):
...
+                        **watch_args
                 ):
                     cur_delay = 0.1
                     resource = ev['object']
+                    if not resource.metadata.resource_version:
+                        # reset resources_version if we don't have one
+                        del watch_args['resource_version']
+                    elif (watch_args.get('resource_version') != resource.metadata.resource_version):
+                        # only set resources_version if new resource_version is not None
+                        watch_args['resource_version'] = resource.metadata.resource_version
```
minrk: I don't think this resource_version update is accomplishing its goal. We are starting the list fresh with `._list_and_update` on every restart, and must use that resource_version in order to avoid lost events. Recording the resource_version here could allow us to skip that step on subsequent reconnects, but we would need to be very careful to ensure that we *don't* call `_list_and_update` on resume. Instead, it's probably simpler to drop this resource_version tracking and start fresh with `_list_and_update` each go. This will restart rarely, so the millisecond-level cost of a fresh check is probably worth the simpler logic.

kubespawner/reflector.py
gweis: In my case, 'watch stream timeout' is expected behaviour. The load balanced route may disappear if it isn't kept alive. A warning is ok too, this line is merely there for checking that reconnect to the api server actually happens and works.

kubespawner/reflector.py
gweis: Only a trace / debug log. Just there to see that it actually happens. I'll cleanup and rebase the pull request.

kubespawner/reflector.py
gweis: This private `_request_timeout` is about the underlying urllib3 timeout. I have not found another way to set that. Could probably be raised as an issues with the swagger code generator.

kubespawner/reflector.py
gweis: I completely skipped the `._list_and_update`. An initial watch without resource version retrieves the full list. After reading the Kubernetes API docs again, I found that it has become more specific about this. Watch only returns the full list since the beginning of history, which may not return the full current list. I'll bring that back in the cleanup

gweis:
I'll redo this and rebase to current master.

kubespawner/reflector.py
gweis: here is the "documentation" I have found. https://github.com/kubernetes-client/python/blob/master/kubernetes/client/rest.py#L125

PR #89, opened 2017-10-16
Extract method to build labels for notebook pod

ktong:
Extract two methods for reusing:
1. _build_common_labels to build common labels for both notebook pod and pvc.
2. _build_pod_labels to build label for notebook pod and ensure including labels required by pod_reflector.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -634,6 +634,30 @@ class KubeSpawner(Spawner):
         else:
             return src
 
+    def _build_common_labels(self, extra_labels):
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thank you for the patch!

Is there a specific reason you want this to be a method, instead of just a common_labels dictionary that's a class attribute / triatlet that can be expanded with more keys in pod / pvc creation?

ktong:
First, it’s just refactoring so I would not introduce another triatlet. Second, labels can be customized by extra_labels and extra_storage_labels, it probably does not need another configurable attribution

PR #90, opened 2017-10-18
Use default parameters in make_pod to simple tests

ktong:
It's very hard to add parameter to method make_pod because all parameters are required, which means you add add new parameter into all tests of make_pod (Most of them are None, [] or {}). set default value for non-mandatory extremely simplify test code.


yuvipanda:
Thank you for the patch! Apologies about the delay :)

PR #91, opened 2017-10-20
Extra config

ktong:
Add 3 extra configuration to customize extension of notebook pod


ktong:
Includes #90 and replace #53

kubespawner/spawner.py
```diff
@@ -605,6 +605,50 @@ def _hub_connect_port_default(self):
...
+        None,
+        config=True,
+        help="""
+        Extra configuration (e.g. envFrom) for notebook container which is not covered by other attributes.
+        
+        This dict will be directly merge into `container` of notebook server,
+        so you should use the same structure. Each item in the dict is field of container configuration
+        which follows spec at https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#container-v1-core.
+        
+        The key cloud be either camelcase word (used by Kubernetes yaml, e.g. envFrom)
```
yuvipanda: I think this is 'The keys could' rather than cloud?

Can you also provide a concrete example for this in the docstring?

kubespawner/spawner.py
```diff
@@ -605,6 +605,50 @@ def _hub_connect_port_default(self):
...
+        None,
+        config=True,
+        help="""
+        Extra configuration (e.g. tolerations) for the pod which is not covered by other attributes.
+        
+        This dict will be directly merge into pod,so you should use the same structure. 
+        Each item in the dict is field of pod configuration
+        which follows spec at https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#podspec-v1-core.
+
+        The key cloud be either camelcase word (used by Kubernetes yaml, e.g. dnsPolicy)
```
yuvipanda: Same as above.

Issue #95, opened 2017-11-06
Never orphan pods


yuvipanda:
Right now, if we start a pod but it doesn't successfully launch in the launch timeout, we just sort of let it exist forever, in an orphaned state. This is very problematic!

We should make sure we clean up after ourselves if we do not launch in time.

yuvipanda:
You can repro this by doing something like having a wrong image name in the config or an image that takes longer to download than the configured timeout. It will make it impossible for the user to launch until the pod is manually deleted.

consideRatio:
I think this is a resolved issue, I have not seen us orphan pods for a long time.

PR #98, opened 2017-11-17
Fix poll() when pod exists but is not running

cmoscardi:
`poll()` should only return 1 if the pod is dead-as-a-doornail,
based on the current JupyterHub interface. However, `is_pod_running()`
only returns True if the pod is in a `Running` state, not (for example)
`ContainerCreating`, which is technically a not-dead state. So we remove it.

On the other hand, the reflector's `_watch_and_update()` function, which
runs as its own thread, and keeps `self.pods` up-to-date. In particular,
as soon as kubernetes reports a `DELETED` event, `watch_and_update()`
removes that pod information from the list of active pods (see
reflector.py:L112). So we're safe to have /just/ the
`if data is not None` check.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -835,7 +835,7 @@ class KubeSpawner(Spawner):
         JupyterHub expects.
         """
         data = self.pod_reflector.pods.get(self.pod_name, None)
-        if data is not None and self.is_pod_running(data):
+        if data is not None:
             return None
 
         return 1


cmoscardi:
Fixes https://github.com/jupyterhub/kubespawner/issues/95

cmoscardi:
Last, for posterity, documenting some wisdom  from @minrk 

Calling stop is conditional on the return of poll here.

Min RK @minrk 14:31
But KubeSpawner.poll checks if a pod is running here.
Whereas JupyterHub's expectation would more accurately be described as "has the pod exited", as opposed to is it actively running.
So there's a disagreement about what poll means for pods that have not yet started.
To satisfy JupyterHub's expectation, poll() should probably return None if the pod exists at all, because that's the condition under which stop will be called, which is needed to cleanup the pod.

Analogy (from which the poll expectations derive): a process has a PID immediately, even if the program has not really started yet. As long as that PID is allocated, .poll() reports that the resource is allocated, even if it's in preflight setup. KubeSpawner should match that.

And perhaps there is a case to be made for a dedicated 'cleanup after yourself' call that's not the same as if poll() is not None: stop(), which seems to trip up Spawners other than local processes.

yuvipanda:
@minrk should 'start' return as soon as the pod exists or as soon as it is in 'Running' state? ImagePull and disk setup can take more than the 30s that's the default timeout between start returning and the hub expecting the notebook to be running.

Perhaps the way out is to also change the check in start to not call poll but actively check if the pod is running directly?

Also thanks a lot for taking this on, @cmoscardi!

cmoscardi:
Ah @yuvipanda I was actually typing up the below when you replied... in the kubespawner's `start()` function currently, `is_pod_running` is already directly checked.

As it stands now, `is_pod_running()` needs to remain the way it is for startup purposes - when it's called [in the start() function](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L892), it's used to see if the pod is actually in a fully alive/accessible sate.

yuvipanda:
Ah, cool!

I'll wait for @minrk to chime in since he understands the mechanics better than I do!

minrk:
Thanks! This is the right thing to do for now, and we can try to clean it up a bit on the JupyterHub side.

> should 'start' return as soon as the pod exists or as soon as it is in 'Running' state?

Running. I think KubeSpawner is doing the right thing in `start`, in that JupyterHub expects `start` to return when the Spawner believes the process to have actually started. Although it shouldn't continue forever, which is what I think it is doing now.

What probably makes sense is to avoid that ambiguity over what `.poll()` means during an incomplete start, and instead add an explicit `cleanup_failed_start()` method that is called on timeout, where the *default* could be the same as above, but Spawners would be able to override it as appropriate.

PR #100, opened 2017-11-23
Fix typo in make_pod

### LINKED ISSUE #99: extra_containes instead of extra_containers in make_pod call.
Line:

* https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L771

has:

```
            extra_containes=self.singleuser_extra_containers
```

That will surely die as keyword argument of ``make_pod`` is ``extra_containers``.

yuvipanda:
Fixes #99


minrk:
I was just about to make this PR!

Issue #93, opened 2017-10-26
Request GPU


zonca:
I cannot find a way to request a GPU via `kubespawner`, the equivalend `kubectl` specification is:

```
spec:                                                                                                                               
  containers:                                                                                                                       
    - image: XXXX                                                                             
      resources:                                                                                                                    
        limits:                                                                                                                     
          alpha.kubernetes.io/nvidia-gpu: 1 # requesting 1 GPU 
```

yuvipanda:
Thanks to @foxish there's now a 'extra_resource_limits' and 'extra_resource_guarantees' you can use for this!

naeemkhan12:
how can i specify 'extra_resource_limits' and 'extra_resource_guarantees'  in values.yaml file. for nvidia gpu
```
cpu:
    limit:
    guarantee:
  memory:
    limit:
    guarantee: 1G
  extraResource:
```

twbrandon7:
@naeemkhan12 Just add these parameters in extraConfig.
For example:
```
hub:
  cookieSecret: "<COOKIE>"
  db:
    type: sqlite-pvc
  extraConfig: |-
    c.KubeSpawner.extra_resource_limits = {"nvidia.com/gpu": "1"}
    c.KubeSpawner.extra_resource_guarantees = {"nvidia.com/gpu": "1"}
```

PR #102, opened 2017-11-27
Added new methods for specifying other resource limits/requests

foxish:
Added some additional config options that can be overridden from a form.
Tested against a form in https://gist.github.com/foxish/3407150729b106a05811d672d905c27b

@yuvipanda @liyinan926 PTAL

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -668,6 +668,28 @@ class KubeSpawner(Spawner):
         """
     )
 
+    arbitrary_resource_guarantee = Dict(
+        {},
+        config=True,
+        help="""
+        The dictionary used to request arbitrary resources.
+        Default is None and means no additional resources are requested.
+        For example, to request 3 Nvidia GPUs
+            `{"nvidia.com/gpu": "3"}`
+        """
+    )
+
+    arbitrary_resource_limits = Dict(
+        {},
+        config=True,
+        help="""
+        The dictionary used to limit arbitrary resources.
+        Default is None and means no additional resources are limited.
+        For example, to add a limit of 3 Nvidia GPUs
+            `{"nvidia.com/gpu": "3"}`
+        """
+    )
+
     def _expand_user_properties(self, template):
         # Make sure username and servername match the restrictions for DNS labels
         safe_chars = set(string.ascii_lowercase + string.digits)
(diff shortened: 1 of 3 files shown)


kubespawner/spawner.py
```diff
@@ -771,6 +795,24 @@ def get_pod_manifest(self):
             extra_containers=self.singleuser_extra_containers
         )
 
+    def get_cpu_limit(self):
```
minrk: Why all these getter methods? One of the nice things about Python is that you don't need getter methods to get all of the benefits of getters. Subclasses can define properties or extend the traitlets here without needing to add any methods on the base class.

foxish: I was under the impression that we need these to be present to facilitate overriding as in https://gist.github.com/foxish/3407150729b106a05811d672d905c27b. If we can performing overriding of that kind without really adding those getters, that would be ideal. Can you elaborate a bit more on the mechanism by which one can do that?

yuvipanda: @foxish you can do so by overriding start like https://github.com/berkeley-dsep-infra/datahub/blob/staging/datahub/config.yaml#L12

yuvipanda:
@foxish thank you for the patch! I believe this fixes #93 too! Can you address @minrk's comments and then we can merge this?

minrk:
👍 to merge without the getters. You can also use a property:

```python
class MySpawner(KubeSpawner):
    @property
    def mem_limit(self):
        return construct_mem_limit(self.user)
```

which is the same as having a getter, it just doesn't need to be defined as a function on the parent class for the child to have the extra logic.

foxish:
@minrk @yuvipanda PTAL, removed the getters. Thanks for the review.

yuvipanda:
Can you also add a test for this, @foxish? It's in test_objects.py I believe.

tests/test_objects.py
```diff
@@ -631,6 +631,121 @@ def test_make_pod_with_extra_containers():
         "apiVersion": "v1"
     }
 
+def test_make_pod_with_extra_containers():
+    """
+    Test specification of a pod with initContainers
```
yuvipanda: this description is wrong I think?

yuvipanda:
This lgtm, I'll merge once the tests pass!

PR #111, opened 2017-12-07
Allow easily augmenting the pod config before launching

yuvipanda:
This makes it easy for people to configure this without
having to inherit from KubeSpawner and change it.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -454,7 +454,7 @@ class KubeSpawner(Spawner):
 
         This can be a coroutine if necessary. When set to none, no augmenting is done.
 
-        This is very useful if you want to modify the pod being launched programattically.
+        This is very useful if you want to modify the pod being launched dynamically.
         Note that the spawner object can change between versions of KubeSpawner and JupyterHub,
         so be careful relying on this!
         """
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -19,7 +19,7 @@ from tornado import gen
 from tornado.ioloop import IOLoop
 from tornado.concurrent import run_on_executor
 from traitlets.config import SingletonConfigurable
-from traitlets import Type, Unicode, List, Integer, Union, Dict, Bool, Any, Callable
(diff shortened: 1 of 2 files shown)


kubespawner/spawner.py
```diff
@@ -438,6 +438,28 @@ def _hub_connect_port_default(self):
...
+        Expects a callable that takes two parameters:
+
+           1. The spawner object that is doing the spawning
+           2. The Pod object that is to be launched
+
+        You should modify the Pod object and return it.
+
+        This can be a coroutine if necessary. When set to none, no augmenting is done.
+
+        This is very useful if you want to modify the pod being launched programattically.
```
foxish: nit: typo in "programattically"

kubespawner/spawner.py
```diff
@@ -909,6 +931,8 @@ def start(self):
         # FIXME: Have better / cleaner retry logic!
         retry_times = 4
         pod = yield self.get_pod_manifest()
+        if self.augment_pod:
+            pod = self.augment_pod(self, pod)
```
foxish: Why also pass `self`?

foxish:
This is a good step towards more flexible parameterizing in the spawner. Thanks @yuvipanda

kubespawner/spawner.py
yuvipanda: Fixed!

kubespawner/spawner.py
yuvipanda: So the augmenter can access other traitlets, and also call methods to get things like auth_state...

yuvipanda:
np @foxish! Glad to help :)

yuvipanda:
/cc @minrk too.

I welcome some amount of bikeshedding on the name of this thing.

minrk:
> I welcome some amount of bikeshedding on the name of this thing.

There is some convention around calling configurable callables `hooks`, but I don't have a strong feeling.

yuvipanda:
@minrk renamed!

foxish:
LGTM!

PR #112, opened 2017-12-08
fix service account setting

keithcallenberg:
switched service_account_name setting from pod to pod.spec


yuvipanda:
w00t, awesome, ty @keithcallenberg.

I asked @keithcallenberg to add tests and they've graciously obliged :) So will wait for tests before merging!

yuvipanda:
Thank you very much!

PR #115, opened 2018-01-06
[WIP] Allow creating arbitrary kubernetes objects with each pod

yuvipanda:
This is very useful in cases where you want to create other objects (such as namespaces,
RBAC related objects, new service accounts, etc) with each pod. Example use cases are 
spawning a namespace per user for running dask / spark / TF jobs in.

- [ ] Figure out cleanup of objects (probably with ownerReference)
- [ ] How to distribute common patterns of objects that might be created?


yuvipanda:
Note that this isn't talking about spawning each user in their own namespace (that's separate work that we haven't started).

yuvipanda:
note that I don't like this code at all, will have to refactor (and maybe steal inspiration from kubernetes client-go)

foxish:
My apologies, I haven't looked at this yet. Will take a look today. Thanks @yuvipanda

yuvipanda:
No worries, @foxish. I'll try to push some of my code clanup shortly, but the general approach (a callable that can return arbitrary objects to be created) will be the same.

kubespawner/utils.py
```diff
@@ -45,3 +48,82 @@ def validate(self, obj, value):
...
+    method should be 'create', 'delete', 'list' or 'patch'.
+
+    Set namespaced=False for objects that are not in a namespace (like ClusterRoles or
+    Nodes)
+    """
+    if apiversion == 'v1':
+        class_name = 'CoreV1Api'
+    else:
+        parts = apiversion.replace('k8s.io', '').split('/')
+        parts = [''.join([first_upper(p) for p in part.split('.')]) for part in parts]
```
foxish: General approach makes sense, but this seems a bit brittle.

yuvipanda: Yeah I agree! This is sortof prototype quality code now, I need to spend a solid day on this soon.

kubespawner/spawner.py
```diff
@@ -479,6 +479,35 @@ def _hub_connect_port_default(self):
...
+
+          1. The spawner object that is doing the spawning
+
+        It should return a iterable of dictionaries with the following keys:
+
+          1. object - The Kubernetes object to create, instantiated from the kubernetes client library
+          2. kind - The `kind` of `object`. For example, 'Pod', 'Deployment', 'Namespace', etc
+          3. api_version: The `apiVersion` of object to be created.
+          4. namespaced (bool): True (default) if the object is namespaced (like pods), false otherwise
+          5. ifexists: Action to take if object already exists. Possible options are: ignore, patch (default) & error.
```
foxish: do we have a use-case for patch? I feel like it might not do what people expect. Maybe ignore/replace are sufficient?

yuvipanda: How would this work on the long run without patch? If I changed which service account or namespace annotation (for example) gets created, then changing every object created until then is a tedious manual operation...

But in some other cases (like PVCs) patch is probably not what you want.

vpavlin:
FYI if anyone is interested we have a library that does exactly this for us in Open Data Hub

Implementation: https://github.com/opendatahub-io/jupyterhub-singleuser-profiles/blob/master/jupyterhub_singleuser_profiles/service.py

Definition of the "service": https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/jupyterhub-singleuser-profiles-configmap.yaml#L36-L56

Service template: https://github.com/opendatahub-io/odh-manifests/blob/master/jupyterhub/jupyterhub/base/jupyterhub-spark-operator-configmap.yaml

Main part of integration into JH: https://github.com/opendatahub-io/jupyterhub-odh/blob/master/.jupyter/jupyterhub_config.py#L233-L244

consideRatio:
@yuvipanda this is a very old PR by now, and while I find it to be practically useful, I consider it out of scope for KubeSpawner project to maintain I think. What do you think at this point in time?

yuvipanda:
@consideRatio I think it's very much in scope for KubeSpawner - we already create PVC, and for example creating service accounts alongside would be very nice and in scope. Instead of deciding that, I think we should allow this to happen.

However, it's old and I don't have the time / energy to finish it up, so I'm going to close this one. Hope someone else picks this up, especially now that [server-side apply](https://kubernetes.io/blog/2020/04/01/kubernetes-1.18-feature-server-side-apply-beta-2/) seems to be maturing!

PR #125, opened 2018-01-30
Revert "Bump kubespawner version"

yuvipanda:
Reverts jupyterhub/kubespawner#124

As of v4, creating an ApiClient object immediately creates a ThreadPool,
which takes up CPU by merely existing. Since we create a lot of those,
it takes up a ton of CPU and stops things from doing real work.

Should fix https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/461

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='jupyterhub-kubespawner',
-    version='0.8',
+    version='0.7.1',
     install_requires=[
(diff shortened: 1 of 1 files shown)


minrk:
yeah, nightly should generally be `allow_failures: True`. We probably don't need it in this repo.

yuvipanda:
Note that this was the wrong thing to do, https://github.com/jupyterhub/kubespawner/pull/127 is the right thing to do

PR #127, opened 2018-01-30
pin kubernetes client to 3.*

minrk:
Reverts kube upgrade in #114 to deal with https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/461

(#125 reverted kubespawner version, not kubernetes version)

bumps kubespawner version to 0.8.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,5 @@ script:
 
 matrix:
   fast_finish: true
+  allow_failures:
+    - python: nightly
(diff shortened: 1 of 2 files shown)


yuvipanda:
That PR proves it was indeed way past my bedtime :)

yuvipanda:
I think all of #114 needs reverting, since some of the constructor styles had also changed. but I guess this works fine, so we can keep it as is.

minrk:
Since the tests passed, it's okay, right?

PR #128, opened 2018-01-30
share kubernetes client instances

### LINKED ISSUE #126: Share kubernetes client objects
Instantiating a Kubernetes client object instantiates a max-size ThreadPool with kubernetes 4.0. This can get expensive if many clients are instantiated (see #125).

Sharing kubernetes client instances should avoid this

minrk:
avoids creating a kubernetes client for each Spawner, which creates a bunch of threads in kubernetes 4.0.

uses weakref to avoid breaking garbage collection

closes #126

### DIFF
--- a/kubespawner/clients.py
+++ b/kubespawner/clients.py
@@ -1,0 +1,27 @@
+import weakref
+
+import kubernetes.client
+
+_client_cache = {}
+
+
+def shared_client(ClientType, *args, **kwargs):
+    """Return a single shared kubernetes client instance
+
+    A weak reference to the instance is cached,
+    so that concurrent calls to shared_client
+    will all return the same instance until
+    all references to the client are cleared.
+    """
+    kwarg_key = tuple((key, kwargs[key]) for key in sorted(kwargs))
+    cache_key = (ClientType, args, kwarg_key)
+    client = None
+    if cache_key in _client_cache:
+        client = _client_cache[cache_key]()
+
+    if client is None:
+        Client = getattr(kubernetes.client, ClientType)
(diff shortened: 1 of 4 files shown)


yuvipanda:
we should hold off on this until after 0.6 of z2jh.

minrk:
yes, I was planning to cut the release and finish z2jh before merging this

yuvipanda:
@minrk is this good to go or still WIP?

kubespawner/clients.py
```diff
@@ -0,0 +1,27 @@
...
+    A weak reference to the instance is cached,
+    so that concurrent calls to shared_client
+    will all return the same instance until
+    all references to the client are cleared.
+    """
+    kwarg_key = tuple((key, kwargs[key]) for key in sorted(kwargs))
+    cache_key = (ClientType, args, kwarg_key)
+    client = None
+    if cache_key in _client_cache:
+        client = _client_cache[cache_key]()
```
betatim: Why the trailing `()`? I think `_client_cache` contains instances already so this would end up calling `__call__` on that instance.

minrk: `_client_cache` contains weak references to instances, not the instances themselves. This allows them to be garbage collected. The ref must be called in order to return the instance. If it has been garbage collected in the meantime, resolving the weakref will return None.

I've added a comment to clarify what's happening here.

I've

minrk:
Forgot to take off WIP. This should be all set.

betatim:
Merging as travis is happy.

Thanks for explaining the `weakref` thing.

PR #130, opened 2018-02-01
Allow waiting for reflectors to get initial data

minrk:
and ensure `Spawner.poll()` never returns before the first data load, which would return the wrong state for a Spawner that exited while the Hub was down (a pretty rare occurrence).


yuvipanda:
Thanks, @minrk

Issue #133, opened 2018-02-10
RFE: Provide method of setting supplementalGroups in securityContext of pod.


GrahamDumpleton:
Addition of ability to set ``supplementalGroups`` in ``securityContext`` of pod will allow Jupyter Project docker-stacks images to run on OpenShift under default multitenant security model, or plain Kubernetes where full multitenant RBAC/SCC have been configured. One need only set ``supplementalGroups`` to include group ID ``100`` corresponding to ``NB_GID`` of the docker-stack images.

The ``supplementalGroups`` setting will need to allow a list of integers.

```
$ oc explain pod.spec.securityContext.supplementalGroups
FIELD: supplementalGroups <[]integer>

DESCRIPTION:
     A list of groups applied to the first process run in each container, in
     addition to the container's primary GID.  If unspecified, no groups will be
     added to any container.
```

This is different to ``fsGroup``. Under a ``restricted`` SCC of multitenant configuration you have:

```
NAME               PRIV      CAPS      SELINUX     RUNASUSER          FSGROUP     SUPGROUP    PRIORITY   READONLYROOTFS   VOLUMES
restricted         false     []        MustRunAs   MustRunAsRange     MustRunAs   RunAsAny    <none>     false            [configMap downwardAPI emptyDir persistentVolumeClaim projected secret]
```

So ``FSGROUP`` has ``MustRunAs`` which means it is restricted to the group IDs allowed by project:

```
		openshift.io/sa.scc.supplemental-groups=1010020000/10000
		openshift.io/sa.scc.uid-range=1010020000/10000
```

Because ``SUPGROUP`` is set to ``RunAsAny`` you can set supplemental groups to anything, including group ID 100. Because ``fix-permissions`` of docker-stacks images makes files/groups writable by group ID 100, this should be enough to allow images to mostly work with ``restricted`` SCC.

Note that there is still one other change that will be required to docker-stacks images to avoid potential for issues. This relates to lack of passwd/group file entries when run as arbitrary user ID, but that change I have for that will hopefully be acceptable. :-)

willingc:
I think the work in #134 and docker-stacks addresses this issue. I'm going to close but please feel free to ask for it to remain open if needed.

PR #134, opened 2018-02-11
Add support for adding additional supplemental groups.

GrahamDumpleton:
As described in #133.


yuvipanda:
LGTM, can you add a test for this in test_objects and can merge after?

Thank you very much for the patch! Looking forward to it working well in OpenShift!

GrahamDumpleton:
Added a test, but don't have the environment needed to run the tests.

minrk:
Thanks!

PR #143, opened 2018-03-15
set default restartPolicy to Never for user pods

minrk:
so that user pods exiting do not get restarted

needed for self-culling, otherwise a culled pod gets restarted automatically

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -149,6 +149,7 @@ def make_pod(
     )
 
     pod.spec = V1PodSpec(containers=[])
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks, @minrk!

minrk:
Either this didn't have the desired effect, or I applied the wrong version to z2jh, because [this change](https://github.com/jupyterhub/mybinder.org-deploy/pull/502/files#diff-dfb92f6b6dc9cfa9811bbe898f45bf89R62) was still needed. Need to double-check that this affects the pod spec in the right way.

PR #145, opened 2018-03-15
poll returns exitcode if notebook container stops

### LINKED ISSUE #144: "Completed" pods do not trigger shutdown notifications
When a pod enters the "Completed" state, it is stopped. This should trigger the same notifications that a server is stopped that deleting a pod does. It should probably further delete the pod.

This is coming up with the internal-culler behavior, where a pod shuts itself down. However, it doesn't trigger user-shutdown notifications because KubeSpawner incorrectly reports the server as alive and running at this time.

minrk:
This fixes KubeSpawner failing to notice when a notebook container stops by a mechanism other than deleting the pod

closes #144

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -975,8 +975,15 @@ class KubeSpawner(Spawner):
             yield self.pod_reflector.first_load_future
         data = self.pod_reflector.pods.get(self.pod_name, None)
         if data is not None:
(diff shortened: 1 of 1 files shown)


minrk:
This is needed for in-pod self-culling https://github.com/jupyterhub/mybinder.org-deploy/pull/502

yuvipanda:
Thanks, @minrk!

PR #150, opened 2018-03-26
Use EventReflector to watch pod events

clkao:
This provides the underlying k8s pod events required for #141 and #148.


minrk:
Nice!

PR #153, opened 2018-03-30
progress messages for jupyterhub 0.9

### LINKED ISSUE #141: diagnosis message on failed launches
When a pod fails to launch, we should let the user know the causes: resources full, image pull failure, launch failure, pvc mount failure, crashed, etc.  We can probably start with the event messages coming from `kubctl describe pod`

### LINKED ISSUE #148: Show Kubernetes logs to the user while starting the pod
Having a loading spinner is ok:

screen shot 2018-03-19 at 22 38 58

However showing the user what's going on, i.e. by showing K8s events for example, would work like the mirror in the elevators, users won't suffer the wait as they see some punctual progress.

clkao:
With https://github.com/jupyterhub/jupyterhub/pull/1767 the pending_messages are rendered:

not super fancy, but informative enough for now.

Closes #141 
Closes #148

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1206,7 +1206,10 @@ class KubeSpawner(Spawner):
         )
 
         pod = self.pod_reflector.pods[self.pod_name]
-        self.log.debug('pod %s events before launch: %s', self.pod_name, self.events.events)
+
+        self.log.debug('pod %s events before launch: %s',
+            self.pod_name, "\n".join(["%s [%s] %s" % (event.last_timestamp, event.type, event.message) for event in self.events.events]))
+
         # Note: we stop the event watcher once launch is successful, but the reflector
         # will only stop when the next event comes in, likely when it is stopped.
         self.events.stop()
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -28,7 +28,8 @@ from kubespawner.traitlets import Callable
 from kubespawner.utils import Callable
 from kubespawner.objects import make_pod, make_pvc
 from kubespawner.reflector import NamespacedResourceReflector
-
+from queue import Queue
+from async_generator import async_generator, yield_
 
 class PodReflector(NamespacedResourceReflector):
     kind = 'pods'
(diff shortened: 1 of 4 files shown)


clkao:
This is now using the api provided by https://github.com/jupyterhub/jupyterhub/pull/1771

clkao:
Thanks for the review @minrk!

It seems reasonable to handle this generically in jupyterhub, as progress can indeed be called multiple times when multiple opened tabs, or if we decide to inspect events in admin ui.  this is quite spawner-agnostic.

I suppose a static event array appended by the reflector, with individual queue for generator using an index seems to be the simplest way for doing this?  btw that'd be awfully similar to the original implementation before the async progress api ;)

minrk:
The original implementation couldn't handle multiple watchers or notifying of new events, which are the main issues that we need to solve. To avoid creating multiple reflectors (which is by far the simplest option), we would need to:

1. create a static record of the history of past events
2. register a watcher for each call to receive new events
3. yield events, transitioning carefully from the static list to listeners for new events
4. properly destroy the history when done, to avoid filling up local memory

I'll have a stab at doing this on the jupyterhub side. If Spawner progress yielded events from the beginning every time, this would be vastly simpler, which might be my preference.

minrk:
@clkao I opened #169 because Watch creating a new Client for every spawn is still going to be a problem without nipping the extra threads in the bud. Hopefully after that, we should be safe.

kubespawner/spawner.py
```diff
@@ -1086,6 +1096,24 @@ def poll(self):
...
+            len_events = len(self.events.events)
+            if next_event < len_events:
+                for i in range(next_event, len_events):
+                    event = self.events.events[i]
+                    await yield_({
+                        'progress': 50,
+                        'message':  "%s [%s] %s" % (event.last_timestamp, event.type, event.message)
+                    })
+                next_event = len_events
+            await sleep(1)
```
minrk: This while loop needs to end at some point. How do you signal that there will be no more events?

clkao:
@minrk thanks for the review. I added a check to break if the reflector is stopped for the generator.

However it is unclear to me what happens when the event stream connection is closed.  Does the `async with aclosing` in the api handler finalize the call to progress generator?

minrk:
> Does the async with aclosing in the api handler finalize the call to progress generator?

Yes, I believe it does.

minrk:
Let's give this a try. Thanks!

PR #157, opened 2018-04-14
Simplify setup instructions

yuvipanda:
Stolen from work with @mrocklin on the dask-kubernetes
setup instructions.


SETUP.md
```diff
@@ -16,56 +16,39 @@ at this time, but should work for OS X too.
...
-   ```
-   Note that if you had docker installed on your host, this *will* futz with it! You might have to stop
-   the docker daemon before doing it. Restarting the docker daemon should bring it back to working order,
-   however.
-   
-   Then, we can add a static route that routes all pod traffic to the virtual machine, with:
-   ```
-   LINUX:
-   sudo ip route add 172.17.0.0/16 via $(minikube ip)
+    On Linux::
```
willingc: Looks like you have dropped in restructuredText instead of markdown ;-)

yuvipanda:
@willingc great catch, hahahah! I just fixed it.

PR #160, opened 2018-04-17
Provide a nicer default profile template

yuvipanda:
screenshot-2018-4-17 jupyterhub

It looks like this now, and supports a 'description' field.

### DIFF
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -27,3 +27,22 @@ c.JupyterHub.authenticator_class = 'dummyauthenticator.DummyAuthenticator'
 c.KubeSpawner.user_storage_pvc_ensure = False
 
 c.JupyterHub.allow_named_servers = True
+
+c.KubeSpawner.profile_list = [
+    {
+        'display_name': 'Training Env - Python',
+        'default': True,
+        'kubespawner_override': {
+            'singleuser_image_spec': 'training/python:label',
+            'cpu_limit': 0.5,
+        },
+        'description': 'Something description of what is going on here, maybe a <a href="#">link too!</a>'
+    }, {
+        'display_name': 'Training Env - Datascience',
+        'kubespawner_override': {
+            'singleuser_image_spec': 'training/datascience:label',
+            'cpu_limit': 0.2,
+        },
+        'description': 'Something description of how this is different, maybe a <a href="#">link too!</a>'
+    }
+]
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -803,12 +803,35 @@ class KubeSpawner(Spawner):
 
     profile_form_template = Unicode(
         """
-        <label for="profile">Please select a profile to launch</label>
-        <select class="form-control" name="profile" required autofocus>
-            {% for profile in profile_list %}
-            <option {% if profile.default %}selected{% endif %} value="{{ loop.index0 }}">{{ profile.display_name }}</option>
-            {% endfor %}
-        </select>
+        <script>
+        // JupyterHub 0.8 applied form-control indisciminately to all form elements.
+        // Can be removed once we stop supporting JupyterHub 0.8
+        $(document).ready(function() {
+            $('#kubespawner-profiles-list input[type="radio"]').removeClass('form-control');
+        });
+        </script>
+        <style>
+        /* The profile description should not be bold, even though it is inside the <label> tag */
+        #kubespawner-profiles-list label p {
+            font-weight: normal;
+        }
+        </style>
+
+        <div class='form-group' id='kubespawner-profiles-list'>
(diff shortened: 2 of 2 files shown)


consideRatio:
Woooooo @yuvipanda this looks really good and it matters!! I'm really happy about the work you have done on this @gsemet and @yuvipanda, I'll quickly start utilizing this!

jupyterhub_config.py
```diff
@@ -27,3 +27,22 @@
...
+        },
+        'description': 'Something description of what is going on here, maybe a <a href="#">link too!</a>'
+    }, {
+        'display_name': 'Training Env - Datascience',
+        'kubespawner_override': {
+            'singleuser_image_spec': 'training/datascience:label',
+            'cpu_limit': 0.2,
+        },
+        'description': 'Something description of how this is different, maybe a <a href="#">link too!</a>'
+    }
```
consideRatio: I'll go to sleep before I think this through properly, but I'm considering the implications of pre-pulling images in z2jh-k8s. I'll also need to figure out what it means to have a jupyterhub_config.py file within the kubespawner project.

minrk: It might be good to move this config file to an `examples` directory with a README.

kubespawner/spawner.py
```diff
@@ -803,12 +803,35 @@ def _hub_connect_port_default(self):
...
+        <style>
+        /* The profile description should not be bold, even though it is inside the <label> tag */
+        #kubespawner-profiles-list label p {
+            font-weight: normal;
+        }
+        </style>
+
+        <div class='form-group' id='kubespawner-profiles-list'>
+        {% for profile in profile_list %}
+        <label for='profile-item-{{ loop.index0 }}' class='form-control input-group'>
```
minrk: I don't thin it makes sense for the item to be labeled to be *inside* its own label. Maybe the label should only be around the `profile.display_name`? I think that's how it usually works, and would also remove the need to revert and un-revert the bold styling, which you have here with css and strong tags.

gsemet:
Look really nice :) how does it behave if sat, you have 100 entry? Does a scroll bar can appear ?

kubespawner/spawner.py
minrk: Something like:

```html
<div class='col-md-11'>
<label for="profile-item-{loop.index0}">{{profile.display_name}}</label>
<p>description...
```

jupyterhub_config.py
gsemet: I agree I am not really fan of exposing kubespawner in jupyterhub config, but that was the easiest and safest way of doing so. Or we would have had to expose all settings kubeswpaner propose, and having to maintain it manually at each evolution.
But since each entry is a dict, we can add a easier key that would only gives a subset of the settings, the most common (image, cpu/memory?)
But at the end, we can tell it is an advanced settings and read carefully the documentation.

The thing is, exposing kubesettings in jupyterhub conf means each api break (on kubespawner), for instance on major change, will impact this setting. That fine for me, but it is a dependency to take into account.

minrk: This is how Spawners are intended to be configured, so it's the right way to do it, and the design goal of the configuration system used by Jupyter and JupyterHub. All settings tagged with `config=True` are automatically exposed in jupyterhub_config.py, without taking any additional steps.

Note that this file only serves as *documentation* of configurable options, and has no influence on actual behavior, which is why I suggested moving it to examples.

gsemet: oh, maybe we want to restrict `kubespawner_override` with the option that has a `config=True` ? If the current behavior is ok for you, I'm fine too !

kubespawner/spawner.py
yuvipanda: The label is parent of everything since I want users to be able to click anywhere on it and select the radio button. The only way I've found to do this without even more JS is to put them all in the label.

gsemet:
Let's merge and cover all corner case later :)

kubespawner/spawner.py
Carreau: target input inside `label` is valid according to MDN. There is just a warning that having links _within_ the label can be an accessibility concern as it will make it hard to activate the form element.

Seeing that profile descriptions can heave links concerns me, isn't jinja supposed to auto escape  values unless marked `|safe` ?

minrk: Good point. I don't think that's jinja's default behavior, so we should probably opt-in to autoescape=True here to be safe.

ablekh:
Is this PR (perhaps, with some others) essentially a replacement for WrapSpawner / ProfileSpawner?

gsemet:
Yes, i just updates our env to use the yuvipanda enhancement, it’s great and definitively need a doc in z2jh when available !

PR #161, opened 2018-04-18
Add common labels for Pods & PVCs

### LINKED ISSUE #159: Labeling - Comply with Helms best practices
## Background
While working on [zero-to-jupyterhub-k8s#625](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/625) I've attempted to comply with [Helm's best labeling practices](https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md) and added the labels `app`, `release`, `chart` and `heritage` systematically (as well as `component` which is consider optional).

It would be great if we can make Kubespawner follow these as well even for dynamically created objects for various reasons.
1. Better safe then sorry - deviating from the common labeling practices might break someones assumption.
2. It would be required in order to allow multiple chart releases within the same namespace.
3. It would allow label selector templates that I've defined in `_helpers.tpl` work without exception for the pods spawned by Kubespawner.

## How...
We could utilize environment variables in order to communicate to kubespawner what values `app`, `release` and `chart` should have. We could also put them in a configmap available to the hub container running the kubespawner.

```yaml
# Current labeling by Kubespawner
labels:
  app=jupyterhub
  component=singleuser-server
  heritage=jupyterhub

# Suggested labeling by Kubespawner (dynamic values for app, release and chart)
labels:
  app=jupyterhub
  chart=jupyterhub-v0.7-dev
  component=singleuser-server
  heritage=jupyterhub
  release=prod
```

## Questions...
- [x] The kubespawner is making pods, but also PVC. What about those labels?
Answer: We should label the PVCs but ignore the PVs and StorageClass objects which we don't create ourselves with kubespawner.

consideRatio:
## Summary
- Changes are fully backwards compatible.
- The kubespawner is hardcoded with the label selector `component=singleuser-server` and still does not support multiple deployments in the same namespace.
- Spawned Pods and PVCs get additional `common_labels`, this allows z2jh to set the `chart` and `release` labels on spawned Pods and PVCs.
- PR utilized by [jupyterhub/zero-to-jupyterhub-k8s#625](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/625)

## Background
In [jupyterhub/zero-to-jupyterhub-k8s#625](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/625), I wanted to make all objects created within the z2jh helm chart follow [Helms Best Practices for labels](https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md). This PR helps ensure that and thereby fixes #159.

## Details
These are the labels of relevance in this PR. Note that the `PodReflector` is hardcoded to find only the pods with a `singleuser-server` label.

Object|Label|Traitlet|Default value
-|-|-|-
PVC / Pod|`component`|hardcoded|`'singleuser-storage'` / `'singleuser-server'`
PVC / Pod|`app`|`common_labels`|`'jupyterhub'`
PVC / Pod|`heritage`|`common_labels`|`'jupyterhub'`

## History
I thought of using environment variables to configure certain labels. But configuring specific variables would couple kubespawner unnecessarily (to z2jh and Helm for example).

## Implementation notes
- I made sure `common_labels` couldn't be overridden by extra labels provided to the kubespawner

## PR TODO
- [x] Configure using traitlets
- [x] Exclude all helm chart presumptions
- [x] Test this in conjunction with [jupyterhub/zero-to-jupyterhub-k8s#625](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/625)
- [x] Consider if singleuser-server's need a restart while introducing this code or changes to `common_labels` and how to avoid it. _Update_: I decided to keep a hardcoded pod-selection based on the namespace, `heritage=jupyter,component=singleuser-server` for now.
- [x] Figure out all interactions KubeSpawner might have with pods after spawn. _Update_: `start`, `poll` and `stop`, using the `PodReflector` to find the pods.
- [x] Test this better together with [jupyterhub/zero-to-jupyterhub-k8s#625](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/625). 

#### Excluded from this PR
- Figure out how / when / why KubeIngressProxy is used
- Figure out all interactions KubeIngressProxy might have with pods after spawn
- Consider sharing `common_labels` trait with KubeIngressProxy

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -933,9 +933,9 @@ class KubeSpawner(Spawner):
         labels.update(extra_labels)
         labels.update({
             'app': os.getenv('LABEL_APP', 'jupyterhub'),
-            'release': os.getenv('LABEL_RELEASE', 'unknown')
-            'chart': os.getenv('LABEL_CHART', 'unknown')
-            'heritage': os.getenv('LABEL_HERITAGE', 'jupyterhub')
+            'release': os.getenv('LABEL_RELEASE', 'unknown'),
+            'chart': os.getenv('LABEL_CHART', 'unknown'),
+            'heritage': os.getenv('LABEL_HERITAGE', 'jupyterhub'),
         })
         return labels
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -32,7 +32,6 @@ from kubespawner.reflector import NamespacedResourceReflector
 
 class PodReflector(NamespacedResourceReflector):
     kind = 'pods'
-
     labels = {
         'heritage': 'jupyterhub',
         'component': 'singleuser-server',
@@ -930,22 +929,22 @@ class KubeSpawner(Spawner):
     def _build_common_labels(self, extra_labels):
         # Default set of labels, picked up from
         # https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md
-        labels = {
-            'heritage': 'jupyterhub',
-            'app': 'jupyterhub',
-        }
-
+        labels = {}
         labels.update(extra_labels)
+        labels.update({
+            'app': os.getenv('LABEL_APP', 'jupyterhub'),
+            'release': os.getenv('LABEL_RELEASE', 'unknown')
+            'chart': os.getenv('LABEL_CHART', 'unknown')
+            'heritage': os.getenv('LABEL_HERITAGE', 'jupyterhub')
+        })
         return labels
 
     def _build_pod_labels(self, extra_labels):
-        labels = {
+        labels = self._build_common_labels(extra_labels)
+        labels.update({
             'component': 'singleuser-server'
-        }
-        labels.update(extra_labels)
-        # Make sure pod_reflector.labels in final label list
-        labels.update(self.pod_reflector.labels)
-        return self._build_common_labels(labels)
+        })
+        return labels
 
     def _build_common_annotations(self, extra_annotations):
         # Annotations don't need to be escaped
@@ -1023,6 +1022,9 @@ class KubeSpawner(Spawner):
         Make a pvc manifest that will spawn current user's pvc.
         """
         labels = self._build_common_labels(self._expand_all(self.user_storage_extra_labels))
+        labels.update({
+            'component': 'singleuser-storage'
+        })
 
         annotations = self._build_common_annotations({})
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -32,8 +32,11 @@ from kubespawner.reflector import NamespacedResourceReflector
 
 class PodReflector(NamespacedResourceReflector):
     kind = 'pods'
+    # FUTURE: These labels are the selection labels for the PodReflector. We
+    # might want to support multiple deployments in the same namespace, so we
+    # would need to select based on additional labels such as `app` and
+    # `release`.
     labels = {
-        'heritage': 'jupyterhub',
         'component': 'singleuser-server',
     }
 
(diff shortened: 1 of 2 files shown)


kubespawner/spawner.py
```diff
@@ -33,11 +33,6 @@
 class PodReflector(NamespacedResourceReflector):
     kind = 'pods'
 
-    labels = {
-        'heritage': 'jupyterhub',
-        'component': 'singleuser-server',
-    }
-
```
consideRatio: Heritage were already defined in `_build_common_labels` and `component` was added specifically anyhow in `get_pvc_manifest` and `_build_pod_labels`.

consideRatio: Oh... But perhaps they were utilized by something outside the kubespawner project that I didn't know of? I did not find an internal reference to them though.

kubespawner/spawner.py
```diff
@@ -915,22 +910,22 @@ def _expand_all(self, src):
...
-        }
-
+        labels = {}
         labels.update(extra_labels)
+        labels.update({
+            'app': os.getenv('LABEL_APP', 'jupyterhub'),
+            'release': os.getenv('LABEL_RELEASE', 'unknown')
+            'chart': os.getenv('LABEL_CHART', 'unknown')
+            'heritage': os.getenv('LABEL_HERITAGE', 'jupyterhub')
+        })
```
consideRatio: I make sure to add these after the extra labels are added.

kubespawner/spawner.py
```diff
@@ -915,22 +910,22 @@ def _expand_all(self, src):
...
+        labels = self._build_common_labels(extra_labels)
+        labels.update({
             'component': 'singleuser-server'
-        }
-        labels.update(extra_labels)
-        # Make sure pod_reflector.labels in final label list
-        labels.update(self.pod_reflector.labels)
-        return self._build_common_labels(labels)
+        })
+        return labels
```
consideRatio: This part contained redundant labels before, first `component` was added, then the extra labels, then `heritage` and `component` again from `PodReflector`, and finally `heritage` again and `app` from `_build_common_labels`.

kubespawner/spawner.py
```diff
@@ -1008,6 +1003,9 @@ def get_pvc_manifest(self):
         Make a pvc manifest that will spawn current user's pvc.
         """
         labels = self._build_common_labels(self._expand_all(self.user_storage_extra_labels))
+        labels.update({
+            'component': 'singleuser-storage'
+        })
```
consideRatio: The call to `_build_common_labels` here is the only other place `_build_common_labels` is used. So the PVCs will get all the recommended labels as well.

kubespawner/spawner.py
consideRatio: The `labels` field in the `PodReflector` overrides the `NamespacedResourceReflector` field and I didn't realize that fully until recently. As they fill the purpose of selecting the pods within the namespace with those labels, they must remain.

consideRatio:
@yuvipanda since your last review:

- Now using traitlets instead of environment variables!
- No longer coupling kubespawner with Helm!
- I decided not to introduce the complexity of letting the `PodReflector` select based on labels configured through a traitlet, but instead finds all pods using the hardcoded selection `component=singleuser-server`. The main drawback of this is that we still won't support multiple deployment within the same namespace as we could by selecting on `app=...` and `release=...` labels as well. The benefit of this choice though is that we are fully backwards compatible and avoid adding complex logic and procedures on changes to the label selection.

This PR is a prerequisite for [z2jh#625](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/625).

minrk:
Thanks, @consideRatio!

PR #172, opened 2018-05-13
Removed hack for security in k8s <= v1.5

consideRatio:
This PR reduces some complexity and could perhaps avoid some unexpected complication by removing code only relevant for k8s <= 1.5. It would also avoid adding unexpected volumeMounts to the generated singleuser-server pod specifications.

Pre k8s 1.6, the `automountServiceAccountToken` option was not available in pod specification, so a code hack was used to disable auto mounting of a `ServiceAccount` token. This was done since such token could be used to access the kubernetes master api-server in some way as far as I understand. I think the 'disable-k8s-api-access' hack worked by making a dummy mount to the same path either overriding the previously auto-mounted path or prohibiting a future mount on the path.

This PR simply removes that code and relies setting the k8s >= 1.6 `automountServiceAccountToken` option to false.

#### Questions
Some questions I hope to find answers to...
- __Q1__: What does the typical default `ServiceAccount` secret mounted for pods in `/var/run/secrets/kubernetes.io/serviceaccount` contain? I realize the volumeMount hack might have served a greater purpose than simply removing access to the `...../token` path depending on the content...
- __Q2__: Will `/var/run/secrets/kubernetes.io/serviceaccount/token` be accessible by the pod containers if they are `root`, or even `jovyan`? I think `root` has access at least, but I'd like to have this confirmed... we might expose sensitive information (the `imagePullSecret`) anyhow!

## Related
- [z2jh#447](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/447)

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -233,15 +233,11 @@ def make_pod(
         pod.spec.containers.extend(extra_containers)
 
     pod.spec.init_containers = init_containers
-<<<<<<< 4d0221e2e8c670cfe712551035add23012f48cce
-    pod.spec.volumes = volumes + hack_volumes
+    pod.spec.volumes = volumes
 
     if scheduler_name:
         pod.spec.scheduler_name = scheduler_name
 
-=======
-    pod.spec.volumes = volumes
->>>>>>> Removed hack for security in k8s <= v1.5
     return pod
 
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
I approve of this PR :) 

With `pod.spec.automount_service_account_token = False` we don't mount service tokens at all, accomplishing the same goals as our hack with less hackiness.

Issue #158, opened 2018-04-15
Infinite "Pending stop" pods


gsemet:
Hello. Does anyone already got a pod notebook, from time to time, that seem stuck somehow, andthe hub cannot kill/restart it? I see that from time to time, a given pod of a user seems unkillable by the hub. As an admin, I force stop it, but when the user restart it, it stays always with an error "400 pendning stop".

I need to go on the cluster and kill myself the pod. How can kubespawner handle this case?

gsemet:
Wonder if would not be fixed by #147. Need to test it again...

minrk:
Closing this as potentially fixed by #147 or #174. Feel free to reopen if not.

PR #174, opened 2018-05-17
Add timeout handling to reflector (replaces #88)

gweis:
Add watch and network timeout options to NamespacedResourceReflector.

This replaces #88.

The issue was, that due to some networking issues, the watcher (and sometimes even the urllib3 connection pool) stopped receiving data from the api server, without ever receiving a tcp disconnect. (most likely due to some tcp connection tracking timeout or something similar ?).

This patch allows setting kuberenetes api watch timeout (recommended to use anyway), and setting urllib3 request timeout. In case the kube client never receives a disconnect, then at least the urllib3 read timeout will be triggered, which will cause a watch restart.

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -6,9 +6,12 @@ import time
 import threading
 
 from traitlets.config import LoggingConfigurable
-from traitlets import Any, Dict, Unicode
+from traitlets import Any, Dict, Int, Unicode
 from kubernetes import config, watch
-from tornado.ioloop import IOLoop
+# This is kubernetes client implementation specific, but we need to know
+# whether it was a network or watch timeout.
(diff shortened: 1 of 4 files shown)


gweis:
Also I am wondering, whether the `asynchronise` methods in`KubeSpawner` and `KubeIngressProxy` should pass on some `_request_timeout` as well.

Can do that in a separate pull request if desired.

minrk:
Looks great! There are a couple of merge conflicts, but this should be good to go once resolved.

gweis:
I have rebased everything to current master (also kept the new default watch timeout of 10 seconds).
I think that timeout is a bit low, but it is configurable now.

gweis:
Re-referencing potentially related issues here as well:

These patches may address issues #85, #68 and may help with #65 and #158

PR #181, opened 2018-06-06
Support runAsGroup

dtaniwaki:
The latest version of Kubernetes client is 6.0 and it supports lots of new configurations.
https://github.com/kubernetes-client/python/blob/master/CHANGELOG.md#v600

We especially need [`runAsGroup`](https://github.com/kubernetes/kubernetes/pull/52077) because we sometimes need different GID over user's default GID which is the same as UID.

Could you consider supporting it?

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -29,6 +29,7 @@ def make_pod(
     image_pull_secret=None,
     node_selector=None,
     run_as_uid=None,
+    run_as_gid=None,
     fs_gid=None,
     supplemental_gids=None,
     run_privileged=False,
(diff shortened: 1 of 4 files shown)


kevin-bates:
@dtaniwaki - I'm adding runAsGroup functionality into [Enterprise Gateway](https://github.com/jupyter/enterprise_gateway) and find that the `RunAsGroup` feature-gate must be enabled for the `runAsGroup` entry to take affect, otherwise the effective group for a pod started with `runAsGroup: n` is 0.  However, I don't see anywhere in the jupyterhub organization that talks about enabling this feature-gate.  Is it assumed that users have enabled that feature-gate or is this documented elsewhere?  Or perhaps I'm missing something at the kubernetes level entirely.

consideRatio:
Hi Kevin!

That was good to know and I didn't know about this. It was apparently introduced as a featureGate in alpha status during k8s 1.10 and currently isn't enabled by default.

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

It could be nice to have a comment about this in the docs of kubespawner for this configuration option.

PR #196, opened 2018-06-27
avoid failure to delete pods that don’t exist

minrk:
stopping could raise 404 if it doesn’t exist (e.g. already deleted). Log this as a warning instead of raising.


minrk:
I'm not sure exactly what cases could cause this. An evicted pod *shouldn't* result in this failure, since the pod still 'exists' and delete shouldn't fail. I think this can only occur due to a bug where we are calling delete twice (or someone else called delete manually).

consideRatio:
@minrk ah thanks for reasoning about this with me! That is probably what I've experienced rather than an unexpected node shutdown.

PR #199, opened 2018-07-01
Rework deprecated config a bit

minrk:
- non-deprecated config takes precedence over deprecated config if both are present (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/745)
- warn about collisions
- handle deprecated config in the Config object before loading it,
  rather than via adding configurable traits (in order to detect the collisions)
- handle runtime use of deprecated config (e.g. in subclasses, hooks, etc.) via properties to ensure things stay in sync

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+.pytest_cache
 nosetests.xml
 coverage.xml
 *,cover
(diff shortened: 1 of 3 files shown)


consideRatio:
Thank you @minrk! I'll testrun this once in my cluster at home today - the code lgtm. I appreciate your comments it was easy for me to follow.

PR #203, opened 2018-07-04
add unescaped_username to template vars

gweis:
Don't escape dash '-' in user names. It is an allowed symbol in DNS labels.

I am relying on username to set up persistent volume claims and pre-populate the persistent volume with the users data. My usernames do contain one or more '-' characters, and the current escaping even escapes all dash character to '-2d'. Not sure what the preferred way of dealing with this would be. This patch is one option.

Unfortunately, this patch makes removes the ability to unescape user names. If unescaping is required, I would suggest to add up to two additional template variable names. One for the irreversible escaped version and maybe even one for the unescaped username. (can create a pull request for this as well)

Of course it would als be possible to 'legacy_escaped_username', but given that name, it sounds like, this template variable would go away in some time, and looking at the code it felt like it's doing the desired escaping by 'accident' :) .

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -912,7 +912,7 @@ class KubeSpawner(Spawner):
         List of profiles to offer for selection by the user.
 
         Signature is: List(Dict()), where each item is a dictionary that has two keys:
-        
+
         - 'display_name': the human readable display name (should be HTML safe)
         - 'description': Optional description of this profile displayed to the user.
         - 'kubespawner_override': a dictionary with overrides to apply to the KubeSpawner
@@ -1079,6 +1079,7 @@ class KubeSpawner(Spawner):
 
     def _expand_user_properties(self, template):
         # Make sure username and servername match the restrictions for DNS labels
+        # Note: '-' is not in safe_chars, as it is being used as escape character
         safe_chars = set(string.ascii_lowercase + string.digits)
 
         # Set servername based on whether named-server initialised
@@ -1092,6 +1093,7 @@ class KubeSpawner(Spawner):
         return template.format(
             userid=self.user.id,
             username=safe_username,
+            unescaped_username=self.user.name,
             legacy_escape_username=legacy_escaped_username,
             servername=servername
             )


minrk:
Yeah, this has been annoying, but it's hard to come up with something that's reliable.

Allowing collisions makes this unsafe because two users can end up with the same escaped username:

```python
In [11]: safe = '-' + string.ascii_letters + string.digits

In [12]: escapism.escape('a_x', safe=safe, escape_char='-', allow_collisions=True).lower()
Out[12]: 'a-5fx'

In [13]: escapism.escape('a-5fx', safe=safe, escape_char='-', allow_collisions=True).lower()
Out[13]: 'a-5fx'
```

Essentially, any user with an escaped character in their name can be spoofed by another user creating a username with the escaped form. So you can only do this if you know for sure that you cannot possibly have any users with names like `a-5fx`.

Another way to do it without losing safety is to use a different, less common escape character. However, if it has to be domain-safe, then `-` is the only escape character available, I think, unless you want to get creative and use [`q`](http://home.uchicago.edu/~jsfalk/misc/baby_names/) or something.

gweis:
Always these corner cases :)

My use case is, that I have guaranteed domain-safe names, which is required for some kubernetes object values, andI want my created objects to use these exact names.

How about adding a template field 'unescaped_username' instead?

Another (additional) option would be to allow callables as templates and maybe even in _expand_all. This could become a bit funny though. E.g. in the case of init_containers, one could supply a dictionary, where some values would be callables. If the callable is treated as separate option in _expand_all, then it would even be possible to create things like a full custom env for init containers based on the current spawner. I this is of interest to more users, I can create a separate issue to track discussion and start work on a pull request.

minrk:
> How about adding a template field 'unescaped_username' instead?

Yes, that's a good idea, and we should do that regardless. It's going to involve a lot of configuration to opt-in to using it everywhere, but it should be possible.

For the 'definitely-safe' case like yours sounds to be, we could add an option that skips escaping altogether, so that actions will fail if a name that requires escaping is required (essentially putting the same requirements on usernames that are applied to pods, etc.). This could be making `escape` an overrideable hook/method in general, and then setting this single value to a no-op for users that want to enforce safety at another level. How does that sound?

gweis:
Updated the pull request to add 'unescaped_username' template variable.

Not sure whether I follow you about making `escape` an overridable hook. If I get it right, it seems like it would be a nice to have, but also a potentially dangerous option, if you don't consider all edge cases like I did with the allow_collision thing. I have the feeling it would be a very rarely used customisation option, which would only help in very specific use cases. I have nothing against it, and it might be good to have, but I don't think I would use that feature.

(I might play around with the callable idea mentioned in the first comment. I will most likely need some spawner/jupyter env vars to init container and side car containers. This env will be different to the one passed to the main notebook container, s a simple `get_env` call wouldn't work here.)

minrk:
> Not sure whether I follow you about making escape an overridable hook. If I get it right, it seems like it would be a nice to have, but also a potentially dangerous option

Yeah, it's definitely true that it would make it *possible* to make unsafe decisions. But it's also possible to make safe choices, and I think by far the most common would be to disable escaping altogether, as I think would be true in your use case, which is perfectly safe. No escaping at all also has no possibility of collision, so if you know that your names are already going to be domain-safe, you can safely disable escaping altogether (e.g. if you already do some escaping at another level in the Authenticator or other source of usernames). The only failure mode for disabled escaping is if you were wrong that all your names are domain-safe, in which case the unsafe names would fail to launch due to invalid specs. For that reason, maybe it makes sense to make escape just a disableable option, rather than overridable to arbitrary schemes. What do you think about that as a safer alternative for your use case?

gweis:
I am happy with the `unescape_username`. I can add that as custom label or annotation wherever needed, or even override the existing templates.

Being able to disable escaping would have solved my use case as well. I can try and add that feature, but it probably will take a little bit to squeeze that work in.

PR #208, opened 2018-07-08
Don’t trust pod reflector when deciding whether to delete

minrk:
If pod reflector hasn’t noticed this pod yet, removal won’t be attempted, leading to a leak in pods.

Instead, just attempt deletion and handle the 404 if it’s really not there.

This exacerbates a failure mode where the pod_reflector starts missing events because pods that start but aren't noticed will be believed to be not running and won't be deleted.

This optimization was introduced in #196, but isn't worth the cost.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1472,9 +1472,6 @@ class KubeSpawner(Spawner):
                 self.event_reflector.stop()
             self.event_reflector = None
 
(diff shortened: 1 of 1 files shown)


choldgraf:
@minrk we're running into a "culling runaway" failure mode again, which I *think* this would fix, yes? Should we just merge this and deploy?

PR #216, opened 2018-07-18
restart the pod reflector when it didn’t receive highly probable messages

### LINKED ISSUE #209: pod-reflector can stop working and doesn't recover
We recently had an outage on mybinder.org, *likely* due to the pod reflector stopping receiving events. We should check for bugs in this logic, and perhaps add some more proactive "restart if suspicious" logic to ensure the pod reflector isn't allowed to be stuck in this state.

minrk:
When a pod doesn’t get deleted or (especially) doesn’t get created, this can mean that the pod reflector has started missing events. Restart the reflector when this happens.

includes #215

closes #209

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -172,9 +172,22 @@ class NamespacedResourceReflector(LoggingConfigurable):
         and as long as we don't try to mutate them (do a 'fetch / modify /
         update' cycle on them), we should be ok!
         """
+        selectors = []
+        log_name = ""
+        if self.label_selector:
+            selectors.append("label selector=%r" % self.label_selector)
+        if self.field_selector:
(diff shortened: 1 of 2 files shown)


elgalu:
We are currently facing a stuck user state when for example the singleuser pod is stuck on `ErrImagePull`.

The only we have now is restarting the Hub. I will install this to see if that solves the issue by at least avoiding the admin having to restart the Hub.

Will use this in conjunction with https://github.com/jupyterhub/jupyterhub/pull/2042

thanks!!!

minrk:
It is unlikely that this will fix an issue where a pod is *actually* in an error state. This should only fix the case where pods are starting/stopping but the PodReflector has stopped receiving events and hasn't restarted itself.

Issue #79, opened 2017-08-29
Jupyterhub: supporting multiple service accounts & namespaces with a single server instance


foxish:
The readme says:

> Spawn multiple hubs in the same kubernetes cluster, with support for namespaces. You can limit the amount of resources each namespace can use, effectively limiting the amount of resources a single JupyterHub (and its users) can use. This allows organizations to easily maintain multiple JupyterHubs with just one kubernetes cluster, allowing for easy maintenance & high resource utilization.

In another model, where there is a single hub, we may have individual user pods (notebook instances) are created in different namespaces using different service accounts (on behalf of a user), where the association between JupyterHub identity and pod service-account is maintained separately in a configmap.
Has something like this been explored before? Are you aware of limitations if one were to try and do this?

This would be related to https://github.com/jupyterhub/kubespawner/issues/76 and https://github.com/jupyterhub/kubespawner/issues/75

cc @yuvipanda

foxish:
The broad goal is to have: a multi-tenant JupyterHub capable of launching pods on behalf of a user, with a particular service account into **any** existing namespace.  
Based on some of the conversation we had on gitter and adding some more thoughts.

More concretely, we could achieve the above with the following (hypothetically).
1. Kubespawner adds a mechanism to read from a ConfigMap, and applies a service account identity to a pod, based on the OAuth identity. This ConfigMap would contain associations between JupyterHub user/group and Service Accounts.
2. UI provides a list of namespaces to choose from, into which a user can launch their notebook-pod (not sure how this can be plumbed through). 
3. If user X cannot launch their pod in namespace Y that they chose, it would surface that error back up to the UI in some way.

erolosty:
What would work best for me (and I suspect many others) would be if the function that defines namepspace resuses the templating supported for pod_name and pvc eg:

c.KubeSpawner.pod_name_template = '{username}-jupyter'   #already supported
c.KubeSpawner.namespace = '{username}-spark-cluster'       #would be great

yuvipanda:
@foxish @erolosty with https://github.com/jupyterhub/kubespawner/pull/84 we can now launch singleuser pods with access to the namespace' default service account.

yuvipanda:
@erolosty with namespace_template, would we also be creating the namespace if it doesn't exist? If we create them, do we also clean up?

foxish:
@yuvipanda, thank you! I think that's a good step ahead.

erolosty:
[...]

Issue #76, opened 2017-08-29
Allow spawning a namespace per user


yuvipanda:
Right now we use one namespace for all user pods. It'll be great to allow spawning one namespace per user, so users can do additional things in their namespaces (like spawning other clusters)

yuvipanda:
The biggest problem here is that our reflector only watches one namespace. Should we have multiple watchers? Multiple kinds of watchers? 

Perhaps we need a kind of SharedInformer design that lets us hold multiple watchers? Don't want to have one watcher per pod.

zflamig:
Is anyone working on this currently? It would be a great feature to have...

viroos:
I will be also interested in having this feature.

agt-ucsd:
I'm looking to integrate JupyterHub into an existing k8s cluster which utilizes a large number (>1000) of per-user namespaces.  

As discussed in #79 , templating Namespace seems like a reasonable approach.  Our namespaces are provisioned externally; would anyone object to deferring ns creation/cleanup to a future enhancement?

zflamig:
Based on @yuvipanda's comment, I think those would actually be separate enhancements anyways @agt-ucsd? I think this PR handles the creation/cleanup and then we need to develop a separate one which handles the watching multiple namespaces which is the thing you want?

agt-ucsd:
@zflamig  thanks, it sounds like you're in agreement that creation/cleanup of per-user namespaces is separable from generic support for multiple user namespaces (which besides converting namespace to a template/callable, would also entail changes to watching).    I'll open a separate issue for the latter.

SolarisYan:
@yuvipanda 
Hi all, How is this feature going?

lresende:
I believe this has been fixed by #458

PR #218, opened 2018-07-20
Support multiple/per-user namespaces within a single Hub instance

agt:
This PR implements placement of users' notebook containers into multiple or per-user namespaces, via a templated/callable `KubeSpawner.namespace_template` attribute.  PodReflector and related code is modified to support watching multiple namespaces.    Default behavior continues to spawn within a single namespace.  

These changes address a specific local need at our site[1], but should also facilitate future work to implement #76 and #79.  *This PR does not address provisioning or teardown of namespaces or service accounts, though as discussed in #76, these operations could parallel what's done for PVCs.*

I welcome any/all feedback on this proposal - thanks!

One use case would be to direct every user's pod into a private namespace: 

```
    c.KubeSpawner.namespace_template = 'ns-{username}'
```

Or, to direct specific named users to alternate namespaces, while maintaining all others in the default:  *(The dict below could of course be a ConfigMap passed into the Hub)*

```
    # Override namespace for certain users only
    c.KubeSpawner.namespace_template = lambda spawner: { "user_a": "group1",
                                                         "user_b": "group2"}.get(spawner.user.name, spawner.namespace)
```

Notes/comments & areas possibly needing discussion:

* In order to retain a single, shared PodReflector across all Spawner instances, while monitoring multiple namespaces, the PodReflector will invoke `list_pods_for_all_namespaces` rather than `list_namespaced_pod`. (This is roughly equivalent to running `kubectl get pods --all-namespaces --watch` in a shell.)  Pod labels will confine monitoring to Hub-related pods.    

*  `NamespacedResourceReflector` has been modified to handle non-namespaced (global) operation.  Should it be renamed to simply `ResourceReflector`?  Or split into multiple classes?

*  I introduced a separate `namespace_template` attribute, rather than converting the existing `namespace` into a template/callable, as `namespace` is referenced ubiquitously in the code and I wanted to avoid having to resolve the template/callable multiple times.  (As an alternative, I suppose we could resolve once within the constructor, then immediately swap our instance's Traitlet attribute with the resulting static string; I'm new to Python and Jupyter and don't know if such gyrations pass muster.)

*  Resolving the callable within the constructor requires a direct functional call, not a yield/Future, so the function can't block.   Perhaps resolution could be moved to `KubeSpawner.start()` to permit longer-running code (e.g. database/directory lookups)?

[1] we currently offer our students personal or team-specific k8s namespaces with variable CPU/GPU/memory quotas, preconfigured persistent storage, etc.  We're now working to shift a separate JupyterHub deployment into this k8s model, so that quotas, datasets, etc. can be freely shared between each user's Jupyter and non-Jupyter environments.  *(External provisioning processes maintain these namespaces, quotas, etc.)*

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -6,7 +6,7 @@ import time
 import threading
 
 from traitlets.config import LoggingConfigurable
-from traitlets import Any, Dict, Int, Unicode
+from traitlets import Any, Dict, Int, Unicode, Bool
 from kubernetes import config, watch
 # This is kubernetes client implementation specific, but we need to know
 # whether it was a network or watch timeout.
@@ -75,6 +75,14 @@ class NamespacedResourceReflector(LoggingConfigurable):
         """
     )
 
+    list_method_omit_namespace = Bool(
+        False,
+        help="""
+        If True, our calls to API `list_method_name` will omit a `namespace` argument.
+        Necessary for non-namespaced methods such as `list_pod_for_all_namespaces`
+        """
+    )
+
     api_group_name = Unicode(
         'CoreV1Api',
         help="""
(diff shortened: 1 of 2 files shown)


bskaggs:
Thanks for contributing this!  I agree, this approach looks extendable for things like per-user service accounts (#79), and per-user extra_annotations.

consideRatio:
@agt sorry for not picking up the ball on this contribution of yours! @lresende I'd love to hear your thoughts about this PR. I'd like for us to find a way forward with this and/or #304.

My key point of concern regarding using multiple namespaces is the permission one, that escaling the requested required permissions to the need ClusterRole's setup by default is a big deal, so making it something we opt into is a big win.

agt:
> My key point of concern regarding using multiple namespaces is the permission one, that 
> escaling the requested required permissions to the need ClusterRole's setup by default is a 
> big deal, so making it something we opt into is a big win.

I agree, smaller installations would likely view a ClusterRole requirement as a breaking change.  
I'll respond to @lresende in #304 to discuss merging our two efforts - my site derives quite a bit of benefit from this functionality and I'd love to see it pulled into the mainline code.

soumentrivedi:
do we have any update on this? Is anyone fixing the conflict issues?

agt:
@soumentrivedi Discussion has moved to #304 - I'm happy to work with you and @lresende to get that merged .  We've used this PR (#218) at my site for the past academic year, 1000's of users, and have been happy with the results.

brunocous:
How can I help with this? I would also like to see this feature in master for our use case.

manics:
I think https://github.com/jupyterhub/kubespawner/pull/387 is the most up-to-date version

minrk:
closed by #458

PR #222, opened 2018-08-01
scheduler_name - Complementary pr to #190

consideRatio:
In #190 I only added support in the make_pod function to create pods with the `spec.schedulerName` field set, but this is pointless without having a traitlet to configure what it should be in the spawner. This is added in this PR.

I also added a note on what I often need to do in order to get my tests to run, but I don't get whats going on really, but I figure it is something that is cached and needs to be rebuilt within `.eggs`.


consideRatio:
@minrk could I get a quick look from you on this?

It would be great to get this merged for my ability to test and develop #205, which is required by https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/758

consideRatio:
Closed in favor for #205 where the commits now resides.

PR #229, opened 2018-08-06
Add kubernetes service creation/destruction along with the pod

cybertony:
When you create a Jupyter pod, if you want to connect to a Spark cluster (e.g. standalone mode) in your notebook, you need to set up a k8s service for allow Spark worker to respond to your pod.

This PR code create/destruct a k8s service along to the pod with the same pod name. The service is only accessible inside the cluster (for Spark workers).

It's basically the Kubernetes solution equivalent to Docker --net=host describe below:
http://jupyter-docker-stacks.readthedocs.io/en/latest/using/specifics.html#connecting-to-a-spark-cluster-in-standalone-mode

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -150,6 +150,9 @@ def make_pod(
     pod.kind = "Pod"
     pod.api_version = "v1"
 
+    # Add pod name in meta labels
+    labels.update({"name": name})
+
     pod.metadata = V1ObjectMeta(
         name=name,
         labels=labels.copy(),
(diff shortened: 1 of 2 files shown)


minrk:
@yuvipanda can you look at this one? I particular: should this service be created by default, or be opt-in?

yuvipanda:
Can you also tell me a little more about which behavior of --net=host you are trying to reproduce here? --net=host works well when you have only one jupyter notebook server on a host, but starts breaking down when there's more than a few. It also makes more sense in a pure docker context, since there's not an easy way for workers in other nodes to talk back to the notebook server. However, everywhere a service IP (with clusterIP type) is routable, pod IP is also routable, so am not sure what adding the service object does.

PR #230, opened 2018-08-09
[Part 1 of #205] Refreshed development instructions

consideRatio:
# About
Mostly updated the instructions on how to setup a development environment and made sure the docs could build.

PS: This comes from breaking apart #205 into smaller pieces, this is the first piece.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ if v[:2] < (3, 5):
 
 setup(
     name='jupyterhub-kubespawner',
-    version='0.8.1',
+    version='0.9.0-dev',
     install_requires=[
         'jupyterhub>=0.8',
         'pyYAML',
@@ -19,7 +19,7 @@ setup(
         'jinja2',
         'async_generator>=1.8',
     ],
-    python_requires     = ">=3.5",
+    python_requires='>=3.5',
     setup_requires=['pytest-runner'],
     tests_require=['pytest'],
     description='JupyterHub Spawner targeting Kubernetes',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -10,16 +10,16 @@
 
 setup(
     name='jupyterhub-kubespawner',
-    version='0.8.1',
+    version='0.9.0',
     install_requires=[
-        'jupyterhub>=0.8',
+        'jupyterhub>=0.9',
```
betatim: I think we should be conservative here. If this version of kubespawner still works with 0.8 we should not force an upgrade.

consideRatio: Excellent question!

So hmmm, why do we have a `install_requires` jupyterhub at all for kubespawner? Why do we have `jupyterhub_config.py` in our directory?

Hmmm...

Need to think about how `profile_list` is meant to work and be used as well in order to grasp if that is a field of the kubespawner that would work with all JupyterHubs as well.

consideRatio: JupyterHub with the jupyterhub_config and [these instructions](https://github.com/jupyterhub/kubespawner/blob/master/SETUP.md#setting-up-jupyterhub-for-development) are provided as a way to setup a development environment locally.

I have not tested that setup myself, but no matter what, I think jupyterhub should not be listed as a install_requires but be listed in dev-requirements.txt or similar instead.

What do you think @betatim ?

consideRatio:
Will rebase and add build-doc instructions before mergin if approved, gtg for now!

setup.py
consideRatio: jupyterhub as a package was utilized within kubespawner, let us leave it at >=0.8. I figure it will cause more trouble if we bump it than if we don't.

minrk:
Nice!

PR #232, opened 2018-08-09
[Part 3 of #205] Added utils for k8s model verification

consideRatio:
# About
Adding two utility functions that **is not used yet**, they are simply made available and tested in some unit tests. Perhaps this shouldn't been extracted as a separate PR as it is pointless if it is not used... Hmm...
- `get_k8s_model`
- `update_k8s_model`

### About `get_k8s_model(model_type, model_dict)`
In KubeSpawner we handle dictionaries representing segments of resources strictly defined by the Kubernetes API. The user of KubeSpawner most often gives us dictionaries, but internally we could take these and create objects defined in the python kubernetes package such as `V1Container` for example, which can be importat with `from kubernetes.client.models import V1Container`.

`get_k8s_model` will take either a dict or kubernetes model object, but always returns a model object or fail attempting to create one. In this way, we will verify and fail fast if we provide faulty fields to kubespawner.

### `update_k8s_model(target, changes, logger=None, target_name=None, changes_name=None)`
This function updates a model (`target`) in place with some fields that can be set on such model (`source`). It will also warn if the `source`, which is a model object of the same kind as target or a dict, is overriding a field in the target. When it performs the optional logging, it will use the `target_name` and `changes_name` parameters for some helpful tracing.

### DIFF
--- a/kubespawner/utils.py
+++ b/kubespawner/utils.py
@@ -29,13 +29,15 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
     ).lower()
 
 
-def update_k8s_model(target, changes, logger=None, target_name=None):
+def update_k8s_model(target, changes, logger=None, target_name=None, changes_name=None):
     """
     Takes a model instance such as V1PodSpec() and updates it with another
     model, which is allowed to be a dict or another model instance of the same
     type. The logger is used to warn if any truthy value in the target is is
-    overridden. The target_name parameter can for example be "extra_pod_config",
-    allowing the logger to write out something more meaningful to the user.
+    overridden. The target_name parameter can for example be "pod.spec", and
+    changes_name parameter could be "extra_pod_config". These parameters allows
+    the logger to write out something more meaningful to the user whenever
+    something is about to become overridden.
     """
     model_type = type(target)
     if not hasattr(target, 'attribute_map'):
@@ -56,8 +58,16 @@ def update_k8s_model(target, changes, logger=None, target_name=None):
         # falsy, should not use to override the target's values.
         if isinstance(changes, dict) or value:
             if getattr(target, key):
-                if logger and target_name:
-                    logger.warning("Overriding KubeSpawner.{}'s value '{}' with '{}'.".format(target_name, getattr(target, key), value))
+                if logger and changes_name:
+                    warning = "'{}.{}' current value: '{}' is overridden with '{}', which is the value of '{}.{}'.".format(
+                        target_name,
+                        key,
+                        getattr(target, key),
+                        value,
+                        changes_name,
+                        key
+                    )
+                    logger.warning(warning)
             setattr(target, key, value)
 
     return target
--- a/kubespawner/utils.py
+++ b/kubespawner/utils.py
@@ -2,7 +2,7 @@
 Misc. general utility functions, not tied to Kubespawner directly
 """
 import hashlib
-
+import copy
 
 def generate_hashed_slug(slug, limit=63, hash_length=6):
     """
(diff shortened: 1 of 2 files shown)


kubespawner/utils.py
```diff
@@ -27,3 +27,112 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
         prefix=slug[:limit - hash_length - 1],
         hash=slug_hash[:hash_length],
     ).lower()
+
+
+def update_k8s_model(target, source, logger=None, origin=None):
```
consideRatio: rename `origin` to `target_name`, it would make it easier to understand the parameters function

kubespawner/utils.py
```diff
@@ -27,3 +27,112 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
...
+        if key not in target.attribute_map:
+            raise ValueError("The attribute 'source' ({}) contained '{}' not modeled by '{}'.".format(type(source).__name__, key, model.__name__))
+        if getattr(target, key):
+            if logger and origin:
+                logger.warning("Overriding KubeSpawner.{}'s value '{}' with '{}'.".format(origin, getattr(target, key), value))
+        setattr(target, key, value)
+
+    return target
+
+def get_k8s_model(model, model_dict):
```
consideRatio: rename `model` to `model_type` for clarity that `model` isn't an instance of the some `kubernetes.client.model.<modelname>` but actually the type of such model.

kubespawner/utils.py
```diff
@@ -27,3 +27,112 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
...
+
+    if isinstance(model_dict, model):
+        return model_dict
+    elif isinstance(model_dict, dict):
+        _map_dict_keys_to_model_attributes(model, model_dict)
+        return model(**model_dict)
+    else:
+        raise AttributeError("Expected object of type 'dict' (or '{}') but got '{}'.".format(model.__type__.__name__, model_dict.__type__.__name__))
+
+def _get_k8s_model_dict(model, obj):
```
consideRatio: rename `model` to `model_type` for clarity that `model` isn't an instance of the some `kubernetes.client.model.<modelname>` but actually the type of such model.

kubespawner/utils.py
```diff
@@ -27,3 +27,112 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
...
+        raise AttributeError("Expected object of type '{}' (or 'dict') but got '{}'.".format(model.__type__.__name__, obj.__type__.__name__))
+
+def _map_dict_keys_to_model_attributes(model, model_dict):
+    """
+    Maps a dict's keys to the provided models attributes using its attribute_map
+    attribute. This is (always?) the same as converting camelCase to snake_case.
+    Note that the function will not influence nested object's keys.
+    """
+
+    for key in list(model_dict.keys()):
```
consideRatio: Since i modify the model_dict while looping over its keys, I better loop over a copy of the keys rather than evaluating then while modifying them. So, I'm creating a new list from model_dict.keys() by passing it to list(), perhaps i should use copy() to make it more obvious?

kubespawner/utils.py
```diff
@@ -27,3 +27,112 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
...
+        model_dict[_get_k8s_model_attribute(model, key)] = model_dict.pop(key)
+
+    return model_dict
+
+def _get_k8s_model_attribute(model, field_name):
```
consideRatio: make it clear that this parameter actually is a model instance.

kubespawner/utils.py
```diff
@@ -27,3 +27,130 @@ def generate_hashed_slug(slug, limit=63, hash_length=6):
...
+    """
+    Maps a dict's keys to the provided models attributes using its attribute_map
+    attribute. This is (always?) the same as converting camelCase to snake_case.
+    Note that the function will not influence nested object's keys.
+    """
+
+    # it is important to iterate over a copy of the dictionaries keys, as they
+    # will be updated in the loop
+    for key in list(model_dict.keys()):
+        model_dict[_get_k8s_model_attribute(model_type, key)] = model_dict.pop(key)
```
minrk: I think this should create a new dictionary instead of mutating one in-place

consideRatio: Thanks for considering this @minrk.

I'm intending to use this helper twice as written out below, this code is from PR Part 4.

```python
    # the assignments to notebook_container and pod.spec is pointless atm, but won't be if I update the utility function update_k8s_model to be get_updated_copy_of_k8s_model or similar.
    if extra_container_config:
        notebook_container = update_k8s_model(
            target=notebook_container,
            changes=extra_container_config,
            logger=logger,
            target_name="extra_container_config",
        )

    # ...

    if extra_pod_config:
        pod.spec = update_k8s_model(
            target=pod.spec,
            changes=extra_pod_config,
            logger=logger,
            target_name="extra_pod_config",
        )
```

I reasoned that we are already making a lot of changes to `notebook_container` and `pod.spec`, and pod.spec will be a huge object relatively so I was thinking I wasn't sure that I could do a deepcopy and retain good enough performance. Also note that anything else than a deepcopy of objects like pod.spec containing a lot of nested reference objects would be misleading I think.

consideRatio: I trust your experience on this, so I'll go with whatever you think is best and be right on it, the change won't be causing much work at all if we do it!

consideRatio: Hmmm... `target_name` isn't right, it's really `changes_name`

consideRatio:
@minrk I implemented the suggested changes, and made some improvements to docstrings etc.

I will now merge this unused code so I can rebase my next pr on top of this. Thanks for the review @minrk :heart!

PR #236, opened 2018-08-10
[Part 4 of #205] Bugfixes, input verification and testing - Review gamified!

consideRatio:
## About
- Various bug fixes
- Verification of input using the `kubernetes` python package
- Additional tests

## Commit checklist review
Review one commit at the time and check the *checkboxes of approval* or give *helpful feedback*!

All the points in the commit history have passed the the tests!

### Suggested feelings
During a typical review  | During a *checkboxes of approval* review
-|-
image|image

## Commit Checkboxes of Approval
A checked checkbox means the commit by itself has been approved by a JupyterHub team member.

- [x] **Selfchecked after additional + additional self-review** https://github.com/jupyterhub/kubespawner/commit/f630ca646310cd60e5ca29538e2156fc1df1773d - I'm starting to use the utils introduced in #232. The utils and I will attempt to use dictionaries in order to create objects from models defined in the `kubernetes.client.models` package. If it is possible it will mean they had valid contents, as the constructors of the models only has certain parameters. - @consideRatio wrote the following summary:
  1. **SUMMARY:** Given `update_k8s_model` and `get_k8s_model` does what it should, which is verified by tests in #232 and the pre-existing tests, this commit simply adds logging whenever `extra_pod_config` or `extra_container_config` overrides something as well as verifies that all extra_containers dictionaries can represent valid V1Containers.
  2. New utils are imported
  3. I add the optional param `logger=None` to the `make_pod` function 
  4. I move the `pod.spec.containers.append(notebook_container)` statement down in the code until notebook_container has been configured fully.
  5. The `if extra_containers` statement is also moved downwards
  6. I remove the function called `_map_attribute` previously used to to update `extra_pod_config` and `extra_container_config` and use `update_k8s_model` instead - it does the same thing but will log overrides. So this is in reality a refactoring of the `_map_attribute` function to `utils.py`'s [`_map_dict_keys_to_model_attributes`](https://github.com/jupyterhub/kubespawner/pull/232/files#diff-782542afc84f3ec2591f7d7d8facf61bR105 )
 which relies on [`_get_k8s_model_attribute`](https://github.com/jupyterhub/kubespawner/pull/232/files#diff-782542afc84f3ec2591f7d7d8facf61bR118) for the main logic.
  5. `extra_containers` added are verified by letting provided dictionaries be turned into V1Container objects through `get_k8s_model`.

- [x] https://github.com/jupyterhub/kubespawner/commit/a7cee56b86f8c5f3fe956431e3a1d8ef83b5676a - The `singleuser_` deprecation test had a bug where the assertion was always true, like `"something"=="something"` is.

- [x] **Selfchecked**: https://github.com/jupyterhub/kubespawner/commit/e889b22000532f11e69f4dab6c01724b4d6a8fb3 - In #190 I added support for setting `scheduler_name` but it had forgotten to add a commit making the option broken - until now.
  1. **SUMMARY:** Optional to use, super confident that this commit won't influence anyone badly.
  2. Docs reword
  3. Traitlets added to KubeSpawner
  4. Test added to verify functionality and no other test breaks because something new shows up
  5. The traitlet, with default value of `None` is passed to `make_pod(...)` from `get_pod_manifest(...)`

- [@minrk reviewed](https://github.com/jupyterhub/kubespawner/pull/205#pullrequestreview-141453367) and helped me realize that "In Python, default keyword arguments should never be mutable. So anywhere this should be a list or dict, it should be `None`.". So in this and the following commits I make sure we don't do that any more.
  - [x] https://github.com/jupyterhub/kubespawner/commit/ee8a6d4e21f09a11076e01cf8dad3cd83e5f0b33
  - [x] https://github.com/jupyterhub/kubespawner/commit/446e0609000c1194cd708b027886158471c05b01
  - [x] https://github.com/jupyterhub/kubespawner/commit/5a05e9535b5b5bb56b4465d9da8968b74d210100
  - [x] https://github.com/jupyterhub/kubespawner/commit/51445b5141fec11909a13243516dbfb53a4251ca

- [x] https://github.com/jupyterhub/kubespawner/commit/dd7611720e7433f5b57939a6f79bf8aa87ec6932 - In this and the next commit I verify that kubespawner is provided valid input for `lifecycle_hook` and `init_containers`.

- [x] https://github.com/jupyterhub/kubespawner/commit/d6f6070291682c3ac37214995158dcf28f1090b6

- [x] https://github.com/jupyterhub/kubespawner/commit/dc997ccd301214a394686cbb7c87d7960621a5cd - Code brevity improvement - utilizing the `.update` method of dictionaries instead of a for loop.

- [x] https://github.com/jupyterhub/kubespawner/commit/c8a897bd6956ada94e54d9077c52a7cf85f5da8a - After lots and lots of rebasing and recommitting in order to make it easy to review things, I have come to love trailing commas, it makes it a lot easier with git as a lot fewer lines will be modified in general.

- [x] https://github.com/jupyterhub/kubespawner/commit/2448a11a3de7dd67e7da426e153f28a298d8cc1c - I wanted a few spaces cleaned up while making a big rebase in order to avoid being distracted.

- [x] https://github.com/jupyterhub/kubespawner/commit/16f57ba8336059360ed1c0bf86b1ca76bca52a1f - I spent hours trying to understand `traitlets`. If you look in the [documentation](https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.Dict) you will find the latest and unreleased constructor definition. But that does not match the latest released version. When I look at the documentation of the release we are using, I'm led to beleave the first argument should be a `trait_type` or `trait` as it is called in the release we are using, but that was wrong. In the source code, the arguments are swapped place in order to be backward compatible.

  With all this said, it made a great deal of sense to me to ensure we used the parameter names explicitly in order to avoid confusion and perhaps be able to upgrade to a later release without trouble. In this commit I also made sure to use the available decorators that helps anyone grasp whats going on instead of relying on magical function names having special meaning.

- [x] https://github.com/jupyterhub/kubespawner/commit/d11322d149470011aa5bad07dd9ab2e050c6b909 - Some links pointed to old K8s versions, so I updated them.

- [x] https://github.com/jupyterhub/kubespawner/commit/940f0af943a9074c76d8ecb661e5761f67143162 - In this commit I avoid starting up a big machinery in for unit tests, it allows me to make a somewhat bigger test that creates a KubeSpawner being configured with a config I can control and return a pod manifest through the `get_pod_manifest` method that in turns calls `make_pod`.

- [x] https://github.com/jupyterhub/kubespawner/commit/86967b580df210e8b6abc57473391fea4d652b69 - I read in the documentation about being able to provide both `snake_case` and `camelCase`, so I setup a test to verify this.

- [x] https://github.com/jupyterhub/kubespawner/commit/32845b5df9ed820b0fd409356b41ac50daa91d50 - Now I utilize the mocking step introduced in https://github.com/jupyterhub/kubespawner/commit/940f0af943a9074c76d8ecb661e5761f67143162 and verify output all the way from KubeSpawner creation with a certain config, through `my_spawner.get_pod_manifest(...)` which in turn utilizes the `make_pod(...)` function.

- [x] https://github.com/jupyterhub/kubespawner/commit/2de3cd0fa6d584e2bea27109e9515b98140e5011 - I made minor documentation fixes. In a future PR I'll make sure the documentation renders properly as I think the docs contain a lot of inconsistencies when it comes to formatting.


minrk:
I would like to revert most of 16f57ba8336059360ed1c0bf86b1ca76bca52a1f. Most of the changes I think are regressions, not improvements. In particular:

- adding `default_value=` instead of using the first (and only) positional arg as the default value, which is idiomatic for traitlets.
- Similarly, reverting the `.tag(config=True)` changes. Originally, `config=True` as a constructor argument was going to be deprecated, but that decision has been reverted and I think `config=True` as a constructor argument is clearer than waiting until after the end of the traitlet definition to see if it's configurable.

I would say that most configurable traitlet definitions should look like:

```python
name = TraitType(
    default_value,  # omitted for objects with clear defaults, such as List or Dict
    config=True,
    help="""...""",
)
```

Additionally, having 'cleanup' like this in a PR with other behavioral changes I think is something we should try to avoid. Keeping PRs to one task rather than an accumulation of unrelated changes makes it easier to review things one at a time.

Traitlets changes that are good to keep:

- adding the `@default` decorators, which does clarify things, even though the old way is not deprecated

consideRatio:
@minrk I added summaries of the not yet approved commits to make them easier to grasp and consider. I'm 100% about the `scheduler_name` bugfix commit, and confident about the use of the util functions will be fine to merge.

The verification aspects of this PR btw, will make those that has provided faulty `init_containers`, `extra_containers` and `lifecycle_hooks` will fail before an API-call is made to the k8s API server, but only at the time the spawner is attempting to generate a manifest.

Perhaps we should make the spawner fail even faster by verifying the setting of traitlets in the future? That way we could fail on hub startup when config is loaded!

kubespawner/spawner.py
```diff
@@ -726,119 +747,135 @@ def _hub_connect_port_default(self):
...
+        See `the Kubernetes documentation <https://kubernetes.io/docs/concepts/workloads/pods/init-containers/>`_
+        for more info on what init containers are and why you might want to use them!
 
         To user this feature, Kubernetes version must greater than 1.6.
         """
     )
 
     extra_container_config = Dict(
-        None,
+        {},
```
minrk: The default for container (List, Dict) traits is an empty container, so it would be more idiomatic to leave these out, rather than define them as an unused empty dict.

consideRatio: Thanks for helping me grasp whats going on, I though they would have a default value of Undefined from the source code, but I realize now the logic. It makes sense to write the default value of `default_value` to be `Undefined` instead of `{}` - to avoid using a reference type value as a parameters default value just like you made me get to learn previously!

I'll exclude the positional and named way to set the default_value parameter altogether and allow the hidden-in-source initialization logic of Dict() and List() initialize them as `{}` and `[]` respectively. 

image

consideRatio: I managed to see that Dict() initialized to `{}` after closer inspection of the source, but for List(), I fail to see how it ends up with `[]`.

image

I don't see any trace of this in the documentation either:
https://traitlets.readthedocs.io/en/stable/trait_types.html#containers

---

@minrk After inspecting the List traitlet's source code and documentation, and still remaining confused as to how this will become an empty list I'm getting uneasy. I think we should deviate from the idiomatic way to not redundantly specify things, as it is the best way we can to better document the behavior for users and developers of KubeSpawner. Writing a comment at one place won't suffice, and writing it in all the KubeSpawner's traitlets docstrings would really clutter the documentation.

What do you think? I'll happily strip away these default_value positional arguments as long as I know you have heard this concern and still would appreciate to strip them.

kubespawner/spawner.py
```diff
@@ -439,14 +461,14 @@ def _hub_connect_port_default(self):
 
         Default is None and means it will be launched in any available Node.
 
-        For example to match the Nodes that have a label of `disktype: ssd` use::
+        For example to match the Nodes that have a label of `disktye: ssd` use::
```
willingc: typo: disktype

consideRatio: Thank you @willingc !

kubespawner/spawner.py
consideRatio: I went ahead and removed it. At least it is now systematically implicit where it previously was a mix of implicit and explicit.

consideRatio:
@minrk do you approve of the following parts? Then I'd feel very confident about a merge:

- [ ] https://github.com/jupyterhub/kubespawner/pull/236/commits/0ca40a4de02db8a83e70d2f5de19685463c1883c - allow_none on Unions
- [ ] https://github.com/jupyterhub/kubespawner/pull/236/commits/4c4d4881b38c32ecce1ed65fbde49e5247506382 - Dict() and List() no longer specify a default value

willingc:
Thanks for hanging in there through this @consideRatio :heart:

PR #243, opened 2018-08-16
install minikube on Travis

minrk:
and run a couple very basic tests that actually launch pods

this should get us setup so that we can start testing actual functionality

I think there's room to consolidate the boilerplate for tests that look like:

1. given this config
2. instantiate and launch a KubeSpawner
3. assert some condition

so we can get some nice test coverage, but this gets the ball rolling.

additionally, this enables [coverage on codecov](https://codecov.io/gh/jupyterhub/kubespawner/tree/bc1a4ac8febec1df0830999cdbde899781230c3d/kubespawner), which can point us to where we need to assemble those tests. However, since most kubespawner case coverage is really in configuration, code line coverage doesn't give us a useful view of how we are doing there.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@ def kube_ns():
     """Fixture for the kubernetes namespace"""
     return os.environ.get("KUBESPAWNER_TEST_NAMESPACE") or "kubespawner-test"
 
+
 @pytest.fixture
 def config(kube_ns):
     """Return a traitlets Config object
@@ -28,7 +29,7 @@ def config(kube_ns):
 
 
 @pytest.fixture(scope="session")
-def kube_client(kube_ns):
+def kube_client(request, kube_ns):
     """fixture for the Kubernetes client object.
 
     skips test that require kubernetes if kubernetes cannot be contacted
@@ -44,4 +45,6 @@ def kube_client(kube_ns):
         client.create_namespace(V1Namespace(metadata=dict(name=kube_ns)))
     else:
         print("Using existing namespace %s" % kube_ns)
+    # delete the test namespace when we finish
+    request.addfinalizer(lambda: client.delete_namespace(kube_ns, {}))
     return client
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,47 @@
+"""pytest fixtures for kubespawner"""
+
+import os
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,50 @@
...
+    The base configuration for testing.
+    Use when constructing Spawners for tests
+    """
+    cfg = Config()
+    cfg.KubeSpawner.namespace = kube_ns
+    return cfg
+
+
+@pytest.fixture(scope="session")
+def kube_client(request, kube_ns):
```
betatim: why the request fixture?

betatim:
LGTM.

consideRatio:
Wieeee LGTM!

consideRatio:
Awesome! I rebased and tried running tests for #236 and it worked like a charm.

---

```
pip install --pre -e ".[test]"
```

**Q1:** Should we really specify `--pre` to allow installation of pre-releases etc though?

**Q2:** What does it mean to have the ".[test]" as compared to having simply "."?

**Q3:** How to make `python setup.py test` run pytest with the --verbose option?

**Q4:** Is there something we can do about this warning that appears to me at the end of the test on my local computer?
image

betatim:
> Q1

I think it is a good idea to have `--pre` as it gives us a bit of warning when libraries we depend on break things

> Q2

It installs the "extras" from [the test section](https://github.com/jupyterhub/kubespawner/pull/243/files#diff-2eeaed663bd0d25b7e608891384b7298R24)

> Q3

Does running `pytest --verbose` not work? I almost never run `python setup.py test` and go for running `pytest` directly. Not sure what the pros and cons are exactly.

> Q4

What version of kubernetes-python do you have? Seems like something deep inside that library :-/

consideRatio:
@betatim 

> Q1

Good point I agree!

> Q2

Ah!

> Q3

It works, I had never used it before, and had always been running `python setup.py test` as I read about in some developer docs a while back i recall. Then i notices there was a difference of running that and what was run on travis.

We could update the contributor docs so it suggest we run the same command as travis and I'd be happy!

Q4:
ah good point, my cluster was k8s 1.11. I will ignore this and see if problems are resolved with using the kubernetes python client v7 (which has proper k8s 1.11 support).

PR #245, opened 2018-08-17
bump to kubernetes client v7 - k8s 1.11 support

consideRatio:
This client introduces support for new API found in k8s 1.11. It
contains a breaking change where a parameter called async to functions
within the package must be renamed to async_req. We don't use this
parameters so it should be fine.

https://github.com/kubernetes-client/python/releases/tag/v7.0.0


betatim:
We should merge this when we have time to directly deploy it on mybinder.org and then monitor it. There are a few fixes/hacks in kubespawner to work around "leaks" of various kinds in kubernetes-python.

For example https://github.com/jupyterhub/kubespawner/blob/05413482f2e29583a81b1cb6a60a25a1411ab4ba/kubespawner/clients.py#L13-L20

We should track down all of those instances, undo them if we think they are fixed now and then test with a deploy. Maybe something to do after JupyterCon?

consideRatio:
Aha! Lets do that @betatim!

setup.py
```diff
@@ -14,7 +14,7 @@
     install_requires=[
         'jupyterhub>=0.8',
         'pyYAML',
-        'kubernetes==6.*',
+        'kubernetes==7.*',
```
minrk: Let's also take this opportunity to switch this to `>=`. Pinning with `==` in library dependencies is frowned upon in Python.

minrk:
I believe there's only one such patch, and it is still needed in kubernetes 7 because [the relevant code is unchanged](https://github.com/kubernetes-client/python/blob/v7.0.0/kubernetes/client/api_client.py#L67).

clkao:
btw the `async` keyword in the previous v6 client prevented pythong 3.7 from working

consideRatio:
@clkao yes! Apparently the kubernetes package used the keyword `async` as a name for some function parameters! I did not find any occurrences of us using a function with that parameter explicitly named, so we didn't need to update for that.

consideRatio:
Ready for merge then perhaps?

PR #247, opened 2018-08-18
Update docs for 0.9b1

minrk:
After this, I think we can cut the first 0.9b1 release, and make 0.7b1 of the helm chart.


kubespawner/spawner.py
```diff
@@ -641,8 +642,8 @@ def _hub_connect_port_default(self):
            - `mountPath` The path on the container in which we want to mount the volume.
            - `name` The name of the volume we want to mount, as specified in the `volumes` config.
 
-        See `the Kubernetes documentation <https://kubernetes.io/docs/concepts/storage/volumes>`_ for more information on how
-        the `volumeMount` item works.
+        See `the Kubernetes documentation <https://kubernetes.io/docs/concepts/storage/volumes>`__
```
consideRatio: Another double underscore, I did not know about this syntax, I've seen a single one being used within the z2jh docs throughout.

docs/source/objects.rst
```diff
@@ -2,13 +2,8 @@
 Objects
 =======
 
-Module: :mod:`kubespawner.objects`
-----------------------------------
-
 .. automodule:: kubespawner.objects
 
-.. currentmodule:: kubespawner.objects
-
```
consideRatio: AHA! Have this perhaps caused documentation duplications in the past? I remember such bug.

consideRatio:
LGTM! I'd really like to see #238 and #239 in the release though, while I also think we should wait with #245 as @betatim suggest.

kubespawner/spawner.py
minrk: double-underscore means "anonymous link" whereas single underscore is "named link" which creates a named ref other links can refer to. Having multiple named refs with the text "Kubernetes documentation" was causing a collision.

docs/source/objects.rst
minrk: I think this was harmless, it's just duplicated in automodule as well.

kubespawner/spawner.py
consideRatio: Ah!

consideRatio:
Is this ready for merge?

PR #251, opened 2018-08-21
Fix pod default restart_policy and set it to OnFailure

clkao:
Follow up to #143 correcting the restart_policy key.

Change the behaviour to OnFailure as this would:
- cleanly shutdown when shutdown_no_activity_timeout is triggered.
- restart user notebook when random notebook crash happened.  In this case users would only see minor disruption (likely 500) as oppose to the notebook completely gone and have to be relaunched.

### DIFF
--- a/tests/test_objects.py
+++ b/tests/test_objects.py
@@ -36,7 +36,6 @@ def test_make_simplest_pod():
                         "name": "notebook-port",
                         "containerPort": 8888
                     }],
-                    'restartPolicy': 'OnFailure',
                     'volumeMounts': [],
                     "resources": {
                         "limits": {},
@@ -44,6 +43,7 @@ def test_make_simplest_pod():
                     }
                 }
             ],
+            'restartPolicy': 'OnFailure',
             'volumes': [],
         },
         "kind": "Pod",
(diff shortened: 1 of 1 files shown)


tests/test_objects.py
```diff
@@ -719,6 +732,7 @@ def test_make_pod_with_extra_pod_config():
                         "containerPort": 8888
                     }],
                     'volumeMounts': [],
+                    'restartPolicy': 'OnFailure',
```
consideRatio: Restart policy in extra_pod_config was set to Always, but should output OnFailure? Probably not right?

I was looking to find why you had a build failure, this isn't the main reason but it will probably need to be fixed as well.

consideRatio: Oh hmmm wait, this is the containers restartPolicy, eehmmm okay I need to read up on Pod / Container restart-policies.

consideRatio:
Sounds good thanks @clkao, lets just fix the tests!

If the pod is shutdown by the hub or "Stop my Server", the pod will exit without failure right?

kubespawner/objects.py
```diff
@@ -228,7 +228,7 @@ def make_pod(
     )
 
     pod.spec = V1PodSpec(containers=[])
-    pod.spec.restartPolicy = 'Never'
+    pod.spec.restart_policy = 'OnFailure'
```
consideRatio: I considered this change between restartPolicy and restart_policy, V1PodSpec - the python model representation, contains restart_policy and will render to restartPolicy later, so it makes sense. I think. Nice catch!

consideRatio:
Thanks for all the work you do @clkao ! :D

PR #256, opened 2018-08-23
advance progress bar with events

minrk:
with each event, step 33% closer to 90% complete.

we may not know how close we actually are, but this avoids the progress bar sitting still at 50%, instead moving as events occur.

It takes 6 events to get to 80% and 6 more until it holds steady at 89%. A typical launch seems to produce ~5-10 events.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1524,6 +1524,7 @@ class KubeSpawner(Spawner):
         pod_id = None
         first_run = True
         start_future = self._start_future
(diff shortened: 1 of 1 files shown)


clkao:
Cool! another heuristic to consider is the number of volumes, which is definitely at least one event each.

minrk:
@clkao Good idea! I think it would be cool to count volume mounts, container starts, etc. in the future to get a more realistic bar.

PR #259, opened 2018-08-30
Added support for username and userid expansion for working_dir argument

DerekHeldtWerle:
This PR aims to provide the same functionality already present in the expansion of username and userids in extra_annotations, notebook_dir, pod_name_template, and more to be extended to the working_dir argument. 

With this, in the jupyterhub_config.yml the following can be properly expanded/supported:
`c.KubeSpawner.working_dir = '/home/{username}'`


kubespawner/spawner.py
```diff
@@ -138,6 +138,7 @@ def __init__(self, *args, **kwargs):
         # runs during both test and normal execution
         self.pod_name = self._expand_user_properties(self.pod_name_template)
         self.pvc_name = self._expand_user_properties(self.pvc_name_template)
+        self.working_dir = self._expand_user_properties(self.working_dir)
```
minrk: Tests are failing. Looks like because the default working_dir is None, so need to check `if self.working_dir` here before rendering the template.

DerekHeldtWerle:
Hmm looks like my update and rebase caused some unintended behavior for this PR. Shall I close and open a new one to clean this up? @minrk

PR #260, opened 2018-09-08
Try getting image_spec from POST parameters, fallback to self.image_spec

yoavtzelnick:
Allow specifying the docker image to be spawned.
Example usecase: 
`POST /user/test/server/projectA
{
    "image_spec": "projectA:latest"
}`

`user_options` is provided by JupyterHub as seen [here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py)

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1324,6 +1324,9 @@ class KubeSpawner(Spawner):
         else:
             real_cmd = None
 
+        # JupyterHub user options via REST
+        self.image_spec = self.user_options.get("image_spec", self.image_spec)
+
         labels = self._build_pod_labels(self._expand_all(self.extra_labels))
         annotations = self._build_common_annotations(self._expand_all(self.extra_annotations))
 


yoavtzelnick:
Personally I don't like deployment changes, that's a good way to lose track of the original repo in my opinion.
Is there anywhere that you don't want maximum flexibility for your users in a Dev oriented project?

minrk:
> Personally I don't like deployment changes, that's a good way to lose track of the original repo in my opinion.

I don't think so. Specifying user_options and the options form is typically set as part of individual deployments, not in spawner defaults, because what inputs a given deployment wants to expose to a user varies widely. Some Spawners like KubeSpawner can generate a default form from a set of "profiles", but this only includes a whitelist of known-good configuration combinations, but this is standard practice for a configured JupyterHub deployment.

> Is there anywhere that you don't want maximum flexibility for your users in a Dev oriented project?

Yes, most of the time. Typical JupyterHub deployments, especially on Kubernetes, have some sort of resource restrictions, such as limiting resources or whitelisted images, etc. Allowing totally free user input (e.g. to request 100 CPUs or arbitrary images) is not a common pattern, but it is quite achievable by standard customization of `options_form` and handling of `user_options`.

Here is a`jupyterhub_config.py` that seems to achieve your goal with standard customization:

```python
from kubespawner import KubeSpawner

class MySpawner(KubeSpawner):
    def start(self):
        if 'image' in self.user_options:
            # users are allowed to pick arbitrary images
            self.image_spec = self.user_options['image']
        return super().start()

c.JupyterHub.spawner_class = MySpawner
```

minrk:
Closing as this is a configuration option, not default behavior.

PR #275, opened 2018-10-23
Fix image_pull_policy

BerserkerTroll:
`image_pull_policy` **defaults**, i.e. gets derived from `image_spec` tag, only if it is not set explicitly!
`image_pull_policy='IfNotPresent'` + `:latest` tag ≠ `image_pull_policy='Always'`. Event `:latest` image won't be pulled if it exists!
If it is desired to have `image_pull_policy` 'dependent' on the tag (as-if it were `'Always'` for `:latest`/omitted tag and `'IfNotPresent'` otherwise), it should be set to `None`, which corresponds to `imagePullPolicy` omission. https://kubernetes.io/docs/concepts/containers/images#updating-images:
> If you would like to always force a pull, you can do one of the following:
> - omit the `imagePullPolicy` and use `:latest` as the tag for the image to use.
> - omit the `imagePullPolicy` and the tag for the image to use.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -397,8 +397,9 @@ class KubeSpawner(Spawner):
     )
 
     image_pull_policy = Unicode(
-        'IfNotPresent',
+        None,
         config=True,
+        allow_none=True,
         help="""
         The image pull policy of the docker container specified in
         `image_spec`.


betatim:
Do we want to support a "not best practice" of using `latest` as a tag instead of an explicit tag? Using `:latest` in production setups makes it very hard to track which version of the image is actually running and makes it hard to reproduce bugs (even for the same user as the next time they launch their notebook server things might start or stop working because the base image changed).

---

As a point of communication style: calling this a "fix" isn't helpful as it implies that the current behaviour is somehow "broken". Which might or might not be true. So best to avoid that implying that in order to foster a good discussion.

BerserkerTroll:
> Do we want to support a "not best practice" of using latest as a tag instead of an explicit tag?

Are you serious? The goal of the project should be to correctly reflect the k8s behavior, not to force someone's view on which practice is best.

> As a point of communication style: calling this a "fix" isn't helpful as it implies that the current behaviour is somehow "broken". 

The `help` message of the traitlet says "For more information on image pull policy, refer to `the Kubernetes documentation <https://kubernetes.io/docs/concepts/containers/images/>`", which implies that its intent to reflect the behavior of the `imagePullPolicy` k8s parameter. But omitting `imagePullPolicy` in the .yaml-file and using the `:latest` tag is not the same as using the `:latest` tag and not setting the `image_pull_policy`. So `image_pull_policy` is obviously broken.

> best to avoid that implying that in order to foster a good discussion.

Did you adopt some brain-damaged CoC forbidding calling bugs "bugs"? How should I call them now? "Alternatively correct behavior"?

consideRatio:
Thank you for this PR @BerserkerTroll! I did not know about this behavior before and I always appreciate learning more about kubernetes.

**Technical summary based on my understanding:**
Kubespawner is currently always applying a `imagePullPolicy` to the generated pod spec since we have a truthy default value for it (`IfNotPreset`). To supply a default value like this to the pod spec is not a common practice within kubespawner. Typically kubespawner omits details not explicitly configured!

---

I think it makes sense to not specify it in the generated pod spec due to this alone, and especially since the omitting of the pull policy in the pod spec can enable the dynamic behavior as @BerserkerTroll described where the actual pull policy depends on the tag of the image.

BerserkerTroll:
>  To supply a default value like this to the pod spec is not a common practice within kubespawner. Typically kubespawner omits details not explicitly configured!

Yes. Even more obvious argument would be that `V1Container` constructor's `image_pull_policy` parameter has the default value `None` (https://github.com/kubernetes-client/python/blob/3fb2be14e18d84edef094bbd908b6bb3e39aafe6/kubernetes/client/models/v1_container.py#L81)
I don't think `KubeSpawner` ever had the intent to supply default config values different from k8s API default values.

choldgraf:
Hi @BerserkerTroll , thanks for your contributions and your thoughts. I'd like to quickly highlight some of the main points of the [Jupyter community Code of Conduct](https://github.com/jupyter/governance/blob/master/conduct/code_of_conduct.md).

* Be friendly and patient.
* Be welcoming.
* Be considerate. 
* Be respectful. 
* Be careful in the words that you choose.

Suggesting the package is broken or suggesting that the project's decisions are "brain damaged" is not constructive. Please keep a positive and constructive tone when you engage with this community.

BerserkerTroll:
The funny thing is that this is not the first commit/PR having the "fix" word in its title, but it looks like it is the first one (I didn't check all of them) where this word caused such a reaction. So I conclude that the reason is not the word, but me. And the "F" word in the title is just a far-fetched reason.

choldgraf:
Sorry you feel that way @BerserkerTroll - to be transparent, these are your comments that raised a flag in my mind:

> Are you serious? The goal of the project should be to correctly reflect the k8s behavior, not to force someone's view on which practice is best.

> Did you adopt some brain-damaged CoC forbidding calling bugs "bugs"?

consideRatio:
Thanks for raising this @choldgraf.

I struggled with this thread, communication in text messages is so extreamly limited and challenging.

I've learned more about myself. I mean to address something similar without fail next time and not become a passive bystander.

:heart:

willingc:
Hi folks. I've gone ahead and locked this PR so all can pause communication here for a few days. 

All, this is a gentle reminder to reflect that the written word loses tone and context. Here's a few tips that I recently shared on another project.

> Just a reminder about the difficulty of email negotiation [1]:

> Email negotiations are also fraught with misunderstanding, both because emotion and tone are difficult to convey accurately and because parties neglect to consider the other side’s perspective. Notably, email communicators 1 are largely unaware of their limitations.

> The following tips may help [2]:

> **Stamp out conflict.** When rudeness is encountered, or you are angry, don’t respond immediately and don’t respond in kind. Instead, take a short break, and then contact your counterpart by phone, or email a simple statement of concern and schedule a face-to-face meeting as soon as possible. Conflict with emotional intensity is rarely ever resolved over email.
> **Ask more questions, not less.** There is a tendency to limit your questions over email because it appears tedious. Don’t fall into this trap. To avoid lengthy and exhausting lists, start with broad questions, intersperse phone conversations to discuss the answers, and use shorter emails to group follow-up questions by topic.
> **Keep the climate positive.** Maintain a friendly tone in emails. Interpret e-mail messages with caution and sensitivity, and leave room for personality, style and cultural differences. Make sure to clarify any ambiguities right away. And remember to use generally accepted best practices in email etiquette.

[1] https://www.pon.harvard.edu/daily/conflict-resolution/email-more-cons-than-pros/
[2] https://www.watershedassociates.com/learning-center-item/negotiating-over-email.html

PR #276, opened 2018-11-02
Expand username and userid in extra_containers

LoicAG:
This will help in our use-case: a volume is mounted on the jupyter container, this volume's is parametrized with `{username}`.
We have an extra container (to collect metrics) that also needs to mount this volume, but currently it's not possible, because the volume's name doesn't get expanded in the `extra_containers` section.


elgalu:
👍

LoicAG:
Hi @minrk , thanks for the quick merge! Do you already know when the next release will be?

minrk:
I think we should probably do one soon. I don't think we are in the middle of anything, so we can get started on it.

PR #278, opened 2018-11-02
changelog for 0.10

minrk:
Small changes, but the unpinning of the kubernetes package is important already.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -15,7 +15,7 @@ KubeSpawner 0.10
   instead, require kubernetes client >= 7.
   If desired, pinning should be done in images/installations
 - Expand username template variables in extra_containers
-- Set pod restart policy to OnFailure, so that notebook servers that crash restart automatically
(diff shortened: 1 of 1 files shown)


minrk:
Need to add #280 here, then I think we can ship it

docs/source/changelog.rst
```diff
@@ -6,6 +6,20 @@ Changes in KubeSpawner
...
+KubeSpawner 0.10
+================
+
+0.10 is a small release, with minor changes and fixes.
+
+- Stop pinning an exact kubernetes client version;
+  instead, require kubernetes client >= 7.
+  If desired, pinning should be done in images/installations
+- Expand username template variables in extra_containers
+- Set pod restart policy to OnFailure, so that notebook servers that crash restart automatically
```
clkao: I assume this is #251 ? actually crashed notebook servers did restart, the fix is for notebook-initiated cull to correctly stop the server, not triggering restart.

minrk: Thanks for the correction!

PR #280, opened 2018-11-07
deprecate KubeSpawner.image_spec

minrk:
use KubeSpawner.image so that config matches kubernetes terminology. In general, let's try to always have a perfect match of config name -> field name when there is a 1:1 correspondence.

The deprecated traits functionality is expanded to add arbitrary old->new mappings and deprecated versions.


yuvipanda:
Thanks, @minrk!

PR #285, opened 2018-12-14
using user_options in kubespawner

hhuuggoo:
Untested, and no tests written.  Just wanted to get validation that this approach is correct before doing all that.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1620,6 +1620,13 @@ class KubeSpawner(Spawner):
         start returns, which we can use to terminate
         .progress()
         """
+        for k, v in self.user_options.items():
+            if callable(v):
+                v = v(self)
+                self.log.debug(".. overriding KubeSpawner value %s=%s (callable result)", k, v)
+            else:
+                self.log.debug(".. overriding KubeSpawner value %s=%s", k, v)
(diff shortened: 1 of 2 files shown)


hhuuggoo:
Added tests.  Not sure what else is required, but seems to work ok

hhuuggoo:
```
>   request.addfinalizer(lambda: client.delete_namespace(kube_ns, {}))
E   TypeError: delete_namespace() takes 2 positional arguments but 3 were given
tests/conftest.py:49: TypeError
```
The tests are failing with this...  Which I don't believe is related?  Do you guys know what this is about?

Thanks!

yuvipanda:
@hhuuggoo I think the test failures should be fixed by https://github.com/jupyterhub/kubespawner/pull/292

hhuuggoo:
Well so I spoke with @minrk about this - and he pointed out that this could be dangerous (you might not want your users to be able to pass user options)

So not sure if y'all want to close it, or make it so this behavior is off by default?

minrk:
#301 fixes user_options loading to be consistent whether loaded via API or form, and safe by default (i.e. options are not passthrough to kubespawner attributes, they are interpreted by overrideable load_user_options method)

PR #292, opened 2019-01-28
Pin kubernetes version to 8.0

yuvipanda:
9.0 alpha has breaking changes, and our version selector
was including 9.0alpha.

Breaking changes listed here:

https://github.com/kubernetes-client/python/blob/master/CHANGELOG.md#v900a1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     install_requires=[
         'jupyterhub>=0.8',
         'pyYAML',
-        'kubernetes>=7',
+        'kubernetes==7.0.0',
         'escapism',
         'jinja2',
         'async_generator>=1.8',
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     install_requires=[
         'jupyterhub>=0.8',
         'pyYAML',
-        'kubernetes==7.0.0',
+        'kubernetes==8.0.*',
         'escapism',
         'jinja2',
         'async_generator>=1.8',


consideRatio:
@yuvipanda the PR commit pins it to 0.7.0 now, there is 8.0.1 out with a bugfix and a securityfix. Perhaps we can pin it so it allows for 8.0.1 and updates to the patch version, but not major or minor?

Perhaps it is written like `8.0.*`

setup.py
```diff
@@ -14,7 +14,7 @@
     install_requires=[
         'jupyterhub>=0.8',
         'pyYAML',
-        'kubernetes>=7',
+        'kubernetes==7.0.0',
```
choldgraf:
```suggestion
        'kubernetes==8.0.*',
```

choldgraf:
Just to add to @consideRatio's comment, I think this is what he means:

https://github.com/jupyterhub/kubespawner/pull/292/files#diff-2eeaed663bd0d25b7e608891384b7298R17

the z2jh helm chart uses an earlier version of kubespawner, right? So this would be a breaking change for v0.9?

consideRatio:
Max confusion over here in my brain department ;D This LGTM though haha

minrk:
Oops, missed this one while buried under grants. Packages shouldn't pin dependencies strictly like this, that should be the responsibility of deployments. Reverting and fixing the actual compatibility issue in #294

PR #294, opened 2019-02-06
compatibility with kubernetes 9.0

### LINKED ISSUE #288: CI tests are failing
See: https://travis-ci.org/jupyterhub/kubespawner/jobs/476452270 from the recent #287 pr that only changed documentation. I assume some dependency has been upgraded and changed an argument definition.

minrk:
signature change in kubernetes 9.0

only affects tests, so this doesn't need to prompt a new release

closes #288
reverts #292

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ python:
 # install dependencies
 install:
   - pip install --upgrade setuptools pip
-  - pip install --pre -e ".[test]"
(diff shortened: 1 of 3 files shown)


minrk:
I added a note to travis.yml about why we test with prereleases of dependencies, which seemed to cause some confusion that prompted #292. The failing test was exactly what was supposed to happen so that we could catch and fix the issue before it made it into a real release.

PR #296, opened 2019-02-06
Update Kubernetes python client to 8.0.1

lresende:
Update the Kubernetes python client dependency declaration to
use version 8.0.1 or greater (but below 9.x). The Version 8.0.1
brings security fix for CVE-2018-20060.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,12 +12,12 @@ setup(
     name='jupyterhub-kubespawner',
     version='0.10.2.dev',
     install_requires=[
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for the PR! This is already fixed by #294, though, which removes the pinning altogether since libraries shouldn't be pinning dependencies strictly like this.

The current version in master will also get 8.0.1, since that's the current version that matches the 8.0.*

PR #297, opened 2019-02-08
Update documentation regarding run_as_gid behavior

### LINKED ISSUE #290: runAsGroup behaviors yield effective group of root
I'm not sure if this is considered an issue or not, but thought it best to at least raise it.  

While adding runAsUser and runAsGroup capability to [Enterprise Gateway](https://github.com/jupyter/enterprise_gateway) (where kernels can be launched as single pods), I found a couple issues surrounding `runAsGroup` so started looking around for other uses.  Since I didn't see any specific documentation about these issues in kubespawner, I can't tell if others are aware and I'd like EG to handle things as similarly as possible.  I had a side-bar with @consideRatio and he suggested a broader discussion.

_**tl;dr:** Unless RunAsGroup feature gate is enabled AND both runAsUser and runAsGroup are provided, pods will run with effective group Id of 0 (root).  Seems like this should be documented at a minimum.  We could also ensure both values are provided, but there doesn't appear to be a way to determine feature-gate settings via the API._

Here are the issues:
1. `runAsGroup` is only recognized if the feature-gate `RunAsGroup=true`.  As of 1.13.2, this feature-gate is still false by default.  As a result, it's value is ignored, but worse is that the container's primary group is not used, and root is used instead. Here's a screen shot from an EG kernel pod...
both not enabled
2. Even if feature-gate `RunAsGroup=true`, if `runAsGroup` is NOT provided, the container's primary group is still not used...
runasuser enabled

Only if feature-gate `RunAsGroup=true` AND BOTH runAsUser and runAsGroup are provided is the expected result produced...
both enabled

In addition, runAsGroup cannot be specified if runAsUser is NOT specified and feature-gate is enabled.  This results in a pod creation failure.  However, if the feature-gate is not enabled, then runAsGroup is ignored - resulting in the container's built-in UID/GID being used (which is good)...
runasgroup not enabled

As a result, it seems that you only achieve expected behavior if both runAsUser and runAsGroup are provided AND feature-gate RunAsGroup=true.  Anyone not wanting their containers to run as the root group will likely be surprised and should probably be aware of this.

kevin-bates:
Added brief information for `run_as_gid` that it requires enablement of
feature-gate RunAsGroup and how various settings could yield effective
GIDs of root.

Fixes #290

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -99,6 +99,10 @@ def make_pod(
     run_as_gid:
         The GID used to run single-user pods. The default is to run as the primary
         group of the user specified in the Dockerfile, if this is set to None.
+        Setting this parameter requires that *feature-gate* **RunAsGroup** be enabled,
+        otherwise the effective GID of the pod will be 0 (root).  In addition, not
+        setting `run_as_gid` once feature-gate RunAsGroup is enabled will also
+        result in an effective GID of 0 (root).
     fs_gid
         The gid that will own any fresh volumes mounted into this pod, if using
         volume types that support this (such as GCE). This should be a group that


minrk:
Thanks!

meeseeksmachine:

https://discourse.jupyter.org/t/what-is-with-the-weird-jovyan-user/1673/12

PR #304, opened 2019-03-06
Enable pods to be launched on user namespace

lresende:
Enables namespace template via `namespace_name_template` which
causes pods and other resources to be created on the expanded
user namespace.

`namespace_name_template` example: `hub-{username}`

Note: In order for this functionality to work, the service account
used to manage the JupyterHub deployment needs to be properly configured
globally using ClusterRole and ClusterRoleBinding.

closes #218

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -124,7 +124,7 @@ def make_user_rules():
                      resources=['events'],
                      verbs=['get', 'list', 'watch']),
         V1PolicyRule([''],
-                     resources=['pods', 'persistentvolumnes', 'persistentvolumeclaims'],
+                     resources=['pods', 'persistentvolumes', 'persistentvolumeclaims'],
                      verbs=['get', 'list', 'watch']),
     ]
 
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -3,28 +3,170 @@ Helper methods for generating k8s API objects.
 """
 import json
 from urllib.parse import urlparse
-import escapism
 import re
-import string
 from kubespawner.utils import get_k8s_model, update_k8s_model
 
 from kubernetes.client.models import (
-    V1Pod, V1PodSpec, V1PodSecurityContext,
-    V1ObjectMeta,
-    V1LocalObjectReference,
-    V1Volume, V1VolumeMount,
-    V1Container, V1ContainerPort, V1SecurityContext, V1EnvVar, V1ResourceRequirements, V1Lifecycle,
-    V1PersistentVolumeClaim, V1PersistentVolumeClaimSpec,
-    V1Endpoints, V1EndpointSubset, V1EndpointAddress, V1EndpointPort,
-    V1Service, V1ServiceSpec, V1ServicePort,
-    V1beta1Ingress, V1beta1IngressSpec, V1beta1IngressRule,
-    V1beta1HTTPIngressRuleValue, V1beta1HTTPIngressPath,
-    V1beta1IngressBackend,
-    V1Toleration,
     V1Affinity,
-    V1NodeAffinity, V1NodeSelector, V1NodeSelectorTerm, V1PreferredSchedulingTerm, V1NodeSelectorRequirement,
-    V1PodAffinity, V1PodAntiAffinity, V1WeightedPodAffinityTerm, V1PodAffinityTerm,
+    V1beta1HTTPIngressPath,
+    V1beta1HTTPIngressRuleValue,
+    V1beta1Ingress,
+    V1beta1IngressBackend,
+    V1beta1IngressRule,
+    V1beta1IngressSpec,
+    V1ClusterRole,
+    V1ClusterRoleBinding,
+    V1Container,
+    V1ContainerPort,
+    V1EndpointAddress,
+    V1EndpointPort,
+    V1Endpoints,
+    V1EndpointSubset,
+    V1EnvVar,
+    V1Lifecycle,
+    V1LocalObjectReference,
+    V1Namespace,
+    V1NodeAffinity,
+    V1NodeSelector,
+    V1NodeSelectorTerm,
+    V1ObjectMeta,
+    V1PersistentVolumeClaim,
+    V1PersistentVolumeClaimSpec,
+    V1Pod,
+    V1PodAffinity,
+    V1PodAffinityTerm,
+    V1PodSecurityContext,
+    V1PodSpec,
+    V1PolicyRule,
+    V1PreferredSchedulingTerm,
+    V1ResourceRequirements,
+    V1Role,
+    V1RoleBinding,
+    V1RoleRef,
+    V1SecurityContext,
+    V1Service,
+    V1ServiceAccount,
+    V1ServicePort,
+    V1ServiceSpec,
+    V1Subject,
+    V1Toleration,
+    V1Volume,
+    V1VolumeMount,
+    V1WeightedPodAffinityTerm,
 )
+
+
+def make_namespace(name):
+    """
+    Make a k8s namespace specification for launching pods
+    :param name: namespace name
+    """
+    namespace_metadata = V1ObjectMeta(name=name)
+
+    return V1Namespace(metadata=namespace_metadata)
+
+
+def make_service_account(name):
+    """
+    Make a k8s service account specification
+    :param name: service account name
+    """
+    service_account_metadata = V1ObjectMeta(name=name)
+    return V1ServiceAccount(metadata=service_account_metadata)
+
+
+def make_cluster_role(name):
+    """
+    Make a k8s role specification
+    :param name: role name
+    :return:
+    """
+
+    role_metadata = V1ObjectMeta(name=name)
+
+    rules = [
+        V1PolicyRule([''],
+                     resources=['events'],
+                     verbs=['get', 'list', 'watch']),
+        V1PolicyRule([''],
+                     resources=['pods', 'persistentvolumnes', 'persistentvolumeclaims'],
+                     verbs=['get', 'list', 'watch']),
+    ]
+
+    return V1ClusterRole(metadata=role_metadata, rules=rules)
+
+
+def make_cluster_role_binding(namespace, name, service_account):
+    """
+    Make a k8s role binding specification
+    :param namespace: namespace where the role will be created
+    :param name: role binding name
+    :param service_account: service account name
+    """
+    role_binding_name = name
+    role_binding_metadata = V1ObjectMeta(name=role_binding_name)
+    role_binding_role_ref = V1RoleRef(api_group='', kind='ClusterRole', name=name)
+    role_binding_subjects = V1Subject(api_group='', kind='ServiceAccount', namespace=namespace, name=name)
+
+    return V1ClusterRoleBinding(kind='ClusterRoleBinding',
+                         metadata=role_binding_metadata,
+                         role_ref=role_binding_role_ref,
+                         subjects=[role_binding_subjects])
+
+
+def make_user_rules():
+    """
+    Make RBAC rules for user
+    """
+
+    return [
+        V1PolicyRule([''],
+                     resources=['events'],
+                     verbs=['get', 'list', 'watch']),
+        V1PolicyRule([''],
+                     resources=['pods', 'persistentvolumnes', 'persistentvolumeclaims'],
+                     verbs=['get', 'list', 'watch']),
+    ]
+
+
+def make_user_role(name):
+    """
+    Make RBAC user role
+    :param name:
+    """
+    role_metadata = V1ObjectMeta(name=name)
+    rules = make_user_rules()
+
+    return V1Role(metadata=role_metadata, rules=rules)
+
+
+def make_role(name, rules):
+    """
+    Make a k8s role specification
+    :param name: role name
+    """
+
+    role_metadata = V1ObjectMeta(name=name)
+
+    return V1Role(metadata=role_metadata, rules=rules)
+
+
+def make_role_binding(namespace, name, service_account):
+    """
+    Make a k8s role binding specification
+    :param namespace: namespace where the role will be created
+    :param name: role binding name
+    :param service_account: service account name
+    """
+    role_binding_name = name
+    role_binding_metadata = V1ObjectMeta(name=role_binding_name)
+    role_binding_role_ref = V1RoleRef(api_group='', kind='Role', name=name)
+    role_binding_subjects = V1Subject(api_group='', kind='ServiceAccount', namespace=namespace, name=service_account)
+
+    return V1RoleBinding(kind='RoleBinding',
+                         metadata=role_binding_metadata,
+                         role_ref=role_binding_role_ref,
+                         subjects=[role_binding_subjects])
 
 def make_pod(
     name,
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -1,5 +1,6 @@
 # specifically use concurrent.futures for threadsafety
 # asyncio Futures cannot be used across threads
+
 from concurrent.futures import Future
 
 import time
@@ -14,7 +15,7 @@ from urllib3.exceptions import ReadTimeoutError
 
 from .clients import shared_client
 
-class NamespacedResourceReflector(LoggingConfigurable):
+class ResourceReflector(LoggingConfigurable):
     """
     Base class for keeping a local up-to-date copy of a set of kubernetes resources.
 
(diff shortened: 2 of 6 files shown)


kubespawner/reflector.py
```diff
@@ -195,8 +188,8 @@ def _watch_and_update(self):
         cur_delay = 0.1
 
         self.log.info(
-            "watching for %s with %s in namespace %s",
```
minrk: Shouldn't this be only for `if not self.namespace`?

kubespawner/reflector.py
```diff
@@ -150,12 +139,16 @@ def _list_and_update(self):
...
         Overwrites all current resource info.
         """
-        initial_resources = getattr(self.api, self.list_method_name)(
-            self.namespace,
-            label_selector=self.label_selector,
-            field_selector=self.field_selector,
-            _request_timeout=self.request_timeout,
-        )
+        list_args = {
+            'label_selector': self.label_selector,
```
minrk: This should include `self.namespace` if self.namespace, right?

kubespawner/reflector.py
```diff
@@ -150,12 +139,16 @@ def _list_and_update(self):
...
+            'field_selector': self.field_selector,
+            '_request_timeout': self.request_timeout
+        }
+
+        print(list_args)
```
minrk: debug statement

kubespawner/spawner.py
```diff
@@ -53,17 +60,17 @@ class PodReflector(NamespacedResourceReflector):
...
     def pods(self):
         return self.resources
 
 
-class EventReflector(NamespacedResourceReflector):
+class EventReflector(ResourceReflector):
     kind = 'events'
 
-    list_method_name = 'list_namespaced_event'
+    list_method_name = 'list_event_for_all_namespaces'
```
minrk: The list method should depend on whether a namespace is set, right?

kubespawner/spawner.py
```diff
@@ -1640,6 +1790,9 @@ def _start(self):
         if events:
             self._last_event = events[-1].metadata.uid
 
+        if not self._has_namespace(self.namespace):
```
minrk: avoid calling this in the default case:

```suggestion
        if self.namespace_name_template and not self._has_namespace(self.namespace):
```

kubespawner/spawner.py
```diff
@@ -1379,6 +1418,118 @@ def get_pvc_manifest(self):
...
+        try:
+            client.RbacAuthorizationV1Api().create_namespaced_role_binding(namespace=self.namespace, body=role_binding_metadata)
+        except ApiException as e:
+            if e.status == 409:
+                # ignore when resource already exists
+                pass
+            else:
+                self.log.debug("Error creating user RoleBinding '{}'".format(user_role_binding))
+                raise e
+        self.log.debug('User RoleBinding created.')
```
minrk: Maybe include name in log statement

kubespawner/spawner.py
```diff
@@ -1379,6 +1418,118 @@ def get_pvc_manifest(self):
...
+        try:
+            client.RbacAuthorizationV1Api().create_namespaced_role(namespace=self.namespace, body=role_metadata)
+        except ApiException as e:
+            if e.status == 409:
+                # ignore when resource already exists
+                pass
+            else:
+                self.log.debug("Error creating role '{}' for namespace '{}'".format(user_role_name, self.namespace))
+                raise e
+        self.log.debug('Role created.'.format())
```
minrk: no args passed to format

kubespawner/objects.py
```diff
@@ -3,29 +3,171 @@
...
+    """
+    Make RBAC rules for user
+    """
+
+    return [
+        V1PolicyRule([''],
+                     resources=['events'],
+                     verbs=['get', 'list', 'watch']),
+        V1PolicyRule([''],
+                     resources=['pods', 'persistentvolumnes', 'persistentvolumeclaims'],
```
minrk:
```suggestion
                     resources=['pods', 'persistentvolumes', 'persistentvolumeclaims'],
```

kubespawner/spawner.py
```diff
@@ -1313,6 +1347,11 @@ def get_pod_manifest(self):
         else:
             real_cmd = None
 
+        if self.namespace_name_template:
+            service_account = self.namespace
```
minrk: Maybe the name shouldn't exactly *be* the namespace, but derived from it?

lresende:
The changes to hub RBAC updates are basically two things: make role/role binding actually be a cluster role/cluster role binding and add capabilities to create new namespaces, service account, role, role binding. My deployment scripts have [a version](https://github.com/lresende/ansible-kubernetes-cluster/blob/master/roles/jupyterhub/templates/jupyterhub-rbac.yaml.j2) that I have been using, that could probably be revised to remove any no required permissions.

Let me respond to the other comments inline.

kubespawner/spawner.py
lresende: There are two options here. 
- Always search globally without the namespace scope (which is the current implementation and `might` have a small performance penalty compared to looking up in a namespace)
- Conditionally use different APIs depending on how namespaces are being used (this would be similar to what was implemented in #218 , and based on the current implementation (as we always have a namespace), it will need a bunch of subclasses/ifs and I felt the code was a little convoluted) but I could go back to revisit that if you would prefer that.

kubespawner/reflector.py
lresende: good catch, It was also left over that should be removed

kubespawner/reflector.py
lresende: see comments below about different implementation approaches.

kubespawner/spawner.py
lresende: In EG we use the pattern where we suffix the RBAC resources with something that indicates what it is, but in this case, I was following the same pattern as JupyterHub that creates hub as the name of the namespace, service account, role, role binding. Should I depart from the JupyterHub pattern ?

lresende:
@minrk @consideRatio  Please let me know if you are ok with this direction, and how we could handle the z2jh possible changes...

consideRatio:
@lresende sorry for not getting back to you quicker, I'm not able to do this during working hours so I do it on my spare time which I've had little of lately. I'll get it done though!

consideRatio:
Forgive me for failing to review this in a timely manner @lresende, I'm working hard to prepare for "Hack for Sweden" =/

lresende:
NO worries @consideRatio, when you or @minrk have a few minutes please take a look, if a quick online review is better we can schedule something.

kubespawner/spawner.py
consideRatio: Thanks for working on this Luciano! I've read it through but not tested it.

Is it correct that not using this feature, due to this part will currently still force the privilege escalation of the pod running KubeSpawner to have ClusterRole permissions?

This may become a deal breaker for various users of KubeSpawner and Z2JH that relies on KubeSpawner. I know clusterrole wide permission can be a big ask for various administrators in general. For those that this isn't an issue, it may also require additional labor.

The more I think about it, the more afraid I'd be to click merge on something that require privilege escalation no matter what instead of making it something one opts into if needing to use a feature that requires it.

lresende: One would only require to use the cluster-wide roles if kubespawner is creating the per-user namespaces. Otherwise, the pods would be running on the already created 'hub' namespace and the regular `role/rolebinding` would be sufficient. 

One aspect is that the current implementation does try to create the 'user namespace' if the new function is enabled which will fail if the `cluster role` is not available. What I could do is to enhance the implementation to first check if the namespace is already present before trying to create, this would enable `bring your own namespace` where the admins would previously create the namespaces for the set of users without requiring the additional `cluster roles`

lresende: I could also provide a quick table (validated with some tests), with what works with `role` versus `cluster role`.

consideRatio: > One would only require to use the cluster-wide roles if kubespawner is creating the per-user namespaces. Otherwise, the pods would be running on the already created 'hub' namespace and the regular role/rolebinding would be sufficient.

Hmmm, but right now we are are invoking a different method on the `kubernetes` python package, that refers to listing all pods and all events etc rather than those in the namespace of kubespawner / hub itself right? I don't think we would be permitted to do that no matter if we create pods only within our current namespace.

This is the role attached to the hub service account currently through a RoleBinding, that is in use by the hub pod, that kubespawner runs within. It is not going to permit anything outside the namespace, so I wonder even if you are allowed to list clusterroles (`kubectl get clusterrole`) or pods in all namespaces.

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  creationTimestamp: 2019-04-01T22:14:42Z
  labels:
    app: jupyterhub
    chart: jupyterhub-0.9-61cf357
    component: hub
    heritage: Tiller
    release: jhub
  name: hub
  namespace: jhub
  resourceVersion: "714609"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/jhub/roles/hub
  uid: 8c77a641-54cb-11e9-9277-42010aa600ec
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - persistentvolumeclaims
  verbs:
  - get
  - watch
  - list
  - create
  - delete
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - get
  - watch
  - list
```

I figure we need the kubespawner/hub pod to run with additional permissions granted to it by default if we invoke a kubernetes api call to get stuff from all namespaces as default. When we set the method to `list_event_for_all_namespaces` instead of `list_namespaced_event`, it is my understanding that the `kubernetes` python package that helps us make kubernetes API calls like those run by in `kubectl` the CLI will demand it.

If you got this code working on your deployment, perhaps it is due to having a custom service account on the hub pod with powerful permissions?

lresende: @consideRatio Let me revisit this and make sure I document what works with `Role` versus `Cluster Role` as it has been a couple of weeks.

agt: @lresende thanks picking up the ball with this effort!   We've been running #218 in production since last September on top of z2jh, with the following ClusterRole and ClusterRoleBindings in place:   

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: hub
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - events
  - persistentvolumeclaims
  verbs:
  - get
  - watch
  - list
  - delete
  - create
```

and

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: hub
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: hub
subjects:
- kind: ServiceAccount
  name: hub
  namespace: jupyterhub
```

lresende:
Thanks @minrk and @consideRatio, I am traveling this week and might find a few hours to take a look at this next week before my next trip. Hopefully, I can clean this up and get both scenarios working very soon.

GrahamDumpleton:
FWIW, we have a much more elaborate scheme already implemented whereby a separate namespace can be created dynamically as necessary, and as well as setting up service account access to the namespace, you can optionally deploy a set of resources into the namespace as well, plus set limit ranges and quotas. Further, if it was an anonymous session, you can have the whole namespace deleted some time after the notebook instance idles out.

We have all this, not for spawning up Jupyter notebooks, but workshop training environments. One can drop the original Jupyter notebook image back in though and it still works. The result is that you could setup an instance where you could do something like spawn your Jupyter notebook, along with the separate namespace, and in the namespace deploy a Dask cluster on demand, with the Jupyter notebook hooked into it automatically.

It isn't quite the same as what you are trying to do here, as the notebook still runs in the same namespace as the spawner, but all the workloads you might want to deploy automatically, or from the notebook, such as Dask, or Nextflow, would be in the separate per user namespace, meaning it is much easier to give them a larger quota or set other restrictions on them.

We have configurations for a range of use cases. If curious, you can dig through the different configurations at:

https://github.com/openshift-labs/workshop-spawner/tree/develop/jupyterhub/.jupyter

I know in general there is probably no interest in what we have done, but still thought I would mention it.

athornton:
Also wondering what the status here is; I've been working on a large refactoring of my spawning framework, and would love to stop carrying my own multi-namespace work.

betatim:
Given how old the PR is I think it would be Ok for someone new to take this over by opening a new PR based on the commits in this PR (to preserve authorship and credit). There are a few comments that aren't marked as resolved so we'd have to go through those to see what is what, resolve the merge conflicts (see https://github.com/jupyterhub/kubespawner/pull/304#issuecomment-542722989) and then restart the review of the new PR.

Do you want to tackle that @athornton ?

consideRatio:
Note that both #218 and this pr #304 attempts to address the user-namespace feature. I think a good action point is to attempt to specify the changes suggested clearly.

For example:
- what is the default behavior meant to be, one namespace or multiple? Going multiple by default would cause a lot of disruptions so I assume we need to have the default be one namespace.
- what is the planned kubespawner way to work with a multiple namespaces? Switch a single boolean flag? Something more?
- how do we plan to support the mutli namespace feature? We are using a Singleton pattern of NamespacedResourceReflectors to monitor for the events/pods in a specific namespace currently, then we filter out the resources relevant for specific pods or events belonging to a specific pod. Should we start to monitor the entire cluster instead of specific namespaces, or should we create one cluster-scoped-resource-reflector instead?

manics:
Another factor is what RBAC permissions are required. Ideally kubespawner should only require permissions on a namespace (`Role`) unless it's configured to use multiple namespaces (`ClusterRole`).

lresende:
> Another factor is what RBAC permissions are required. Ideally kubespawner should only require permissions on a namespace (`Role`) unless it's configured to use multiple namespaces (`ClusterRole`).

AFAIK (remember), this is the only remaining thing that needs to be fixed.

lresende:
As for my availability, I could try to come back to this during my end of year vacation :) but it's going to be hard for me before that.

athornton:
....maybe?  Let me bring it up to my boss today and see if she is OK with my tackling it (we have a lot of stuff of our own we're trying to close out by the end of the year).

Default should be a single namespace to preserve existing behavior and configs.

What I've been doing is my own subclass, "MultiNamespacedKubeSpawner" which has the KubeSpawner API but supports spawning pods into individual namespaces.  That way, we can just document the behavior and what additional (scary) RBAC privileges it needs, but most people will be fine continuing to use the old KubeSpawner.  I don't think there's really a need to fold the multi-namespace stuff into the same class.  I could be wrong about that.

Currently it doesn't have much in the way of automatic support; I'm further subclassing it with an LSSTSpawner class that adds flags like "delete_namespace_on_stop" and some brittle and ugly PV/PVC manipulators (which I do not recommend pushing upstream) to support the LSST use cases.  But that is certainly fine for discussion on a subsequent PR.

athornton:
I have the boss's OK to take this on.

Do we want to roll the support into the base KubeSpawner class, or leave it alone and add a MultiNamespaceKubeSpawner ?

betatim:
Inheritance is a cool thing, I feel like we should use it. Especially because it doesn't look like you would want to "casually" enable this. You'd have to think about it a bit and understand what you are doing, changing the class name is probably one of the least worries in this whole story.

Thanks a lot for asking the boss and getting the Ok! Both to you for asking and to the boss for saying yes!

athornton:
The merge conflicts were pretty substantial.  I suspect I'm going to end up with something that looks more like #218 did (which was less intrusive to the base classes than #304).  Anyway, I'll open a new PR when I have something that kind of works.  (I'll start from this PR, though, so commit history is preserved).

It may be a while; I'm juggling several other things right now as well.

clkao:
@athornton did you check the recently rebased branch https://github.com/InfuseAI/kubespawner/tree/namespaces-rebased ? there are still new conflict but should be better.

athornton:
Oh cool!  I'll do that on Monday.

athornton:
Well, a bunch of year-end stuff caught fire.  I got very lost trying to reconcile the various proposals, so my initial PR is going to be a transplantation of the stuff I've done in lsst-sqre/jupyterhubutils, which is basically @agt 's work from #218 , back into kubespawner.  There's a FIXME for adding user template names back, and I can easily put in the namespace manipulation methods again.  

The way I'm doing that all of the namespaced object creation/deletion is done in an LSST-specific business logic class that I mix into the spawner, but hopefully we can figure out (collectively) where it belongs.  

#385

elgalu:
This feature seems very useful from a SEC PoV, wondering if there are plans to resolve the conflicts and move it FWD.

athornton:
I have a PR for user namespaces that works against current JupyterHub Kubespawner: https://github.com/jupyterhub/kubespawner/pull/458

It creates the namespace for you, but you could certainly subclass it and make the method that creates the namespace just check for existence.  Or we could put that behind a config boolean.

minrk:
closed by #458

PR #309, opened 2019-03-22
Fix pod name prefix escaping for named servers

### LINKED ISSUE #300: Named server user properties expansion
In case named servers are enabled, the server name is expanded in the user properties as follows:
https://github.com/jupyterhub/kubespawner/blob/cae58559fb672c8470d7b252f3518ea16f1ce839/kubespawner/spawner.py#L1228-L1231
The name is always prefixed with `-` before escaping. As `-` is not part of `safe_chars`, it is escaped to `-2d`. For example, a server name `test` is expanded to `-2dtest`. Is this behavior really intended? Could someone please explain why there is a general prefix for the server name at all?

dmarth:
Add prefix after escaping the server name (#300).

Fixes #300

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1247,7 +1247,7 @@ class KubeSpawner(Spawner):
         # Set servername based on whether named-server initialised
         if self.name:
             servername = '-{}'.format(self.name)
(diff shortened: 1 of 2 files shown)


consideRatio:
LGTM @dmarth and thanks for adding tests as well!

There is a merge conflict to be resolved, I'll go to my Linux computer and inspect it.

PR #310, opened 2019-04-01
Configurable pod termination grace period

### LINKED ISSUE #254: Make stop -> grace_seconds configurable
Right now `grace_seconds` for stopping a pod is either 0 or 1 second. I think is is quite a low number, the reason that is probably 30 seconds by default is because there should be enough time for a graceful container shutdown, if for example the singleuser containers run some services like a per-user Postgres or whatever.

Perhaps it would be valuable to have this configurable somehow.

https://github.com/jupyterhub/kubespawner/blob/5040e75022f302b3cbfadaa1ae18a5ab9d5563a0/kubespawner/spawner.py#L1760

arturozv:
Attempt to close #254, any feedback is appreciated

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1106,6 +1106,21 @@ class KubeSpawner(Spawner):
         """
     )
 
+    termination_grace_period = Integer(
+        1,
+        config=True,
+        help="""
+        Time in seconds for the pod to be in `terminating` state before is forcefully killed.
+        
+        Increase this if you need more time to execute a `preStop` lifecycle hook.
+
+        See https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods for
+        more information on how pod termination works.
+
+        Defaults to `1`.
+        """
+    )
+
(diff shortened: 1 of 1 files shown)


consideRatio:
@arturozv :heart: :tada: :100:!

We have not done great in the past with this, but we have learned our lesson that it is good when things are named exactly like in the kubernetes api, for example that if you want to configure something in the kubernetes pod specification, kubespawner names the setting exactly the same.

@minrk what do you think the traitlet (the configurable option) should be named?

I did not grasp the error right away sadly and now it is time to sleep for me.

arturozv:
That makes sense @consideRatio, let me know what name would you like 😃  

My thought for it was `{context}_{kubernetesExactName}` to make it more self explanatory.  I was thinking on the case of someone (me, for example) reading a jhub config file without having a strong knowledge of the pod lifecycle. `c.KubeSpawner.termination_grace_period` or `c.KubeSpawner.delete_grace_period` would make a lot of sense

minrk:
delete_grace_period sounds perfect, thanks

consideRatio:
@arturozv wieeee thanks!

PR #312, opened 2019-04-04
codecov badge

choldgraf:
adds a codecov badge so we know how much test coverage we've got


choldgraf:
added a docs badge while we were at it

betatim:
What do we do about the failed tests? New PR to try and get them fixed? (I just restarted the builds on the chance that this is a flaky test)

minrk:
Failed tests are due to prereleases of dependencies: jupyterhub 1.0 beta and kubernetes 10 alpha. They are probably small updates needed, but we do need to update the repo to get the tests passing again.

PR #327, opened 2019-06-05
Add image data to state exposed via API

vilhelmen:
Looking at [#2598 over at JH](https://github.com/jupyterhub/jupyterhub/issues/2598), it would be beneficial to add more runtime data to the data exposed through the API to allow for more flexible services.

Right now, this only adds the image data to the state dict. I'm wary of the implications of loading that field and overriding configuration since I'm not as familiar with how that will affect the spawner process. If it's not going to break anything, that can be added as well.

Still need to check if it even records valid data in combination with profile_list and options_form overrides.

Update: Looks to work fine w/ options_form and profile_list overrides.

It also appears that Kubespawner already exposes the selected profile name under `user_options` in the JH db. Perhaps the `/users` API endpoint should be corrected to return that data instead. Opinions? The selected profile name may be a better, more flexible choice as it's going to be independent from the actual image (and tag).

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1424,6 +1424,7 @@ class KubeSpawner(Spawner):
         """
         state = super().get_state()
         state['pod_name'] = self.pod_name
+        state['image'] = self.image
         return state
 
     def get_env(self):


kubespawner/spawner.py
```diff
@@ -1424,6 +1424,7 @@ def get_state(self):
         """
         state = super().get_state()
         state['pod_name'] = self.pod_name
+        state['image'] = self.image
```
rkdarst: This is sort of a hack, but what about adding
```
if hasattr(self, 'extra_state'):
    state['extra'] = self.extra_state
```
That way, if there's no explicitly defined attribute, nothing changes.  If someone defines `extra_state` in `kubespawner_override`, it'll be passed to state and can be used to cull.

But... this is a textbook definition of "quick hack that can't ever be removed"... any better ideas?

consideRatio:
@minrk @rkdarst --- can I get your help to review this PR?

kubespawner/spawner.py
vilhelmen: Oh wow, sorry for the delay. Maybe? I think extra_state is supposed to be merged into the spawner by the time this function is called, but of course that means things not explicitly dumped will disappear. I'm not sure if there's a better solution. I'm not sure, but won't hasattr always return true because `extra_state` is a configurable? Might want to check hasattr and then if it's not None and/or empty?

meeseeksmachine:

https://discourse.jupyter.org/t/binder-jupyterhub-activity-round-up-week-1/2704/1

consideRatio:
I find this hard to review, it is a one liner about something I lack the overview to figure out well, help to review this wanted!

vilhelmen:
To be fair, I'm even mixed on this PR. The driver behind this is really that there's no way to provide a smarter culler service that can alter time limits based on containers or container properties. Placing the image name in the saved state data exposes it via JH's API for others to act upon. JH issue here: https://github.com/jupyterhub/jupyterhub/issues/2598

We appear to already store the user's selected options_form entry, but it goes into a separate dictionary that is not exposed by the API for whatever reason. A discussion on adding that data to what is exposed via the API didn't really go anywhere. There were concerns over the sensitivity of the data, but the data is only accessible to administrators so it's moot in my opinion.

I suggested moving the culler service internally where it can access everything it could possibly need without duplicate (and therefore brittle) secondary configuration and submitted a POC but we didn't go in that direction, leaving us without a solution or route to one. https://github.com/jupyterhub/jupyterhub/pull/2664

Adding this to the state data is a step towards something configurable to achieve container-based time limits, but I will admit that this does not personally feel like the best solution, as the image name is going to be big and noisy wrt tags and versions and whatnot.

consideRatio:
I'm sorry for not helping this PR resolve faster, I think maybe a strategy to accomplish something like this can be do use a hook (`pre_spawn_hook`?) to update the user authentication state in any way based on the the spawner state.

I'll go for a close here, since its a very old PR thats a one-liner change as well.

yuvipanda:
https://github.com/jupyterhub/jupyterhub/issues/4406 is kinda a more generalized ask of this category of problems I think.

PR #329, opened 2019-06-20
fix: avoid escaping valid characters in server names

### LINKED ISSUE #324: username for the form "aaa-bbb" gets PVC like: "claim-aaa-e2bbb-XXXX"
I suggest to keep the "-" character like this. 

I can make a PR if accepted.

johanfleury:
Avoid escaping uppercase letters (they're transformed to lowercase) and
`-` (which is a valid character) in server name.

Close #324.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1241,19 +1241,20 @@ class KubeSpawner(Spawner):
 
     def _expand_user_properties(self, template):
         # Make sure username and servername match the restrictions for DNS labels
-        # Note: '-' is not in safe_chars, as it is being used as escape character
-        safe_chars = set(string.ascii_lowercase + string.digits)
+        safe_chars = set(string.ascii_letters + string.digits)
 
         # Set servername based on whether named-server initialised
         if self.name:
             servername = '-{}'.format(self.name)
-            safe_servername = '-{}'.format(escapism.escape(self.name, safe=safe_chars, escape_char='-').lower())
+            safe_servername = '-{}'.format(
+                escapism.escape(self.name, safe=safe_chars, escape_char='-', allow_collisions=True).lower()
+            )
         else:
             servername = ''
             safe_servername = ''
 
         legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
-        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
+        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-', allow_collisions=True).lower()
         return template.format(
             userid=self.user.id,
             username=safe_username,
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -289,7 +289,7 @@ def test_pod_name_escaping():
 
     spawner = KubeSpawner(config=c, user=user, orm_spawner=orm_spawner, _mock=True)
 
-    assert spawner.pod_name == "jupyter-some-5fuser-test-2dserver-21"
+    assert spawner.pod_name == "jupyter-some-5fuser-test-server-21"
 
 
 def test_pod_name_custom_template():


johanfleury:
No sure if the failing test is related to my changes.

kubespawner/spawner.py
```diff
@@ -1241,19 +1241,20 @@ def _set_deprecated(name, new_name, version, self, value):
...
         # Make sure username and servername match the restrictions for DNS labels
-        # Note: '-' is not in safe_chars, as it is being used as escape character
-        safe_chars = set(string.ascii_lowercase + string.digits)
+        safe_chars = set(string.ascii_letters + string.digits)
 
         # Set servername based on whether named-server initialised
         if self.name:
             servername = '-{}'.format(self.name)
-            safe_servername = '-{}'.format(escapism.escape(self.name, safe=safe_chars, escape_char='-').lower())
+            safe_servername = '-{}'.format(
```
betatim: I think it would be good to leave a comment here saying something like "if you ever run into weird bugs when unescape()-ing the servername `allow_collisions=True` is why". Basically some kind of hint that we decided on purpose to make unscaping fragile as a trade off for nicer server names.

betatim:
Restarted the tests and left a comment about adding a comment.

kubespawner/spawner.py
johanfleury: I can totally do that, but I can't find call to `unescape` anywhere in the code. Is `unescape()`-ing something that could really happen?

betatim: I couldn't find any calls either (also not in the projects that I know that use kubespawner). This is more of a reminder for people from the future who come cursing because they started using `unescape()` and it is failing on them in weird ways. A comment here that says we made this decision on purpose and that broken unescpae is "by design" might help them avoid hours of debugging time.

minrk: allow_collisions isn't used for unescape roundrip, though that is one effect. It's used for safety. Without it, you can have two valid usernames map onto the same value. Take for example:

```
In [8]: safe_chars = string.ascii_letters + string.digits

In [9]: escape('a.b', safe=safe_chars, escape_char='-', allow_collisions=True).lower()
Out[9]: 'a-2eb'

In [10]: escape('a-2eb', safe=safe_chars, escape_char='-', allow_collisions=True).lower()
Out[10]: 'a-2eb'
```

Both `a.b` and `a-2eb` are valid users, but both are going to have the same pod name `jupyter-a-2eb`!
The escape character cannot be considered a safe character because that allows collisions across usernames. This in general makes all usernames that have a character that needs escaping vulnerable to collisions from other users (or other server names). In the right circumstances, this can enable malicious users access to other users' private user data if users can pick their own username.

A more robust solution is to make a less common but still valid character the escape. I don't have the lists of safe characters for various fields in front of me, but maybe we can use `_` or something.

On the other hand, if *all* names are assumed safe, then we can skip escaping altogether and just let the errors raise through for invalid names. This is something we could do in e.g. BinderHub, and defer to the components picking the names to pick safe ones. I think allowing disabling escaping altogether is probably preferable to allowing collisions.

johanfleury: We're not escaping usernames here, only server names. The main goal is to avoid having pod or PVC names like `jupyter-notebook-username-foo-2dbar` when the user created a server named `foo-bar`.

yuvipanda:
Apologies for PR necromancy! Coming back slowly from a time out...

Yeah, this is a safety issue than anything else - particularly the collision that @minrk talks about in https://github.com/jupyterhub/kubespawner/pull/329#discussion_r305745331. I'm also hesitant to do uppercase / lowercase stuff ourselves, since there are cases where usernames are actually case sensitive (such as [Mediawiki](https://phabricator.wikimedia.org/T64396)). We also can't use underscores since kubernetes object names need to be valid DNS labels, so no '_'s or '.'s. This would also break current PVCs where users with -s are already mapped to the escaped names.

IMO, we just have to live with the ugliness here, there isn't really a way out of that.

consideRatio:
I've tried to make the naming nicer myself once, but it is tough compromises that also adds quite a bit of complexity to the code base. I'm closing this PR like I've shut down my own attempts before on this specific detail in the past.

Thank you @johanfleury for working to improve this open source project! :heart:

PR #333, opened 2019-07-15
Set upper limit on kubernetes client version

### LINKED ISSUE #332: kubernetes version limit
The setup.py file specifies kubernetes client version `'kubernetes>=7.0'` but there is a backwards-compatibility breaking change in the kubernetes client for version 10.0.0a1 as noted in https://github.com/kubernetes-client/python/issues/808.

Specifically, the imports in https://github.com/jupyterhub/kubespawner/blob/3bec1ba6802de97095835fd393ed7fc53bf2cca2/kubespawner/objects.py#L449 will now throw an error.

jmabry:
Close #332 .  This specifies that a compatible version of the kubernetes client is installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     install_requires=[
         'jupyterhub>=0.8',
         'pyYAML',
-        'kubernetes>=7.0',
+        'kubernetes>=7.0,<=9.0',
         'escapism',
         'jinja2',
         'async_generator>=1.8',


minrk:
Thanks for the PR! We generally try to avoid pinning down dependencies in Python. That's normally done at the application level, see #294. Additionally, the issue appears to have been a mistake in the 10.0 alpha, and fixed in 10.0 stable, so there's no action to take.

PR #334, opened 2019-07-16
Apply security context to container instead of pod where it is possible

shoelsch:
# PR Summary (update by @consideRatio)

KubeSpawner can spawn pods with a securityContext field set. But, this exists both for the pod itself, and the notebook server container we put into the pod. Some of these fields of the securityContext could be scoped more narrow to the container only, while others couldn't. This PR scopes what could be scoped narrowly to the container only, to the container only.

Pod scoped:
- `fsGroup` as configured by `kubespawner.fs_gid`
- `supplementalGroups` as configured by `kubespawner.supplemental_gids`
Container scoped:
- `runAsUser` as configured by `kubespawner.run_as_uid`
- `runAsGroup` as configured by `kubespawner.run_as_gid`
- `privileged` as configured by `kubespawner.run_privileged`

For some additional details about security contexts for pods and containers, see:
- about / how to: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
- container reference: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#securitycontext-v1-core
- pod reference: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#podsecuritycontext-v1-core

# Problem

I'm trying to deploy JupyterHub to a Kubernetes cluster with Istio enabled. When any pod is launched, an Istio sidecar container (`istio-init`) will be injected and requires root permissions.

I first ran into this issue when launching the hub pod and pulled in this fix (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1240), which moves the security context from the pod scope to the container scope.

I'm now running into this issue after authenticating with the hub and initiating a notebook pod launch via KubeSpawner. My understanding is that when I use default `jovyan` (userid=1000), this security context is being applied at the pod level, thereby instructing all containers to run as this user. As a result, my `istio-init` container fails to run because it requires root permissions.

# Implementation Compass

I'm opening this pull request early to get feedback before investing too much time. I plan to deprecate the use of the `V1PodSecurityContext` in favor of specifying the security context at the container level via the `V1SecurityContext` object. Does that seem reasonable?

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -8,7 +8,7 @@ from urllib.parse import urlparse
 from kubespawner.utils import get_k8s_model, update_k8s_model
 
 from kubernetes.client.models import (
-    V1Pod, V1PodSpec, V1PodSecurityContext,
+    V1Pod, V1PodSpec,
     V1ObjectMeta,
     V1LocalObjectReference,
     V1Volume, V1VolumeMount,
@@ -230,17 +230,6 @@ def make_pod(
     pod.spec = V1PodSpec(containers=[])
     pod.spec.restart_policy = 'OnFailure'
 
-    security_context = V1PodSecurityContext()
-    if fs_gid is not None:
-        security_context.fs_group = int(fs_gid)
-    if supplemental_gids is not None and supplemental_gids:
-        security_context.supplemental_groups = [int(gid) for gid in supplemental_gids]
-    if run_as_uid is not None:
-        security_context.run_as_user = int(run_as_uid)
-    if run_as_gid is not None:
-        security_context.run_as_group = int(run_as_gid)
-    pod.spec.security_context = security_context
-
     if image_pull_secret is not None:
         pod.spec.image_pull_secrets = []
         image_secret = V1LocalObjectReference()
@@ -252,6 +241,19 @@ def make_pod(
 
     if lifecycle_hooks:
         lifecycle_hooks = get_k8s_model(V1Lifecycle, lifecycle_hooks)
+
+    security_context = V1SecurityContext()
+    if fs_gid is not None:
+        security_context.fs_group = int(fs_gid)
+    if supplemental_gids is not None and supplemental_gids:
+        security_context.supplemental_groups = [int(gid) for gid in supplemental_gids]
+    if run_as_uid is not None:
+        security_context.run_as_user = int(run_as_uid)
+    if run_as_gid is not None:
+        security_context.run_as_group = int(run_as_gid)
+    if run_privileged:
+        security_context.privileged = True
+
     notebook_container = V1Container(
         name='notebook',
         image=image,
@@ -263,6 +265,7 @@ def make_pod(
         lifecycle=lifecycle_hooks,
         resources=V1ResourceRequirements(),
         volume_mounts=[get_k8s_model(V1VolumeMount, obj) for obj in (volume_mounts or [])],
+        security_context=security_context,
     )
 
     if service_account is None:
@@ -272,8 +275,6 @@ def make_pod(
     else:
         pod.spec.service_account_name = service_account
 
-    if run_privileged:
-        notebook_container.security_context = V1SecurityContext(privileged=True)
 
     notebook_container.resources.requests = {}
     if cpu_guarantee:
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -259,7 +259,9 @@ def make_pod(
        pod_security_context.fs_group = int(fs_gid)
     if supplemental_gids is not None and supplemental_gids:
        pod_security_context.supplemental_groups = [int(gid) for gid in supplemental_gids]
-    pod.spec.security_context = pod_security_context
+    # Only clutter pod spec with actual content
+    if not all([e is None for e in pod_security_context.to_dict().values()]):
+        pod.spec.security_context = pod_security_context
 
     container_security_context = V1SecurityContext()
     if run_as_uid is not None:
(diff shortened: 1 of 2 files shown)


consideRatio:
@shoelsch thanks! I think this makes a lot of sense in general.

@kevin-bates i recall you working with `gid` etc. Are you using containers other than the main notebook container that would be influenced by only setting such fields on one of the containers?

I think it would make sense to only configure the notebook container, as compared to the whole pod and all potential other containers with configuration. But, it does have potential to break something for someone. I'm thinking it makes sense though, and that we should version bump kubespawner and write an entry about this in a changelog - that these settings now became specific to the notebook-server container and potentially added sidecar containers would no longer be influenced by these settings any more.

@minrk perhaps you can weigh in here?

minrk:
I believe applying this to the notebook pod makes the most sense. init containers, etc. can include their own context(s) if needed.

kubespawner/objects.py
```diff
@@ -252,6 +241,19 @@ def make_pod(
...
+        security_context.fs_group = int(fs_gid)
+    if supplemental_gids is not None and supplemental_gids:
+        security_context.supplemental_groups = [int(gid) for gid in supplemental_gids]
+    if run_as_uid is not None:
+        security_context.run_as_user = int(run_as_uid)
+    if run_as_gid is not None:
+        security_context.run_as_group = int(run_as_gid)
+    if run_privileged:
+        security_context.privileged = True
+
```
consideRatio: This will make output become cleaner and tests be less cluttered, without it, a lot of `"securityContext": {}` will show up.

```suggestion
    if [e is None for e in security_context.to_dict().values()]:
        security_context = None
```

consideRatio: After this, updating tests to not assume generation of securityContext within the pod specs. But, when either of the following is set, there should be a non-blank entry in the notebook containers securityContext:
fs_gid, supplemental_gids, run_as_uid, run_as_gid, run_privileged

kevin-bates:
Thanks for the ping @consideRatio - much appreciated.  Enterprise Gateway includes the yaml for the kernel pods as part of the kernelspec resources.  As a result, installations could move/add the security context from the Pod spec to Container as they wish.  Of course, this template doesn't apply to spark-based kernels since Spark creates the pod(s) and I'm not immediately finding ways to influence the security context location from those settings, so there's potential for issues there - but this shouldn't affect your decision.

kevin-bates:
Thanks for the ping @consideRatio - much appreciated.  Enterprise Gateway includes the yaml for the kernel pods as part of the kernelspec resources.  As a result, installations could move/add the security context from the Pod spec to Container as they wish.  Of course, this template doesn't apply to spark-based kernels since Spark creates the pod(s) and I'm not immediately finding ways to influence the security context location from those settings, so there's potential for issues there - but this shouldn't affect your decision.

kevin-bates:
Thanks for the ping @consideRatio - much appreciated.  Enterprise Gateway includes the yaml for the kernel pods as part of the kernelspec resources.  As a result, installations could move/add the security context from the Pod spec to Container as they wish.  Of course, this template doesn't apply to spark-based kernels since Spark creates the pod(s) and I'm not immediately finding ways to influence the security context location from those settings, so there's potential for issues there - but this shouldn't affect your decision.

kevin-bates:
Thanks for the ping @consideRatio - much appreciated.  Enterprise Gateway includes the yaml for the kernel pods as part of the kernelspec resources.  As a result, installations could move/add the security context from the Pod spec to Container as they wish.  Of course, this template doesn't apply to spark-based kernels since Spark creates the pod(s) and I'm not immediately finding ways to influence the security context location from those settings, so there's potential for issues there - but this shouldn't affect your decision.

kevin-bates:
Thanks for the ping @consideRatio - much appreciated.  Enterprise Gateway includes the yaml for the kernel pods as part of the kernelspec resources.  As a result, installations could move/add the security context from the Pod spec to Container as they wish.  Of course, this template doesn't apply to spark-based kernels since Spark creates the pod(s) and I'm not immediately finding ways to influence the security context location from those settings, so there's potential for issues there - but this shouldn't affect your decision.

kubespawner/objects.py
shoelsch: Thanks for the suggestion on cleaning the output :+1: .

I've resolved all test failures.

There's still a remaining question around the parameters `fs_gid` (argument: `fs_group`) and `supplemental_gids` (argument: `supplemental_groups`), as those are Pod-only arguments ([V1 Pod Security Context](https://github.com/kubernetes-client/python/blob/v10.0.0/kubernetes/client/models/v1_pod_security_context.py).

We could resurrect the `V1PodSecurityContext` and ensure backward compatibility. Alternatively, deprecate those parameters in favor of _only_ providing users container-level control of the security context.

kevin-bates:
I suspect the [help text](https://github.com/jupyterhub/kubespawner/pull/334/files#diff-c71580208cd066d647338719829269e0L98-L101) added in #297 may not apply any longer, but don't know for sure.  This should probably be confirmed if possible.

shoelsch:
> I suspect the [help text](https://github.com/jupyterhub/kubespawner/pull/334/files#diff-c71580208cd066d647338719829269e0L98-L101) added in #297 may not apply any longer, but don't know for sure. This should probably be confirmed if possible.

Are we specifically talking about the feature gate portion? I wasn't able to find reference to how to enable/disable that feature gate in the documentation. Mind pointing me in the right direction?

kevin-bates:
Yes - just the feature-gate portion.  Here's a reference in the docs: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

I think ensuring the effective GID is not `root` is sufficient (unless of course run-as-group = 0 😄).

consideRatio:
We have quite a few additional tests setup, close/reopening this PR as I believe that refreshes its attempt to use latest .travis.yaml configuration.

kubespawner/objects.py
consideRatio: I rebased this PR on current master and added back those as it was easy to not break backward compatibility by doing that.

consideRatio:
Thanks a lot @shoelsch for your work on this! :heart: :tada:

Issue #335, opened 2019-07-17
It is valid to have storage_class be an empty string.


GrahamDumpleton:
In `make_pvc` the code uses:

```
    if storage_class:
        pvc.metadata.annotations.update({"volume.beta.kubernetes.io/storage-class": storage_class})
        pvc.spec.storage_class_name = storage_class
```

This is incorrect as it is valid to be able to specify a storage class name which is an empty string. Quoting the docs:

> PVCs don’t necessarily have to request a class. A PVC with its storageClassName set equal to "" is always interpreted to be requesting a PV with no class, so it can only be bound to PVs with no class (no annotation or one set equal to ""). A PVC with no storageClassName is not quite the same and is treated differently by the cluster depending on whether the DefaultStorageClass admission plugin is turned on.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

The code should thus read:

```
    if storage_class is not None:
        pvc.metadata.annotations.update({"volume.beta.kubernetes.io/storage-class": storage_class})
        pvc.spec.storage_class_name = storage_class
```

where default value for `storage_class` is `None`.

This is notification of issue. PR will likely be coming.

minrk:
closed by #337

PR #337, opened 2019-07-17
Permit storage class to be empty string.

GrahamDumpleton:
Adds support for using empty storage class per https://github.com/jupyterhub/kubespawner/issues/335

Test has been added, with modification to existing test to use `None` for storage class in most simple pvc test, but can't run test suite myself.

Still need to validate change in actual deployment, creating as draft PR for now so can be reviewed.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -674,8 +674,7 @@ class KubeSpawner(Spawner):
         config=True,
         allow_none=True,
         help="""
-        The storage class that the pvc will use. If left blank, the kubespawner will not
-        create a pvc for the pod.
+        The storage class that the pvc will use.
 
         This will be added to the `annotations: volume.beta.kubernetes.io/storage-class:`
         in the pvc metadata.
@@ -684,6 +683,9 @@ class KubeSpawner(Spawner):
         that matches the criteria of the StorageClass, the pvc will mount to that. Otherwise,
         b/c it has a storage class, k8s will dynamically spawn a pv for the pvc to bind to
         and a machine in the cluster for the pv to bind to.
+
+        Note that an empty string is a valid value and is always interpreted to be
+        requesting a pv with no class.
 
         See `the Kubernetes documentation <https://kubernetes.io/docs/concepts/storage/storage-classes/>`__
         for more information on how StorageClasses work.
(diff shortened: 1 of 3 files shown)


consideRatio:
Some background to why this PR can matter to others: an empty string as a storageClass value will ensure a default storage class isn't used, and thereby disabling dynamic provisioning of a disk!

Being able to set the storage class to empty string as compared to None makes kubespawner more in line with Kubernetes specs etc I think. It is unexpected to set it to an empty string but ending up with an unset value, as that still allows for dynamic provisioning. Yes! This PR is very relevant I think!

This looks good to me @GrahamDumpleton ! :tada: :heart:

---

## Off topic note
To disable automatic provisioning in the JupyterHub Helm chart, this can be configured:

```yaml
singleuser:
  storage:
    type: none
```

GrahamDumpleton:
Thanks. I managed to miss the reply from someone who had separately said they tested this for me, so hadn't moved it out of draft.

kubespawner/spawner.py
```diff
@@ -674,8 +674,7 @@ def _deprecated_changed(self, change):
         config=True,
         allow_none=True,
         help="""
-        The storage class that the pvc will use. If left blank, the kubespawner will not
-        create a pvc for the pod.
+        The storage class that the pvc will use.
```
consideRatio:
```suggestion
        The storage class that the PVC will use. If the default value of `None`
        is used, Kubespawner will not create a PVC for the pod.
```

consideRatio:
@minrk this LGTM overall. Agree to merge?

kubespawner/spawner.py
GrahamDumpleton: I couldn't see in the code where setting storage class to `None` would prevent a pvc being created. It is valid to have `None`, meaning don't define a storage class. Kubernetes still work in that case and will use whatever the default storage class is. The comment about `None` only seemed to apply to storage size and why I removed it. It seems it was copies from other comment when storage class was added and not removed when should.

consideRatio: Oh! Yes you are right, calling get_pvc_manifest will always return a manifest no matter of this, and that is a public facing function not invoked conditionally based on this.

It is probably a confusion of kubespawners use in the z2jh helm chart where such logic is implemented. Let's not commit this suggestion and resolve this conversation!

Issue #336, opened 2019-07-17
Add support for storage_selector through configuration.


GrahamDumpleton:
There should be a way of configuring a selector for a persistent volume claim.

```
$ kubectl explain pvc.spec.selector
KIND:     PersistentVolumeClaim
VERSION:  v1

RESOURCE: selector <Object>

DESCRIPTION:
     A label query over volumes to consider for binding.

     A label selector is a label query over a set of resources. The result of
     matchLabels and matchExpressions are ANDed. An empty label selector matches
     all objects. A null label selector matches no objects.

FIELDS:
   matchExpressions	<[]Object>
     matchExpressions is a list of label selector requirements. The requirements
     are ANDed.

   matchLabels	<map[string]string>
     matchLabels is a map of {key,value} pairs. A single {key,value} in the
     matchLabels map is equivalent to an element of matchExpressions, whose key
     field is "key", the operator is "In", and the values array contains only
     "value". The requirements are ANDed.
```

This will allow for a pool of persistent volumes setup in a certain way, including with existing content loaded in the volume, to be pre-created and for the persistent volume claim to match them via the storage label selector.

consideRatio:
A note about the implementation, it should be made as closely mapped to the k8s naming as possible. So, if it is a specification of matchLabels, it should reflect that.

PR #338, opened 2019-07-17
Add support for matching persistent volume using storage selector.

GrahamDumpleton:
Add support for matching using storage selector per https://github.com/jupyterhub/kubespawner/issues/336

Test has been added, but can't run test suite myself.

Note that because of overlapping changes in tests in https://github.com/jupyterhub/kubespawner/pull/337 one of these two PRs will likely need to be updated after the first of the two is merged.

Still need to validate change in actual deployment, creating as draft PR for now so can be reviewed.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -394,6 +394,7 @@ def make_pvc(
     name,
     storage_class,
     access_modes,
+    selector,
     storage,
     labels=None,
     annotations=None,
@@ -410,6 +411,8 @@ def make_pvc(
         String of the name of the k8s Storage Class to use.
     access_modes:
         A list of specifying what access mode the pod should have towards the pvc
+    selector:
+        Dictionary Selector to match pvc to pv.
     storage:
         The ammount of storage needed for the pvc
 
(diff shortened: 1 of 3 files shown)


GrahamDumpleton:
This has been separately tested as doing what was needed. I will wait until https://github.com/jupyterhub/kubespawner/pull/337 is merged and then I will fix up tests for this and move out of draft.

consideRatio:
This makes sense to me!

Two questions bubble to my mind:
1. Naming conventions: sticking exactly with k8s specs, or sticking with kubespawner non-systematic conventions, where we add `storage_` sometimes to the configuration name for PVC matters.
2. Setting a plain dictionary vs using it to be parsed to a V1LabelSelector object. It is my understanding, but I'm not at all 100%, that this would allow us to fail quicker or get better error messages etc.
    This contains details about the k8s API: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#persistentvolumeclaimspec-v1-core
    This contains details about the V1LabelSelector object as part of the kubernetes-client python library that we use: https://github.com/kubernetes-client/python/blob/master/kubernetes/client/models/v1_label_selector.py#L19

I'm not decided what I think about this matters fully. Hope to return again post #336 merge!

GrahamDumpleton:
Have updated the tests after merge of https://github.com/jupyterhub/kubespawner/pull/337

consideRatio:
Thanks @GrahamDumpleton !! :heart: :tada: !

Issue #354, opened 2019-09-25
kubespawner stopped working after kubernetes upgrade from 1.15 to 1.16


apoliakevitch:
```
[E 2019-09-25 05:16:53.998 JupyterHub web:1788] Uncaught exception GET /hub/user/apoliakevitch/ (10.244.4.1)
    HTTPServerRequest(protocol='http', host='conda.corp-apps.com', method='GET', uri='/hub/user/apoliakevitch/', version='HTTP/1.1', remote_ip='10.244.4.1')
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/tornado/web.py", line 1699, in _execute
        result = await result
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py", line 1013, in get
        raise copy.copy(exc).with_traceback(exc.__traceback__)
      File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 589, in error_callback
        future.result()
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py", line 636, in finish_user_spawn
        await spawn_future
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 489, in spawn
        raise e
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 409, in spawn
        url = await gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
      File "/usr/local/lib/python3.6/dist-packages/kubespawner/spawner.py", line 1636, in _start
        events = self.events
      File "/usr/local/lib/python3.6/dist-packages/kubespawner/spawner.py", line 1491, in events
        for event in self.event_reflector.events:
      File "/usr/local/lib/python3.6/dist-packages/kubespawner/spawner.py", line 72, in events
        key=lambda x: x.last_timestamp,
    TypeError: '<' not supported between instances of 'datetime.datetime' and 'NoneType'
```

dbricare:
This problem may be a result of the kubernetes python api only supporting up to version 1.14 [https://github.com/kubernetes-client/python](https://github.com/kubernetes-client/python)

AnotherCodeArtist:
So, closing this issue means closing Kubespawner and therefore Jupyterhub on K8s?
Will there never be a solution?

apoliakevitch:
[...]

PR #356, opened 2019-10-14
More idiomatic python syntax

AnotherCodeArtist:
Fixing the K8s 1.16 issue in a more idiomatic way

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -70,7 +70,7 @@ class EventReflector(NamespacedResourceReflector):
     def events(self):
         return sorted(
             self.resources.values(),
-            key=lambda x: x.last_timestamp,
+            key=lambda x: x.last_timestamp if x.last_timestamp is not None else 0.,
         )
 
 


manics:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1444#issuecomment-541707683
There's a small error in this, it should be either
`x.last_timestamp if x.last_timestamp is not None else datetime.min`
or
`x.last_timestamp.timestamp() if x.last_timestamp is not None else 0`

dhirschfeld:
It's quite a common idiom is to jsut use `or` to coalesce a null:
```
 key=lambda x: x.last_timestamp or datetime.min
```

GrahamDumpleton:
FWIW, may relate to older version of `kubernetes` module being used (I can't necessarily use latest at the moment, need to verify), but seeing the error:

```
  File "/opt/app-root/lib/python3.6/site-packages/kubespawner/spawner.py", line 73, in events
    key=lambda x: x.last_timestamp if x.last_timestamp is not None else 0.,
TypeError: '<' not supported between instances of 'float' and 'datetime.datetime'
```

in relation to this code.

It actually predates this change though as old code would also blow up as well.

```
  File "/opt/app-root/lib/python3.6/site-packages/kubespawner/spawner.py", line 72, in events
    key=lambda x: x.last_timestamp,
TypeError: '<' not supported between instances of 'NoneType' and 'datetime.datetime'
```

Still trying to understand why.

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T23:37:22Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:27:17Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

GrahamDumpleton:
Okay. Relates to https://github.com/jupyterhub/kubespawner/issues/354

Issue #347, opened 2019-09-05
KubeIngressProxy, using make_ingress -> can't import V1beta1Ingress


retypuj:
Hi,

I want JupyterHub to create a service and ingress automatically for every new user/pod. So I wrote in my `jupyterhub_config.py`:
```
from kubespawner import KubeIngressProxy
c.JupyterHub.proxy_class = KubeIngressProxy
C.KubeIngressProxy.should_start = False  # because we don't have a start method
```

But now I can't start JupyterHub:
`ImportError: cannot import name 'V1beta1Ingress'`

What I've tried:
- pip install kubernetes
- renaming V1beta1Ingress to V1Ingress 

How can I fix this?

kuromt:
Are you using kubespawner 0.10.1?

I install kubespawnerv0.10.1 and  it uses kubernetes==10.0.1, but python client's Ingress API is changed from 10.0.0.a.
https://github.com/kubernetes-client/python/blob/release-10.0/CHANGELOG.md#v1000a1

How about install kubernetes==9.0.1?

retypuj:
Thanks, I actually solved the issue by just renaming the kubernetes.client.models import in objects.py of kubespawner. I'm using kubernetes latest.
```
from kubernetes.client.models import (
    V1Pod, V1PodSpec, V1PodSecurityContext,
    V1ObjectMeta,
    V1LocalObjectReference,
    V1Volume, V1VolumeMount,
    V1Container, V1ContainerPort, V1SecurityContext, V1EnvVar, V1ResourceRequirements, V1Lifecycle,
    V1PersistentVolumeClaim, V1PersistentVolumeClaimSpec,
    V1Endpoints, V1EndpointSubset, V1EndpointAddress, V1EndpointPort,
    V1Service, V1ServiceSpec, V1ServicePort,
    V1Toleration,
    V1Affinity,
    V1NodeAffinity, V1NodeSelector, V1NodeSelectorTerm, V1PreferredSchedulingTerm, V1NodeSelectorRequirement,
    V1PodAffinity, V1PodAntiAffinity, V1WeightedPodAffinityTerm, V1PodAffinityTerm,
    ExtensionsV1beta1Ingress as V1beta1Ingress, 
    ExtensionsV1beta1IngressSpec as V1beta1IngressSpec, 
    ExtensionsV1beta1IngressRule as V1beta1IngressRule,
[...]

PR #360, opened 2019-10-17
Smoke py-kubernetes 9.* and 10.* in CI

clkao:


clkao:
Well this doesn't seem to show the error in #347 for kubernetes=10.*, perhaps we are not covering that code path of importing?

minrk:
I thought #347 only affected kube-py 10 *alpha*, and not the 10.0 final release. Is that not right?

consideRatio:
I'm closing this in favor of #362. I learned that since python Kubernetes 11's beta release we fail tests. I set it up to allow a failure on that specific test.

PR #362, opened 2019-11-16
Fix for Kubernetes 1.16 regarding datetime comparison

consideRatio:
Tested to work on z2jh with k8s 1.13-1.16!

Thanks @stv0g for demonstrating the fix in https://github.com/stv0g/kubespawner/commit/795167b98e548347bf75fd8b18def1e7cea0c901

Closes #357
Closes #354

## About CI changes
I ran into CI failures that was caused by a new release of kubernetes-client/python 11.0.0b1, I also lacked testing of version 9 and 10 of this client, so I fixed them for my current DX.

## Review
- Do we really want to sort the events at all now that we don't always have timestamp information? @bitnik can you give your input on this?

## Open questions
What kind of events that doesn't come with this timestamp value, and why? Here are examples: https://github.com/jupyterhub/kubespawner/issues/354#issuecomment-538203621

PR #365, opened 2019-11-16
Add relevant badges to README.md

consideRatio:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
-# kubespawner (jupyterhub-kubespawner) #
+# [kubespawner](https://github.com/jupyterhub/kubespawner) (jupyterhub-kubespawner @ PyPI)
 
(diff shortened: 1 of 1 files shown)


betatim:
Was going to merge this but after restarting the CI jobs I looked at their logs and am confused why they fail :(

consideRatio:
@betatim it relates to the release of 11.0.0b1 of kubernetes-client/python that we rely on. It is also fixed in the commit of #362 where I allow kubernetes==11 to fail. The issue to track this failure is tracked in #366

betatim:
So should we be setting the dependencies in https://github.com/jupyterhub/kubespawner/blob/18847a9cfcccb6e1644d8f3f24c3675a7c5e65b7/setup.py#L17 so that we limit ourselves to https://pypi.org/project/kubernetes/10.0.1/ instead of allowing the beta version?

manics:
Looks like it's a deliberate decision in Travis:
https://github.com/jupyterhub/kubespawner/blob/18847a9cfcccb6e1644d8f3f24c3675a7c5e65b7/.travis.yml#L13-L15

consideRatio:
@manics aha! that explains my confusion about this behavior where 11.0.0b1 was installed!

PR #367, opened 2019-11-17
[WIP] K8s 1.16 fix by not sorting events

consideRatio:
Testing an alternative to #362 by simply removing the sorting.

A technical review point on this PR is regarding passing by reference or by value. I wonder if we return a reference to a dictionary that can be modified now that we don't sort it. That doesn't feel great. Do I need to make a copy here?


betatim:
https://github.com/jupyterhub/kubespawner/pull/362#issuecomment-554905189 for a comment on why I think we need to sort.

consideRatio:
@betatim yepp, and the test failures in this branch confirms that it is a bigger undertaking to not sort. On rerunning some of the job that failed, i noted kinds of failures, so it became very non-deterministic.

I'm closing this.

PR #368, opened 2019-11-18
[MRG] CI reworked and bugs affecting tests fixed

### LINKED ISSUE #358: Utilize pyflakes in CI
We realized the benefit of having something like pyflakes to catch basic python errors in https://github.com/jupyterhub/kubespawner/pull/357.

This issue could be closed by updating the CI system to run a pyflake on our python code, and fail on basic errors.

### LINKED ISSUE #373: Decide on Py35 support
In #368 I learned about working with the `progress()` function was problematic if you have py35 as it use `yeild` in an `async` function, but not really, because it use `_yeild` from the `async_generator` library. But, a lot of errors popped up in tests when using py35 specifically.

I'm not eager to fix that compatibility issue, what should we do? Document it? Require py36 when installing kubespawner? Letting kubespawner crash hard or provide a warning if using events with py35?

Here is some stuff to read about py35 and async for loops:
- https://quentin.pradet.me/blog/using-asynchronous-for-loops-in-python.html

consideRatio:
## Final summary
- flake8 used for static code testing and that closes #358
- tests now run in 18.04 (bionic) from 16.04 (xenial)
- k8s version 1.12-1.16 tested instead of 1.10 only
- Kubespawner now require `kubernetes>=8`, which is the kubernetes-client/python library, not kubernetes itself.
- We now test pre-releases of the kubernetes client version but allow for failures in it
- We now test nightly python
- We now test a kubernetes client version with the latest not officially supported kubernetes cluster version.
- I corrected a bad solution in #362 where k8s events without `last_timestamp` was set to the 0 timestamp. We now instead fallback to use `event_time` which is a high resolution timestamp that sometimes replaces `last_timestamp` in modern k8s versions.
- I fixed bugs causing test instabilities and potential dropping of events reported `spawner.py`'s `progress()` function. The solution was to ensure to never return without being sure all events that has arrived during the spawning process got a chance to be parsed before returning.
- I conclude that `async for` loops are unstable in python 3.5 even together with the `async_generator` library, and that we should warn about this or require kubespawner to use Py36 or higher in another PR.
- Closes #373 by adding a requirements of Python 3.6+
- Adds inline docs that has helped me grasp important pieces of the code base.
- Test and mute that a DeprecationWarning is warned about from deprecated code.

## PS
I've spent about 30-40 hours on this PR. I've struggled with a mix of non-deterministic errors that doesn't always reproduce, intertwined with py35 specific errors, intertwined with k8s version specific errors. It feels sooooooo good to have the test reliably go green now.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+select = F
+ignore = E,W,C,F401,F841
+exclude = __init__.py
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,12 @@
+# Artifacts from running kubespawner tests with minikube
+bin/
+lib64
+minikube-linux-amd64
+node_modules/
+package-lock.json
+pyvenv.cfg
+share/
+
 # JupyterHub running from kubespawner folder for development purposes
 jupyterhub-proxy.pid
 jupyterhub.sqlite
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,14 +26,26 @@ after_success:
   - codecov
 
 matrix:
-  fast_finish: true
   allow_failures:
     - python: nightly
+  fast_finish: true
   include:
-    - python: nightly
-    - python: 3.5
+    # Test the oldest supported python on
+    # the latest supported kubernetes-client,
+    # and use the latest supported kubernetes
+    # version along for the kubernetes-client.
+    - python: 3.8
       env: KUBE_PY=7
-    - python: 3.6
-      env: KUBE_PY=7
+      # env: KUBE_PY=7 KUBE_VERSION=1.11.*
     - python: 3.7
       env: KUBE_PY=8
+      # env: KUBE_PY=8 KUBE_VERSION=1.12.*
+    - python: 3.6
+      env: KUBE_PY=9
+      # env: KUBE_PY=9 KUBE_VERSION=1.13.*
+    - python: 3.5
+      env: KUBE_PY=10
+      # env: KUBE_PY=10 KUBE_VERSION=1.14.*
+    - python: nightly
+      env: KUBE_PY=11
+      # env: KUBE_PY=11 KUBE_VERSION=1.15.*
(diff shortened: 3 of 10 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/binder-jupyterhub-activity-round-up-week-1/2704/1

consideRatio:
# AHA!
The instability of spawner.progress() return-value relates to sleeping for 1 second and then checking if one should continue, but sometimes on k8s 1.14+ this goes so fast that an event of "started container" has been sent without getting parsed because it arrived in this time window of 1 second.

https://github.com/jupyterhub/kubespawner/blob/908d123892644d87ee5f7171b19d67073fcbbbb4/kubespawner/spawner.py#L1541-L1571

This is fixable!

consideRatio:
# Uhu...

Well, it was a decent idea, but it wasn't what was needed to go all the way. Now the issue that remains I believe is our adding of events with a timestamp that is set before the past event, this becomes very weird for the progress() function that assumes with its use of a `next_event` index that the event list is only appended, while now it is can be prepended. This prepending comes from adding an even twithout a timestamp that gets the 0 timestamp and hence is sorted to be the first event - a prepending.

I got to solve that as well to get tests to stabilize.

consideRatio:
# Aha finally!

1. last_timestamp isn't always set, neither is first_timestamp. But, when they are not set, the more high resolution version may be set, it is named `event_time`.
2. python 3.5 has issues that I fail to solve. I note that we use asyncio_generator with spawner.py's progress() function, and that may be a way to extend the ability of using yeild within async functions to py35, but it's not robust enough it seems. I don't know for sure what to do about this. I'd personally like to bump the requirements to py36 to make it simple.
3. minikube has issues setting up for kubernetes 1.11, and I have soo horrible experiences already fixing CI stuff regarding that so I'm dropping testing of k8s 1.11.

PR #379, opened 2019-11-29
Don't require deploy job to also test

consideRatio:


consideRatio:
Ping @manics as this relates to PyPI CD setup

manics:
:+1: Mixing build matrices and build stages and defaults is complicated.... probably worth thinking about in https://github.com/jupyterhub/team-compass/issues/213

consideRatio:
So I did something, and now it failed in another way that I believe may relate to travis being upset that it didn't run any test at all etc.

Arrrrgh I don't get travis matrix stuff and how the custom "deploy" fields work. They add magic to something magical and it is hard to read what it translates to..

https://travis-ci.org/mapbox/vtvalidate/jobs/492280599

PR #381, opened 2019-11-29
Fix progress serialization issue

consideRatio:
Fixes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1501

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1628,7 +1628,7 @@ class KubeSpawner(Spawner):
                     progress += (90 - progress) / 3
                     await yield_({
                         'progress': int(progress),
-                        'raw_event': event,
+                        'raw_event': event.to_dict(),  # V1Event isn't serializable
                         'message':  "%s [%s] %s" % (
                             event.last_timestamp or event.event_time,
                             event.type,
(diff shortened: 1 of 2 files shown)


betatim:
In some other projects I use something like the following to serialise datetime objects (this is whats failing/why travis is unhappy):

```python
import json

class JSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, datetime.datetime):
            return obj.isoformat()
        return super().default(obj)
json._default_encoder = JSONEncoder()
```

but I am not sure this is "the right way" of doing this.

manics:
What uses the raw-event? Would a combination of `event.to_dict()` and converting the datetime field to an iso8601 string work?

consideRatio:
@clkao added it with a motivation that makes sense to me generally in https://github.com/jupyterhub/kubespawner/pull/361.

manics:
```python
v = kubernetes.client.models.V1Event(involved_object={}, metadata={}, event_time=datetime.datetime.now())
d = v.to_dict()
for (k, v) in d.items():
    if isinstance(v, datetime.datetime):
        d[k] = v.isoformat()
```

consideRatio:
manics you have the series field of the event object that van have datetimes in it as well (MicroTime ins the k8s api ref). So it must go deeper if we do this.

PR #382, opened 2019-11-29
Delete remnant now unused parts in spawner.py

bitnik:


kubespawner/spawner.py
```diff
@@ -1613,11 +1611,6 @@ def events(self):
 
             len_events = len(events)
             if next_event < len_events:
-                # only show messages for the 'current' pod
-                # pod_id may change if a previous pod is being stopped
-                # before starting a new one
-                # use the uid of the latest event to identify 'current'
-                pod_id = events[-1].involved_object.uid
```
consideRatio: Hmmmmm...
- Every KubeSpawner get their own pod.
- my_spawner.events returns a list filtered, but this list is filtered to the pod, but perhaps the name of the pod rather than the UID of the pod?
- Two pods with the same name kmim the same namespace cannot happen.
- When does events disappear? I suspect it is after their involved object has been deleted, perhaps delayed or directly.
- Is this code legacy code, not working, meant to avoid getting events from past pods? Is this perhaps something to fix?

I know i dont want to remove this without make a remark about it in the code allowing someone to correct this later if needed with more information available to get started. So perhaps reference this comment?

yuvipanda:
IMO, we should delete the code if it isn't being used - we can always revert it from version control - rather than comment it out. Cleaner

consideRatio:
@bitnik I'm sorry for grinding this PR to a halt. We should indeed remove outdated parts. I force rebased commit with proper deletions after looking into this sufficiently enough to feel confident that I wouldn't loose important insights about the logic by doing so.

- 19c6e6a4 introduced the `pod_id` variable and logic, part of #186.
- 2bb1a943 let `pod_id` variable remain, but removed the logic, part of #219

PR #383, opened 2019-12-06
Expose __version__ in kubespawner module

consideRatio:
I ended up wanting this to be available to the _log_cls helper function
in JupyterHub.

ref: https://github.com/jupyterhub/jupyterhub/blob/abb93ad799865a4b27f677e126ab917241e1af72/jupyterhub/app.py#L2246-L2264

### DIFF
--- a/kubespawner/__init__.py
+++ b/kubespawner/__init__.py
@@ -12,6 +12,7 @@ in your `jupyterhub_config.py` file.
 # Users can simply import kubespawner.KubeSpawner in their applications
 # instead of the more verbose import kubespawner.spawner.KubeSpawner.
 
-from kubespawner.spawner import KubeSpawner, __version__
+from kubespawner.spawner import KubeSpawner
 
-__all__ = [KubeSpawner, __version__]
+__version__ = '0.11.2.dev'
+__all__ = [KubeSpawner]
--- a/kubespawner/__init__.py
+++ b/kubespawner/__init__.py
@@ -12,6 +12,6 @@ in your `jupyterhub_config.py` file.
 # Users can simply import kubespawner.KubeSpawner in their applications
 # instead of the more verbose import kubespawner.spawner.KubeSpawner.
 
-from kubespawner.spawner import KubeSpawner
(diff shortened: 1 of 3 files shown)


kubespawner/__init__.py
```diff
@@ -12,6 +12,6 @@
 # Users can simply import kubespawner.KubeSpawner in their applications
 # instead of the more verbose import kubespawner.spawner.KubeSpawner.
 
-from kubespawner.spawner import KubeSpawner
+from kubespawner.spawner import KubeSpawner, __version__
 
-__all__ = [KubeSpawner]
+__all__ = [KubeSpawner, __version__]
```
betatim: By adding something to `__all__` we mark it as something that should be imported when you do `from kubespawner import *`. I'm not a fan of adding `__version__` because it is considered an anti-pattern to `from kubespawner import *` and I can't think of when you'd want to then have a `__version__` appear in your namespace. Especially in cases where people `from blah import *` multiple times it will lead to a lot of confusion. So having to write
```python
import kubespawner
print("k8s spawner version:", kubespawner.__version__)
```
seems like a good thing to make people do.

kubespawner/__init__.py
```diff
@@ -12,6 +12,6 @@
 # Users can simply import kubespawner.KubeSpawner in their applications
 # instead of the more verbose import kubespawner.spawner.KubeSpawner.
 
-from kubespawner.spawner import KubeSpawner
+from kubespawner.spawner import KubeSpawner, __version__
```
betatim: nit unrelated to this PR: do you know why we use `from kubespawner.spawner import KubeSpawner` here? Normally inside a package I'd expect to see relative imports so `from .spawner import KubeSpawner`.

kubespawner/__init__.py
consideRatio: Aha! Thanks for helping me understand this, I didn't know!

kubespawner/__init__.py
consideRatio: No clue, I lack the python experience to say anything about it.

PR #385, opened 2019-12-23
Multi-namespace support, redux

athornton:
WIP; not tested yet.  Just wanted to get something on the table prior to the end of the year.

In this model users will need to request a MultiNamespacedKubeSpawner in order to get multi-namespace support.  That should leave the majority of people who don't need this functionality unaffected.

### DIFF
--- a/kubespawner/__init__.py
+++ b/kubespawner/__init__.py
@@ -12,7 +12,10 @@ in your `jupyterhub_config.py` file.
 # Users can simply import kubespawner.KubeSpawner in their applications
 # instead of the more verbose import kubespawner.spawner.KubeSpawner.
 
-from kubespawner.spawner import KubeSpawner
+# Same for MultiNamespacedKubeSpawner
 
-__version__ = '0.11.2.dev'
-__all__ = [KubeSpawner]
+from kubespawner.spawner import KubeSpawner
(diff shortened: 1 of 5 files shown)


lresende:
Where/How is the user namespace created in this scenario?

athornton:
In my current setup, it's created in the LSST-specific logic I have in a helper class (I do it at pod start time).  

I would suggest--if we want to make the K8s object manipulation something the spawner class itself is responsible for--that we add a couple of configurables (like a user-namespace template, and a delete-namespace-on-stop boolean), and do namespace (and ancillary object, which should be another configurable) creation in start(), and teardown (if requested) in stop().

Ancillary object manipulation would be creating, for instance, service accounts, roles, and rolebindings within the user namespace allowing the user to spawn additional containers for things like dask workers or Argo workflows (all of which has implications for the RBAC for the service account the Hub itself runs with).

athornton:
Just added first cut at the namespace manipulation methods.

Hopefully I will be able to start doing some testing early next week.

athornton:
Ooof.  I got fairly far, but was having trouble with the pod and event reflector startup.  Then another massive upstream change seems to have landed....gonna take a bit to rework.

athornton:
Oh dear.  I screwed up the repository something fierce.  Closing this one, followup in #387

Issue #223, opened 2018-08-01
Pod Evictions (memory, node shutodowns) - requires manual hub restart!


consideRatio:
## About
We have an issue when pods are removed unexpectedly that seems to require a hub restart. This could happen if a node fails, is preempted, or an admin deletes a user pod I think. This is really bad I'd say as it requires the specific user that has this problem to contact and administrator and then the administrator has to restart the hub to solve it.

I deem this vital to fix as preemtible nodes (google) and spot nodes (amazon) are an awesome way to reduce cost, but usage of them risks causing this kind of huge trouble right now!

I'm still lacking some insight in whats going on in the proxy and the user management within JupyterHub though so I'm hoping someone can pick this issue up from my writeup. @minrk or @yuvipanda perhaps you know someone to /cc or could give me a pointer on where to look to solve this?

## Experience summary
1. Pod dies
2. "503 - Service Unavailable"
3. `hub/admin` stop pod
4. "Your server is stopping"
5. Distracting consequence
6. Issue still not resolved
7. Hub restart
8. Everything is fine

#### 1. A preemptible node was reclaimed and my user pod was lost

- The user pod name was `erik-2esundell`
- The user name was `erik.sundell`

---

#### 2. Later - I visit the hub

![image](https://user-images.githubusercontent.com/3837114/43514641-e62f8208-9580-11e8-94f5-cf506be05880.png)

---

#### 3. Directly after - I visit `hub/admin` and press the stop server button
- The button turns blue and is labeled `start server` again

![image](https://user-images.githubusercontent.com/3837114/43514285-28f8bfc4-9580-11e8-97a6-9cd9597a008b.png)

- The hub logs sais
```
[I 2018-08-01 09:36:05.670 JupyterHub proxy:264] Removing user erik.sundell from proxy (/user/erik.sundell/)                                                                                                                 
[I 2018-08-01 09:36:05.675 JupyterHub spawner:1644] Deleting pod jupyter-erik-2esundell                                                                                                                                      
[W 2018-08-01 09:36:05.688 JupyterHub spawner:1657] No pod jupyter-erik-2esundell to delete. Assuming already deleted.                                                                                                       
[W 2018-08-01 09:36:15.671 JupyterHub base:751] User erik.sundell: server is slow to stop
[I 2018-08-01 09:36:15.672 JupyterHub log:158] 202 DELETE /hub/api/users/erik.sundell/server (erik.sundell@10.20.0.1) 10013.08ms
```

[...]

PR #386, opened 2019-12-24
support internal_ssl for kubespawner

sstarcher:
Here is my attempt at adding ssl support to Kubespawner.  

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1520

### DIFF
--- a/docs/source/ssl.md
+++ b/docs/source/ssl.md
@@ -1,0 +1,14 @@
+# Internal SSL
+
+JupyterHub 1.0 introduces internal_ssl configuration for encryption and authentication of all internal communication.
+
+Kubespawner can mount the internal_ssl certificates as Kubernetes secrets into the jupyter user's pod.
+
+## Setup
+
+```
+c.JupyterHub.internal_ssl = True
+
+c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
+c.KubeSpawner.internal_ssl_directory = '/srv/jupyterhub/internal-ssl/'
+```
--- a/docs/source/ssl.md
+++ b/docs/source/ssl.md
@@ -10,5 +10,4 @@ Kubespawner can mount the internal_ssl certificates as Kubernetes secrets into t
 c.JupyterHub.internal_ssl = True
 
 c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
-c.KubeSpawner.internal_ssl_directory = '/srv/jupyterhub/internal-ssl/'
 ```
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -1,26 +1,32 @@
 """
 Helper methods for generating k8s API objects.
 """
+import base64
 import json
+import os
 import re
 from urllib.parse import urlparse
 
+from kubernetes.client.models import (V1Affinity, V1Container, V1ContainerPort,
+                                      V1EndpointAddress, V1EndpointPort,
+                                      V1Endpoints, V1EndpointSubset, V1EnvVar,
+                                      V1LabelSelector, V1Lifecycle,
+                                      V1LocalObjectReference, V1NodeAffinity,
+                                      V1NodeSelector,
+                                      V1NodeSelectorRequirement,
+                                      V1NodeSelectorTerm, V1ObjectMeta,
+                                      V1PersistentVolumeClaim,
+                                      V1PersistentVolumeClaimSpec, V1Pod,
+                                      V1PodAffinity, V1PodAffinityTerm,
+                                      V1PodAntiAffinity, V1PodSecurityContext,
+                                      V1PodSpec, V1PreferredSchedulingTerm,
+                                      V1ResourceRequirements, V1Secret,
+                                      V1SecurityContext, V1Service,
+                                      V1ServicePort, V1ServiceSpec,
+                                      V1Toleration, V1Volume, V1VolumeMount,
+                                      V1WeightedPodAffinityTerm)
 from kubespawner.utils import get_k8s_model, update_k8s_model
 
-from kubernetes.client.models import (
-    V1Pod, V1PodSpec, V1PodSecurityContext,
-    V1ObjectMeta,
-    V1LocalObjectReference,
-    V1Volume, V1VolumeMount,
-    V1Container, V1ContainerPort, V1SecurityContext, V1EnvVar, V1ResourceRequirements, V1Lifecycle,
-    V1PersistentVolumeClaim, V1PersistentVolumeClaimSpec,
-    V1Endpoints, V1EndpointSubset, V1EndpointAddress, V1EndpointPort,
-    V1Service, V1ServiceSpec, V1ServicePort,
-    V1Toleration,
-    V1Affinity,
-    V1NodeAffinity, V1NodeSelector, V1NodeSelectorTerm, V1PreferredSchedulingTerm, V1NodeSelectorRequirement,
-    V1PodAffinity, V1PodAntiAffinity, V1WeightedPodAffinityTerm, V1PodAffinityTerm,
-)
 
 def make_pod(
     name,
@@ -62,6 +68,8 @@ def make_pod(
     pod_anti_affinity_preferred=None,
     pod_anti_affinity_required=None,
     priority_class_name=None,
+    ssl_secret_name=None,
+    username=None,
     logger=None,
 ):
     """
@@ -215,6 +223,8 @@ def make_pod(
         * https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#podaffinityterm-v1-core
     priority_class_name:
         The name of the PriorityClass to be assigned the pod. This feature is Beta available in K8s 1.11.
+    ssl_secret_name:
+        Specifies the name of the ssl secret
     """
 
     pod = V1Pod()
@@ -235,6 +245,26 @@ def make_pod(
         image_secret = V1LocalObjectReference()
         image_secret.name = image_pull_secret
         pod.spec.image_pull_secrets.append(image_secret)
+
+    if ssl_secret_name:
+        if not volumes:
+            volumes = []
+        volumes.append({
+            'name':'jupyterhub-internal-certs',
+            'secret': {
+                'secretName': ssl_secret_name,
+                'defaultMode': 511
+            }
+        })
+
+        ssl_path = '/etc/jupyterhub/ssl/'
+        env['JUPYTERHUB_SSL_KEYFILE'] = ssl_path + "ssl.key"
+        env['JUPYTERHUB_SSL_CERTFILE'] = ssl_path + "ssl.crt"
+        env['JUPYTERHUB_SSL_CLIENT_CA'] = ssl_path + "notebooks-ca_trust.crt"
+
+        if not volume_mounts:
+            volume_mounts = []
+        volume_mounts.append({'name': 'jupyterhub-internal-certs', 'mountPath': ssl_path})
 
     if node_selector:
         pod.spec.node_selector = node_selector
@@ -578,3 +608,107 @@ def make_ingress(
     )
 
     return endpoint, service, ingress
+
+
+def make_secret(
+    name,
+    username,
+    ssl_directory,
+    labels=None,
+    annotations=None,
+):
+    """
+    Make a k8s secret specification using pre-existing ssl credentials for a given user.
+
+    Parameters
+    ----------
+    name:
+        Name of the secret. Must be unique within the namespace the object is
+        going to be created in.
+    username:
+        The name of the user notebook.
+    ssl_directory:
+        Path to a directory containing all users certificates and keys.
+    labels:
+        Labels to add to the secret.
+    annotations:
+        Annotations to add to the secret.
+    """
+
+    secret = V1Secret()
+    secret.kind = "Secret"
+    secret.api_version = "v1"
+    secret.metadata = V1ObjectMeta()
+    secret.metadata.name = name
+    secret.metadata.annotations = (annotations or {}).copy()
+    secret.metadata.labels = (labels or {}).copy()
+
+    secret.data = {}
+    user_ssl_key = "{}/user-{}/user-{}.key".format(ssl_directory, username, username)
+    with open(user_ssl_key, 'r') as file:
+        encoded = base64.b64encode(file.read().encode("utf-8"))
+        secret.data['ssl.key'] = encoded.decode("utf-8")
+
+    user_ssl_crt = "{}/user-{}/user-{}.crt".format(ssl_directory, username, username)
+    with open(user_ssl_crt, 'r') as file:
+        encoded = base64.b64encode(file.read().encode("utf-8"))
+        secret.data['ssl.crt'] = encoded.decode("utf-8")
+
+    notebook_ca = "{}/notebooks-ca/notebooks-ca.crt".format(ssl_directory)
+    with open(notebook_ca, 'r') as file:
+        encoded = base64.b64encode(file.read().encode("utf-8"))
+        secret.data["notebooks-ca_trust.crt"] = encoded.decode("utf-8")
+
+    hub_ca_trust = "{}/hub-ca_trust.crt".format(ssl_directory)
+    with open(hub_ca_trust, 'r') as file:
+        encoded = base64.b64encode(file.read().encode("utf-8"))
+        secret.data["notebooks-ca_trust.crt"] = secret.data["notebooks-ca_trust.crt"] + encoded.decode("utf-8")
+
+    return secret
+
+
+def make_service(
+    name,
+    port,
+    servername,
+    labels=None,
+    annotations=None,
+):
+    """
+    Make a k8s service specification for using dns to communicate with the notebook.
+
+    Parameters
+    ----------
+    name:
+        Name of the service. Must be unique within the namespace the object is
+        going to be created in.
+    env:
+        Dictionary of environment variables.
+    labels:
+        Labels to add to the service.
+    annotations:
+        Annotations to add to the service.
+
+    """
+
+    metadata = V1ObjectMeta(
+        name=name,
+        annotations=(annotations or {}).copy(),
+        labels=(labels or {}).copy()
+    )
+
+    service = V1Service(
+        kind='Service',
+        metadata=metadata,
+        spec=V1ServiceSpec(
+            type='ClusterIP',
+            ports=[V1ServicePort(port=port, target_port=port)],
+            selector={
+                'component': 'singleuser-server',
+                'hub.jupyter.org/servername': servername,
+                'hub.jupyter.org/username': metadata.labels['hub.jupyter.org/username']
+            }
+        )
+    )
+
+    return service
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -15,6 +15,7 @@ from kubernetes.client.models import (V1Affinity, V1Container, V1ContainerPort,
                                       V1NodeSelector,
                                       V1NodeSelectorRequirement,
                                       V1NodeSelectorTerm, V1ObjectMeta,
+                                      V1OwnerReference,
                                       V1PersistentVolumeClaim,
                                       V1PersistentVolumeClaimSpec, V1Pod,
                                       V1PodAffinity, V1PodAffinityTerm,
@@ -69,7 +70,7 @@ def make_pod(
     pod_anti_affinity_required=None,
     priority_class_name=None,
     ssl_secret_name=None,
-    username=None,
+    ssl_secret_mount_path=None,
     logger=None,
 ):
     """
@@ -225,6 +226,8 @@ def make_pod(
         The name of the PriorityClass to be assigned the pod. This feature is Beta available in K8s 1.11.
     ssl_secret_name:
         Specifies the name of the ssl secret
+    ssl_secret_mount_path:
+        Specifies the name of the ssl secret mount path for the pod
     """
 
     pod = V1Pod()
@@ -257,14 +260,13 @@ def make_pod(
             }
         })
 
-        ssl_path = '/etc/jupyterhub/ssl/'
-        env['JUPYTERHUB_SSL_KEYFILE'] = ssl_path + "ssl.key"
-        env['JUPYTERHUB_SSL_CERTFILE'] = ssl_path + "ssl.crt"
-        env['JUPYTERHUB_SSL_CLIENT_CA'] = ssl_path + "notebooks-ca_trust.crt"
+        env['JUPYTERHUB_SSL_KEYFILE'] = ssl_secret_mount_path + "ssl.key"
+        env['JUPYTERHUB_SSL_CERTFILE'] = ssl_secret_mount_path + "ssl.crt"
+        env['JUPYTERHUB_SSL_CLIENT_CA'] = ssl_secret_mount_path + "notebooks-ca_trust.crt"
 
         if not volume_mounts:
             volume_mounts = []
-        volume_mounts.append({'name': 'jupyterhub-internal-certs', 'mountPath': ssl_path})
+        volume_mounts.append({'name': 'jupyterhub-internal-certs', 'mountPath': ssl_secret_mount_path})
 
     if node_selector:
         pod.spec.node_selector = node_selector
@@ -609,11 +611,25 @@ def make_ingress(
 
     return endpoint, service, ingress
 
+def make_owner_reference(name, uid):
+    """
+    Returns a owner reference object for garbage collection.
+    """
+    return V1OwnerReference(
+            api_version="v1",
+            kind="Pod",
+            name=name,
+            uid=uid,
+            block_owner_deletion=True,
+            controller=False
+           )
 
 def make_secret(
     name,
     username,
-    ssl_directory,
+    cert_paths,
+    hub_ca,
+    owner,
     labels=None,
     annotations=None,
 ):
@@ -627,8 +643,10 @@ def make_secret(
         going to be created in.
     username:
         The name of the user notebook.
-    ssl_directory:
+    cert_paths:
         Path to a directory containing all users certificates and keys.
+    hub_ca:
+        Path to the hub certificate authority
     labels:
         Labels to add to the secret.
     annotations:
@@ -642,25 +660,23 @@ def make_secret(
     secret.metadata.name = name
     secret.metadata.annotations = (annotations or {}).copy()
     secret.metadata.labels = (labels or {}).copy()
+    secret.metadata.owner_references=[owner]
 
     secret.data = {}
-    user_ssl_key = "{}/user-{}/user-{}.key".format(ssl_directory, username, username)
-    with open(user_ssl_key, 'r') as file:
+
+    with open(cert_paths['keyfile'], 'r') as file:
         encoded = base64.b64encode(file.read().encode("utf-8"))
         secret.data['ssl.key'] = encoded.decode("utf-8")
 
-    user_ssl_crt = "{}/user-{}/user-{}.crt".format(ssl_directory, username, username)
-    with open(user_ssl_crt, 'r') as file:
+    with open(cert_paths['certfile'], 'r') as file:
         encoded = base64.b64encode(file.read().encode("utf-8"))
         secret.data['ssl.crt'] = encoded.decode("utf-8")
 
-    notebook_ca = "{}/notebooks-ca/notebooks-ca.crt".format(ssl_directory)
-    with open(notebook_ca, 'r') as file:
+    with open(cert_paths['cafile'], 'r') as file:
         encoded = base64.b64encode(file.read().encode("utf-8"))
         secret.data["notebooks-ca_trust.crt"] = encoded.decode("utf-8")
 
-    hub_ca_trust = "{}/hub-ca_trust.crt".format(ssl_directory)
-    with open(hub_ca_trust, 'r') as file:
+    with open(hub_ca, 'r') as file:
         encoded = base64.b64encode(file.read().encode("utf-8"))
         secret.data["notebooks-ca_trust.crt"] = secret.data["notebooks-ca_trust.crt"] + encoded.decode("utf-8")
 
@@ -671,6 +687,7 @@ def make_service(
     name,
     port,
     servername,
+    owner,
     labels=None,
     annotations=None,
 ):
@@ -694,7 +711,8 @@ def make_service(
     metadata = V1ObjectMeta(
         name=name,
         annotations=(annotations or {}).copy(),
-        labels=(labels or {}).copy()
+        labels=(labels or {}).copy(),
+        owner_references=[owner],
     )
 
     service = V1Service(
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -249,7 +249,7 @@ def make_pod(
         image_secret.name = image_pull_secret
         pod.spec.image_pull_secrets.append(image_secret)
 
-    if ssl_secret_name:
+    if ssl_secret_name and ssl_secret_mount_path:
         if not volumes:
             volumes = []
         volumes.append({
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -629,7 +629,7 @@ def make_secret(
     username,
     cert_paths,
     hub_ca,
-    owner,
+    owner_references,
     labels=None,
     annotations=None,
 ):
@@ -644,7 +644,7 @@ def make_secret(
     username:
         The name of the user notebook.
     cert_paths:
-        Path to a directory containing all users certificates and keys.
+        JupyterHub spawners cert_paths dictionary container certificate path references
     hub_ca:
         Path to the hub certificate authority
     labels:
@@ -660,7 +660,7 @@ def make_secret(
     secret.metadata.name = name
     secret.metadata.annotations = (annotations or {}).copy()
     secret.metadata.labels = (labels or {}).copy()
-    secret.metadata.owner_references=[owner]
+    secret.metadata.owner_references=owner_references
 
     secret.data = {}
 
@@ -687,7 +687,7 @@ def make_service(
     name,
     port,
     servername,
-    owner,
+    owner_references,
     labels=None,
     annotations=None,
 ):
@@ -712,7 +712,7 @@ def make_service(
         name=name,
         annotations=(annotations or {}).copy(),
         labels=(labels or {}).copy(),
-        owner_references=[owner],
+        owner_references=owner_references,
     )
 
     service = V1Service(
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -5,45 +5,36 @@ This module exports `KubeSpawner` class, which is the actual spawner
 implementation that should be used by JupyterHub.
 """
 
-from functools import partial  # noqa
-from datetime import datetime
 import json
-import os
-import sys
-import string
 import multiprocessing
-from concurrent.futures import ThreadPoolExecutor
+import os
+import string
+import sys
 import warnings
+from asyncio import sleep
+from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime
+from functools import partial  # noqa
 
-from tornado import gen
-from tornado.ioloop import IOLoop
+from jinja2 import BaseLoader, Environment
+from tornado import gen, web
 from tornado.concurrent import run_on_executor
-from tornado import web
-from traitlets import (
-    Bool,
-    Dict,
-    Integer,
-    List,
-    Unicode,
-    Union,
-    default,
-    observe,
-    validate,
-)
-from jupyterhub.spawner import Spawner
-from jupyterhub.utils import exponential_backoff
-from jupyterhub.traitlets import Command
-from kubernetes.client.rest import ApiException
-from kubernetes import client
+from tornado.ioloop import IOLoop
+
 import escapism
-from jinja2 import Environment, BaseLoader
+from async_generator import async_generator, yield_
+from jupyterhub.spawner import Spawner
+from jupyterhub.traitlets import Command
+from jupyterhub.utils import exponential_backoff
+from kubernetes import client
+from kubernetes.client.rest import ApiException
+from kubespawner.objects import make_pod, make_pvc, make_secret, make_service
+from kubespawner.reflector import NamespacedResourceReflector
+from kubespawner.traitlets import Callable
+from traitlets import (Bool, Dict, Integer, List, Unicode, Union, default,
+                       observe, validate)
 
 from .clients import shared_client
-from kubespawner.traitlets import Callable
-from kubespawner.objects import make_pod, make_pvc
-from kubespawner.reflector import NamespacedResourceReflector
-from asyncio import sleep
-from async_generator import async_generator, yield_
 
 
 class PodReflector(NamespacedResourceReflector):
@@ -130,6 +121,10 @@ class KubeSpawner(Spawner):
         "events": None,
     }
 
+    # Characters as defined by safe for DNS
+    # Note: '-' is not in safe_chars, as it is being used as escape character
+    safe_chars = set(string.ascii_lowercase + string.digits)
+
     @property
     def pod_reflector(self):
         """
@@ -186,6 +181,14 @@ class KubeSpawner(Spawner):
 
         # runs during both test and normal execution
         self.pod_name = self._expand_user_properties(self.pod_name_template)
+        self.dns_name = self.dns_name_template.format(namespace=self.namespace, service=self.pod_name)
+
+        self.secret_name = None
+        if self.internal_ssl_directory:
+            self.secret_name = self._expand_user_properties(self.secret_name_template)
+            self.ssl_alt_names.append("DNS:"+self.dns_name)
+            self.ssl_alt_names_include_local=False
+
         self.pvc_name = self._expand_user_properties(self.pvc_name_template)
         if self.working_dir:
             self.working_dir = self._expand_user_properties(self.working_dir)
@@ -247,7 +250,6 @@ class KubeSpawner(Spawner):
         config=True,
         help="""
         The IP address (or hostname) the single-user server should listen on.
-
         We override this from the parent so we can set a more sane default for
         the Kubernetes setup.
         """
@@ -306,6 +308,14 @@ class KubeSpawner(Spawner):
         """
     )
 
+    dns_name_template = Unicode(
+        '{service}.{namespace}.svc.cluster.local',
+        config=True,
+        help="""
+        Template to use to form the dns name for the pod.
+        """
+    )
+
     pod_name_template = Unicode(
         'jupyter-{username}{servername}',
         config=True,
@@ -342,6 +352,30 @@ class KubeSpawner(Spawner):
         This must be unique within the namespace the pvc are being spawned
         in, so if you are running multiple jupyterhubs spawning in the
         same namespace, consider setting this to be something more unique.
+        """
+    )
+
+    secret_name_template = Unicode(
+        'jupyter-{username}{servername}',
+        config=True,
+        help="""
+        Template to use to form the name of user's secret.
+
+        `{username}` is expanded to the escaped, dns-label safe username.
+
+        This must be unique within the namespace the pvc are being spawned
+        in, so if you are running multiple jupyterhubs spawning in the
+        same namespace, consider setting this to be something more unique.
+        """
+    )
+
+    internal_ssl_directory = Unicode(
+        None,
+        allow_none=True,
+        config=True,
+        help="""
+        Location of JupyterHub's internal_ssl directory.  This directory is managed by JupyterHub itself.  
+        See https://jupyterhub.readthedocs.io/en/stable/api/app.html
         """
     )
 
@@ -1174,7 +1208,7 @@ class KubeSpawner(Spawner):
         config=True,
         help="""
         Time in seconds for the pod to be in `terminating` state before is forcefully killed.
-        
+
         Increase this if you need more time to execute a `preStop` lifecycle hook.
 
         See https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods for
@@ -1302,20 +1336,16 @@ class KubeSpawner(Spawner):
     del _deprecated_name
 
     def _expand_user_properties(self, template):
-        # Make sure username and servername match the restrictions for DNS labels
-        # Note: '-' is not in safe_chars, as it is being used as escape character
-        safe_chars = set(string.ascii_lowercase + string.digits)
-
         # Set servername based on whether named-server initialised
         if self.name:
             servername = '-{}'.format(self.name)
-            safe_servername = '-{}'.format(escapism.escape(self.name, safe=safe_chars, escape_char='-').lower())
+            safe_servername = '-{}'.format(escapism.escape(self.name, safe=self.safe_chars, escape_char='-').lower())
         else:
             servername = ''
             safe_servername = ''
 
-        legacy_escaped_username = ''.join([s if s in safe_chars else '-' for s in self.user.name.lower()])
-        safe_username = escapism.escape(self.user.name, safe=safe_chars, escape_char='-').lower()
+        legacy_escaped_username = ''.join([s if s in self.safe_chars else '-' for s in self.user.name.lower()])
+        safe_username = escapism.escape(self.user.name, safe=self.safe_chars, escape_char='-').lower()
         return template.format(
             userid=self.user.id,
             username=safe_username,
@@ -1338,7 +1368,9 @@ class KubeSpawner(Spawner):
     def _build_common_labels(self, extra_labels):
         # Default set of labels, picked up from
         # https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md
-        labels = {}
+        labels = {
+            'hub.jupyter.org/username': escapism.escape(self.user.name, safe=self.safe_chars, escape_char='-').lower()
+        }
         labels.update(extra_labels)
         labels.update(self.common_labels)
         return labels
@@ -1346,7 +1378,8 @@ class KubeSpawner(Spawner):
     def _build_pod_labels(self, extra_labels):
         labels = self._build_common_labels(extra_labels)
         labels.update({
-            'component': 'singleuser-server'
+            'component': 'singleuser-server',
+            'hub.jupyter.org/servername': self.name,
         })
         return labels
 
@@ -1389,7 +1422,7 @@ class KubeSpawner(Spawner):
         if self.cmd:
             real_cmd = self.cmd + self.get_args()
         else:
-            real_cmd = None
+            real_cmd = self.get_args()
 
         labels = self._build_pod_labels(self._expand_all(self.extra_labels))
         annotations = self._build_common_annotations(self._expand_all(self.extra_annotations))
@@ -1434,7 +1467,42 @@ class KubeSpawner(Spawner):
             pod_anti_affinity_preferred=self.pod_anti_affinity_preferred,
             pod_anti_affinity_required=self.pod_anti_affinity_required,
             priority_class_name=self.priority_class_name,
+            ssl_secret_name=self.secret_name,
             logger=self.log,
+        )
+
+
+    def get_secret_manifest(self):
+        """
+        Make a secret manifest that contains the ssl certificates.
+        """
+
+        labels = self._build_common_labels(self._expand_all(self.extra_labels))
+        annotations = self._build_common_annotations(self._expand_all(self.extra_annotations))
+
+        return make_secret(
+            name=self.secret_name,
+            username=self.user.name,
+            ssl_directory=self.internal_ssl_directory,
+            labels=labels,
+            annotations=annotations,
+        )
+
+
+    def get_service_manifest(self):
+        """
+        Make a service manifest for dns.
+        """
+
+        labels = self._build_common_labels(self._expand_all(self.extra_labels))
+        annotations = self._build_common_annotations(self._expand_all(self.extra_annotations))
+
+        return make_service(
+            name=self.pod_name,
+            port=self.port,
+            servername=self.name,
+            labels=labels,
+            annotations=annotations,
         )
 
     def get_pvc_manifest(self):
@@ -1784,6 +1852,35 @@ class KubeSpawner(Spawner):
                 else:
                     raise
 
+        if self.internal_ssl_directory:
+            secret = self.get_secret_manifest()
+            if secret:
+                try:
+                    yield self.asynchronize(
+                        self.api.create_namespaced_secret,
+                        namespace=self.namespace,
+                        body=secret
+                    )
+                except ApiException as e:
+                    if e.status == 409:
+                        self.log.info("Secret " + self.secret_name + " already exists, so did not create new secret.")
+                    else:
+                        raise
+
+            service = self.get_service_manifest()
+            if service:
+                try:
+                    yield self.asynchronize(
+                        self.api.create_namespaced_service,
+                        namespace=self.namespace,
+                        body=service
+                    )
+                except ApiException as e:
+                    if e.status == 409:
+                        self.log.info("Service " + self.pod_name + " already exists, so did not create new service.")
+                    else:
+                        raise
+
         # If we run into a 409 Conflict error, it means a pod with the
         # same name already exists. We stop it, wait for it to stop, and
         # try again. We try 4 times, and if it still fails we give up.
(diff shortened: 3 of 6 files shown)


sstarcher:
Failure does not look related to my changes `The command "./ci/install-kube.sh" exited with 70.`

sstarcher:
So one current issue I'm running into is that tornado is doing SSL verification on certificates for the DNS/IP.  The certificates are created prior to the pod running so the IP is unknown.  Another option would be to configure it to use a Kubernetes service and use that DNS name in the certificates.  In this case, the hub/spawner would have to be configured to use the DNS name instead of the pod's IP address.

sstarcher:
Reconfigure around using DNS instead of the pod IP so the certificates can be properly signed

scivm:
Any thoughts on merging this pull request?

kubespawner/spawner.py
```diff
@@ -242,19 +237,8 @@ def _namespace_default(self):
                 return f.read().strip()
         return 'default'
 
-    ip = Unicode(
```
minrk: Why is this removed?

kubespawner/spawner.py
```diff
@@ -242,19 +237,8 @@ def _namespace_default(self):
...
-        The IP address (or hostname) the single-user server should listen on.
-
-        We override this from the parent so we can set a more sane default for
-        the Kubernetes setup.
-        """
-    )
-
     cmd = Command(
-        None,
+        ['jupyterhub-singleuser'],
```
minrk: Is there a reason this default config is changed? It doesn't seem relevant to this PR.

kubespawner/objects.py
```diff
@@ -566,3 +610,99 @@ def make_ingress(
...
+            name=name,
+            annotations=(annotations or {}).copy(),
+            labels=(labels or {}).copy()
+        ),
+        spec=V1ServiceSpec(
+            type='ClusterIP',
+            ports=[V1ServicePort(port=port, target_port=port)],
+            selector={
+                'component': 'singleuser-server',
+                'hub.jupyter.org/username': labels['hub.jupyter.org/username']
```
minrk: username + component is not unique, since there can be named servers. Maybe this needs `hub.jupyter.org/servername` as well? Or maybe this should just be the pod name instead, which is already unique?

kubespawner/spawner.py
```diff
@@ -306,6 +290,14 @@ def _namespace_default(self):
         """
     )
 
+    dns_name_template = Unicode(
```
minrk: 👍

kubespawner/spawner.py
sstarcher: It did not look to me that it was being used?  Am I mistaken?

kubespawner/spawner.py
sstarcher: As long as you are fine with the changes here I can reverse this change.  I need 1 of the two.
https://github.com/jupyterhub/kubespawner/pull/386/files/f75a5790eff424d6db55872463b26691dce1a453#diff-3082573191d96b79e2ec71986d4ef3e2R1386

kubespawner/objects.py
sstarcher: So can the user have 2 servers running at once?  If so than the service name is also non-unique currently and will also need to change?

sstarcher: From attempting to use Jupyter Hub I can't seem to launch 2 user pods at once.  If that's the case this single service looks like it "should" only have one component it points to.

sstarcher: I started my notebook using one spawner config and checked it's labels and had to stop it to start a second one and checked it's labels.

  labels:
    app: jupyterhub
    chart: jupyterhub-0.0.1-set.by.chartpress
    component: singleuser-server
    heritage: jupyterhub
    hub.jupyter.org/network-access-hub: "true"
    hub.jupyter.org/username: jupyter-shane-2estarcher-40xxx-2ecom
    release: jupyterhub

  labels:
    app: jupyterhub
    chart: jupyterhub-0.0.1-set.by.chartpress
    component: singleuser-server
    heritage: jupyterhub
    hub.jupyter.org/network-access-hub: "true"
    hub.jupyter.org/username: jupyter-shane-2estarcher-40xxx-2ecom
    release: jupyterhub

scivm:
hub:
  allowNamedServers: true

And use latest beta.2.  Click on control panel you will see.

sstarcher:
Thanks let me test that out

sstarcher:
So it looks like the pod label for username is not really the username.  This may be a bug in kubespawner/jupyter.

Without me changing the labels in anyway I'm seeing the following on my named server launch.

```
  labels:
    app: jupyterhub
    chart: jupyterhub-0.0.1-set.by.chartpress
    component: singleuser-server
    heritage: jupyterhub
    hub.jupyter.org/network-access-hub: "true"
    hub.jupyter.org/username: jupyter-shane-2estarcher-40xxx-2ecom-test
    release: jupyterhub
```

None of my code is touching the labels on the pod, but it looks like the `username` is really the pod name.  Thoughts on what action I should take?  Currently as written my code works with named services due to this bug.

sstarcher:
Nope, I see what's happening.  Let me fix this.

sstarcher:
Validated the changes now work with allowedNamedServers

kubespawner/spawner.py
```diff
@@ -1784,6 +1822,35 @@ def _start(self):
...
+                    namespace=self.namespace,
+                    body=secret
+                )
+            except ApiException as e:
+                if e.status == 409:
+                    self.log.info("Secret " + self.pod_name + " already exists, so did not create new secret.")
+                else:
+                    raise
+
+        service = self. get_service_manifest()
```
manics:
```suggestion
        service = self.get_service_manifest()
```
Looks like a typo crept in here

sstarcher:
@manics thanks good catch

consideRatio:
Thanks for this PR @sstarcher!

Could you write a summary about what the PR aims to accomplish (encrypted internal traffic between ... and ...), and what steps are taken to accomplish it, and perhaps why if you think it makes sense to clarify implementation choices further.

It would help me and others to review this i think!

sstarcher:
The goal of this PR is to allow the enabling of encrypted traffic between the proxy and the user notebook pod.

* Kubespawner will now create a Kubernetes secret for the user pod that contains the certificates created by JupyterHub
* Kubespawner will now create a Kubernetes service to use for communication vs the current IP of the pod.  This enables us to use a static DNS name for the certificates and for the pod.  In addition to enabling the ability to use SSL it also allows for the pod to be restarted.  Under the current system if the pod is restarted due to OOM or some other issue it will gain a new IP so the Proxy will lose the ability to communicate to the pod.

For any secure system, it is recommended to enable encryption between all services.

sstarcher:
Anything I can do to help move this forward?

kubespawner/objects.py
```diff
@@ -236,6 +245,41 @@ def make_pod(
...
+                ]
+            }
+        })
+
+        if not working_dir:
+            raise Exception('working_dir must be specified when ssl is enabled')
+
+        if not volume_mounts:
+            volume_mounts = []
+        mountPath = os.path.dirname(os.path.dirname(working_dir+'/'+env['JUPYTERHUB_SSL_KEYFILE']))
```
yuvipanda: Should also be `mount_path` to fit Python coding conventions

kubespawner/objects.py
```diff
@@ -566,3 +610,103 @@ def make_ingress(
...
+    secret = V1Secret()
+    secret.kind = "Secret"
+    secret.api_version = "v1"
+    secret.metadata = V1ObjectMeta()
+    secret.metadata.name = name
+    secret.metadata.annotations = (annotations or {}).copy()
+    secret.metadata.labels = (labels or {}).copy()
+
+    secret.data = {}
+    with open(env['JUPYTERHUB_SSL_KEYFILE'], 'r') as file:
```
yuvipanda: We should pass these all in manually, or by passing in an appropriate object. We shouldn't use env vars like this.

kubespawner/objects.py
```diff
@@ -566,3 +610,103 @@ def make_ingress(
...
+
+    folder = os.path.dirname(env['JUPYTERHUB_SSL_CLIENT_CA'])
+    with open(folder+'/hub-ca_trust.crt', 'r') as file:
+        encoded = base64.b64encode(file.read().encode("utf-8"))
+        secret.data["notebooks-ca_trust.crt"] = secret.data["notebooks-ca_trust.crt"] + encoded.decode("utf-8")
+
+    return secret
+
+
+def make_service(
```
yuvipanda: I'm a little torn about this. JupyterHub's internal_ssl supports using 'IP' as part of alt_names, but in Kubernetes we do not know the IP until after the pod is created - at which point we can not add a secret there...

There is already code here for creating a new Service per pod - https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/proxy.py#L122. I'd love for us to reuse that here instead - although it is possible we can not.

kubespawner/spawner.py
```diff
@@ -1437,6 +1442,39 @@ def get_pod_manifest(self):
             logger=self.log,
         )
 
+
+    def get_secret_manifest(self):
```
yuvipanda: We should clean up the Secret and Service object when we kill the pod - this keeps the namespace clean, and helps lock down internal cert secrets. We can let kubernetes do the work for us with `ownerReferences` - https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/. We should add these to the secret & service.

yuvipanda:
What do you think of projects like https://linkerd.io/? This is  much lighter than something like Istio, and will provide us with stuff like encryption, rate limiting, etc

sstarcher:
The issue with adding the certs as environment variables would be that at the kubernetes level someone with the ability to list pods would be able to access the certificates.  We might be able to do something with permissions and dropping the user down into less privileged access to not be able to access the secret files.

sstarcher:
Projects like linkerd and istio are nice, but the added complexity I can't imagine forcing a user to add.

yuvipanda:
I agree re: Linkerd.

Re: certs, if people can list pod contents, they can get the JUPYTERHUB_TOKEN, which basically lets them connect to the notebook and execute arbitrary code

scivm:
Does merging this prevent making linkerd an optional addition in the chart later on?

sstarcher:
@scivm it certainly does not.

@yuvipanda Just because the current design of putting JUPYTERHUB_TOKEN in the pod exists I don't think that means we should also continue that less than desirable practice.  If anything a future update should move the JUPYTERHUB_TOKEN into a secret.

I can certainly remove the secret and put it in the environment variable, but I believe that's a less secure option.

yuvipanda:
@sstarcher Fair enough! Let's ignore the environment values and focus on secret creation as way forward.

sstarcher:
Sounds good, I should be able to start making your suggested changes next week.  Thanks for the review.

kubespawner/objects.py
```diff
@@ -236,6 +245,41 @@ def make_pod(
         image_secret.name = image_pull_secret
         pod.spec.image_pull_secrets.append(image_secret)
 
+    if ssl_enabled(env):
+        key_name = 'user-'+env['JUPYTERHUB_USER']
```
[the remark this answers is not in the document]
sstarcher: I don't understand the first point.  I'm not inserting anything into the env.  I'm using the data that is being passed in by jupyter.  I also don't understand what this line has to do with the secret name?  I can certainly make it configurable, but what do we gain by that?  This creates the secret so what value do we get by making the secret name configurable?

If the goal is to move this logic from objects.py to spawner.py I can certainly do that, but the env does already contain this information.  Let me know what you think is best.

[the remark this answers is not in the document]
yuvipanda: Ah, so 'env' should be treated as a dumb dictionary of stuff to be passed through to the podspec - we shouldn't reach into it to grab stuff. This function should instead get all the info it needs as explicit parameters, rather than implicitly reach into env.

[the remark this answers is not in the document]
yuvipanda: We should make the secret object's name configurable to match the fact that [pvc](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L334) and [pod names](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L309) are also configurable.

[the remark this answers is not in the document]
sstarcher: I will do both of those things as requested.

I don't understand the purpose of not tying ourselves to what comes from Jupyter and instead of making the user duplicate configure the values, but I'll do it.

I also don't understand the value of allowing the secret name to be configurable.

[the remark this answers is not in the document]
yuvipanda: > I also don't understand the value of allowing the secret name to be configurable.

Quoting the doc string from https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L309:

> This must be unique within the namespace the pods are being spawned
> in, so if you are running multiple jupyterhubs spawning in the
> same namespace, consider setting this to be something more unique.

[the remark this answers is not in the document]
sstarcher: Which is why the secret was using the same format as the pod which is always unique.  Hence why I don't know why it needs to be configurable.

sstarcher:
@yuvipanda I believe I have resolved your concerns except for the cleanup of the secrets.

Let me know if I missed something.

samuelmanzer:
Just wanted to bump this to make sure it isn't forgotten as we also need this (accessing user servers as admin doesn't appear to work without it)  - it would solve this problem for z2jh users:

https://github.com/jupyterhub/jupyterhub/issues/2284

I followed the trail of sstarcher's work and it seems to terminate here:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1520
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1535

I'd be happy to help out to get this PR in.

sstarcher:
@samuelmanzer yep, this is the last pole in the tent.  I believe I have everything done except for cleaning up the secret and the service resource.  I'm hoping to get some time in the next week to finish it up.

If you don't mind taking a review pass to make sure I did not miss anything else that was requested to be changed.

samuelmanzer:
@sstarcher I think that's @yuvipanda's call

sstarcher:
@yuvipanda due to the order that the resources are created in I did not use the built-in garbage collection method as I need the POD UID before I create the secret, but the  POD needs the secret before I can create it.

Let me know if you want any further changes.

docs/source/ssl.md
```diff
@@ -0,0 +1,15 @@
...
+Kubespawner can mount the internal_ssl certificates as Kubernetes secrets into the jupyter user's pod.
+
+## Setup
+
+```
+c.JupyterHub.internal_ssl = True
+
+c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
+c.KubeSpawner.internal_ssl_directory = '/srv/jupyterhub/internal-ssl/'
+c.KubeSpawner.working_dir = '/home/jovyan/' # This is required to allow us to mount the certificates into the correct directory
```
consideRatio, line 14: I don't understand this section, the certificates are not mounted there anyhow right? They are mounted in /etc/jupyterhub/ssl as I understand it.

If this is needed, perhaps we could suggest the use of the fact that worker_dir will be expanded. I'm not sure if any of these would make sense though, because perhaps the username refers to the JH username while we want to work from $HOME wherever that is, which can be different from the JH username.

```
        return template.format(
            userid=self.user.id,
            username=safe_username,
            unescaped_username=self.user.name,
            legacy_escape_username=legacy_escaped_username,
            servername=safe_servername,
            unescaped_servername=servername
        )
```

kubespawner/spawner.py
```diff
@@ -1784,6 +1851,35 @@ def _start(self):
...
+                    yield self.asynchronize(
+                        self.api.create_namespaced_secret,
+                        namespace=self.namespace,
+                        body=secret
+                    )
+                except ApiException as e:
+                    if e.status == 409:
+                        self.log.info("Secret " + self.secret_name + " already exists, so did not create new secret.")
+                    else:
+                        raise
```
consideRatio, line 1867: When using a non highly-available k8s api-server, timeout errors will happen from time to time. Will this raise cause an error to be logged but jupyterhub keep running rather than jupyterhub crashing?

docs/source/ssl.md
sstarcher: You are correct `working_dir` is no longer needed.

kubespawner/spawner.py
```diff
@@ -345,6 +355,29 @@ def _namespace_default(self):
...
+    )
+
+    internal_ssl_directory = Unicode(
+        None,
+        allow_none=True,
+        config=True,
+        help="""
+        Location of Jupyter's internal ssl directory.
+        """
+    )
```
consideRatio, line 379: Could we provide a default value for this? Is there a reason not to?

kubespawner/spawner.py
sstarcher: This will cause an exception to raise and be logged, and the code will keep operating.  This is patterned after the code a few lines above.

kubespawner/spawner.py
sstarcher: This can be determined from the Jupyter Hub environment variable which is what I was previously doing, but @yuvipanda asked me to not do that.

If we don't want to use the jupyterhub environment variable I don't believe it makes sense for us to attempt a default.

consideRatio: Hmmm, I'm trying to put pieces together still. Could you help me out by making it more explicit what this traitlet configures?

- Is this a location on the JupyterHub pod or singleuser pod?
- What certificates (and keys?) will be contained here?

I'll open a new comment where I summarize my current understanding, perhaps that is better starting point.

consideRatio:
@sstarcher thank you so much for your work on this, I'm lack some understanding to properly review this, but perhaps you could help me a bit?

My current understanding is not yet fully developed and I'm aware of that, so let me try to make it explicit so you can catch misunderstandings i may have.

To communicate encrypted, my understanding is that:

1. We have the JupyterHub acting as a CA sayin "Trust me! I'm an authority!" and all user pods will get a certificate (but not key) to be able to encrypted messages to send towards JupyterHub, which in turn only can decrypt if it with its private key.
2. So, JupyterHub will have a private key and a public key (certificate) that other know to trust.
3. JupyterHub will create K8s secrets and put in private key / public key pairs in them.
4. Hmmm... do JupyterHub perhaps need to keep the public part around as well?

Hmmm hmmm...

I think what I struggle with when reviewing this at the moment, is that I don't get the full picture when looking at the code. Could you provide a summary of:
- how the encryption process is meant to work, similar to my attempt above
- where keys are generated by kubespawner (is this the folder that we configure with internal_ssl_folder?)
- how keys are distributed (by the kubespawner creation of k8s secrets and mounting of those somewhere - is this the hardcoded location /etc/jupyterhub/ssl?)

sstarcher:
@consideRatio sure, your 1-4 points are correct.  I doubt JupyterHub needs to keep the public part around, but that's JupyterHub's concern and not Kubespawners concern.

* how the encryption process is meant to work, similar to my attempt above
  * JupyterHub creates keys that are given to the pods.  The pods use those keys to do mutual auth with the API and to encrypt and decrypt traffic.

* where keys are generated by kubespawner (is this the folder that we configure with internal_ssl_folder?)
  * JupyterHub generates these SSL keys when `internal_ssl` is true.  JupyterHub configuration can override this location with `internal_certs_location` - https://jupyterhub.readthedocs.io/en/stable/api/app.html
  * JupyterHub passes an environment variable to Kubespawner that defines this location, but I was asked to make it explicit instead of tying it to the JupyterHub environment variable that is exposed.
* how keys are distributed (by the kubespawner creation of k8s secrets and mounting of those somewhere - is this the hardcoded location /etc/jupyterhub/ssl?)
  * Yes, JupyterHub creates the keys and KubeSpawner puts them into a secret
  * That secret is mounted into the pod at the specific hard coded location

Let me know if you have any other questions.

kubespawner/spawner.py
consideRatio: > Is this a location on the JupyterHub pod or singleuser pod?

It is the location on the jupyterhub pod.

> What certificates (and keys?) will be contained here?

All certificates, for users and for JH, i think.

kubespawner/objects.py
```diff
@@ -566,3 +596,107 @@ def make_ingress(
...
+
+def make_secret(
+    name,
+    username,
+    ssl_directory,
+    labels=None,
+    annotations=None,
+):
+    """
+    Make a k8s secret specification for mount ssl credentials for running a user notebook.
```
consideRatio, line 609: If my understanding is correct, this function does not generate key/crt files while doing this, but will instead assume they exist by now in the `ssl_directory` folder.

sstarcher: Correct it does not generate the key/crt.  It uses the existing key/crt created by JupyterHub to create a kubernetes secret.  If you have better wording for that doc I'm all for it.

consideRatio:
```suggestion
    Make a k8s secret specification using pre-existing ssl credentials for a given user.
```

kubespawner/objects.py
```diff
@@ -566,3 +596,107 @@ def make_ingress(
...
+
+    Parameters
+    ----------
+    name:
+        Name of the secret. Must be unique within the namespace the object is
+        going to be created in.
+    username:
+        The name of the user notebook.
+    ssl_directory:
+        Path to the SSL directory for certificates and keys.
```
consideRatio, line 619: I would like to add something about this directory's structure, either by declaring it to look like ...

        /user-{username}/user-{username}.key
        /user-{username}/user-{username}.crt
        /notebooks-ca/notebooks-ca.crt
        /hub-ca_trust.crt

... or by referencing it is created by JupyterHub automatically or something like that.

```suggestion
        Path to a directory containing all users certificates and keys.
```

sstarcher: I'm good either way what do you prefer?  I can link out to the JupyterHub docs.  I can inline your above example of the directory structure.

consideRatio: I don't yet understand this directory, is jupyterhub generating these certs? Referencing that JupyterHub is meant to generate it all, and the docs about that would be the best option in my mind.

sstarcher: JupyterHub can create it and create all of the certs or the user can also do it.

sstarcher: Let me know if this makes sense
```
        Location of JupyterHub's internal_ssl directory.  This directory is managed by JupyterHub itself.  
        See https://jupyterhub.readthedocs.io/en/stable/api/app.html
```

docs/source/ssl.md
```diff
@@ -0,0 +1,14 @@
...
+
+Kubespawner can mount the internal_ssl certificates as Kubernetes secrets into the jupyter user's pod.
+
+## Setup
+
+```
+c.JupyterHub.internal_ssl = True
+
+c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
+c.KubeSpawner.internal_ssl_directory = '/srv/jupyterhub/internal-ssl/'
```
consideRatio, line 13: There is `c.JupyterHub.internal_certs_location` that serves this purpose.

https://jupyterhub.readthedocs.io/en/stable/api/app.html#jupyterhub.app.JupyterHub.internal_certs_location

Ideas:
1. Trying to use c.JupyterHub.internal_certs_location instead of defining this ourselves on KubeSpawner. Perhaps we cannot reference this, but perhaps we can? I'm not sure. If we can't, I think it make sense that JupyterHub would try pass this to kubespawner.
2. If we fail, then we could perhaps try reuse this name?

I note that the default value of c.JupyterHub.internal_certs_location is "internal-ssl", and i think that assumes like you have a path to this location being decided by `c.JupyterHub.data_files_path` which defaults to a dynamic value, determined by this logic here: https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/_data.py

Yikes, perhaps we can request backup to figure out what make sense here.

kubespawner/objects.py
consideRatio: I can provide a suggestion if we get this right first.

https://github.com/jupyterhub/kubespawner/pull/386#pullrequestreview-386026000

docs/source/ssl.md
sstarcher: I was referencing it via an environment variable that is made available to kubespawner, but I was explicitly asked to not tie it to JupyterHub.  Jupyter makes available the following environment variables that I was previously used instead of the explicit internal_ssl_directory.

env['JUPYTERHUB_SSL_KEYFILE']
env['JUPYTERHUB_SSL_CERTFILE']
env['JUPYTERHUB_SSL_CLIENT_CA']

My understanding is these 3 variables are always correct and can be relied on if we are willing to tie to JupyterHub, but @yuvipanda specifically asked me to not use these.

consideRatio:
@sstarcher I'm heading to bed now. I'll continue to work with you to get this merged to the best of my ability.

Thank you so much for your work on this. This is a PR to a machinery that has lots of cogs that by themselves take quite a while to figure out, at least it did for me. The jupyter configuration system (traitlets) took a long time for me to understand well enough for example, and kubernetes :p

So - thank you! I'm excited about getting this merged and I think you have got almost all the way, what remains in my mind is for me to properly understand everything so I can agree it all make sense.

sstarcher:
@consideRatio I very much appreciate your reviews in helping make this better

consideRatio:
@sstarcher I've tried to provide some advice on addressing @yuvipanda's [ask](https://github.com/jupyterhub/kubespawner/pull/386#discussion_r376241763) to add a ownerReferences entry to the k8s secret and k8s service that you add.

## Secrets / Services should have their ownerReference field set

References: [garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/), [ownerReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#ownerreference-v1-meta), [uid's](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids)

We need to first create the pod, and then get it's uid in `metadata.uid` before we create the other resources to set the `ownerReference` to the pod. How do we do this?

1. We get the info from the pod [as a response of the creation request](https://github.com/kubernetes-client/python/blob/b871710b8bda29def1c28aad5c69087e7a6d0bb6/kubernetes/e2e_test/test_client.py#L64-L67).
2. We move the code to create secret/service after the creation of the pod.
   - It is fine for the pod to reference a secret volume that does not exists, the mounting of the secret will simply fail at a later stage until it exists if we do.
   - It is fine for the service to be created afterwards, as it is only pointing traffic to a label that matches the pod.
3. We add a [ownerReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#ownerreference-v1-meta) entry in the `metadata.ownerReferences` array of the secret / service.

## Other things
I'm taking on some responsibilities to offload you in your work, here is a list.
- [ ] for @sstarcher  rebase on current master (no conflicts atm) and fixup some commits
- [ ] for @consideRatio and @sstarcher to find the configuration is comprehensible and make sense to me before we lock the api making future changes hard. This involves aligning with jupyterhub's configuration where we can.
- [ ] for @consideRatio to ensure the documentation is comprehensible enough for himself at least

sstarcher:
I rebased and squashed all commits.  I'll work on implementing the garbage collection setup as you described.

sstarcher:
If you and @yuvipanda can discuss the internal_ssl_directory settings.  I would certainly prefer to rely on the environment variables made available by JupyterHub so users don't have to set this value.

consideRatio:
# Situation summary

KubeSpawner need to get certificates generated by JupyterHub. This is currently attempted in this PR by getting KubeSpawner configured with the same configuration in JupyterHub which defaults to the relative path `internal-ssl` which I'm not sure if kubespawner could also use or not.

I think a better approach though, is too hook into the Spawner base class functions, specifically `move_certs`, as it seems intented for this! The [Spawner base class implementation of `move_certs` is a dummy](https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/spawner.py#L904-L930)!

The order of things happening will need to be:
1. Pod creation (`Spawner.start`)
2. Certificate creation - with info about the pod's IP and service name etc (`Spawner.create_certs`)
3. Secret creation (`KubeSpawner.move_certs`)

Hmmm... But [the current logic](https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/user.py#L551-L557) is to use the order of 2, 3, 1 it seems. Can this also work? I understand it as we need to have the IP of the pod to create the cert, is this correct?

# References
- [JupyterHub docs under spawners about encryption](https://jupyterhub.readthedocs.io/en/stable/reference/spawners.html#encryption)
- [JupyterHub docs under websecurity about encryption](https://jupyterhub.readthedocs.io/en/stable/reference/websecurity.html#encrypt-internal-connections-with-ssl-tls)
- https://github.com/jupyterhub/jupyterhub/pull/2055 - `internal_ssl` PR in JupyterHub: 
- https://github.com/jupyterhub/configurable-http-proxy/pull/170 - `internal_ssl` PR in configurable-http-proxy that channels traffic to the hub and the users, which jupyterhub configures dynamically
- [The `create_certs` and `move_certs` function definition in the base Spawner class](https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/spawner.py#L849-L930)

sstarcher:
Thanks for the detailed look.  I'll work on iteration 4 and see if I run into any issues.

sstarcher:
@consideRatio so Spawner is storing the certs in self.cert_path and get_env is storing them here
https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/spawner.py#L764

Prior I was using the environment data to define where those certs came from and was told not to do so.  I can also leverage self.cert_path directly to do this same thing.  I don't think it makes sense for us to override move_certs to store the information into a separate location as it's already in self.cert_path.

What do you think?

consideRatio:
@sstarcher okay so Spawner base class logic, is to add environment variables about paths, if the Spawner's cert_path is configured. That configurable option, [is undocumented where it is defined](https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/spawner.py#L218).

```python
    # jupyterhub's spawner.py, configuration of the Spawner base class
    internal_ssl = Bool(False)
    internal_trust_bundles = Dict()
    internal_certs_location = Unicode('')
    cert_paths = Dict()
```

But it seems like those are populated by a call to `move_certs` already.

```
            if self.settings.get('internal_ssl'):
                self.log.debug("Creating internal SSL certs for %s", spawner._log_name)
                hub_paths = await maybe_future(spawner.create_certs())
                spawner.cert_paths = await maybe_future(spawner.move_certs(hub_paths))
```

So, after `move_certs` has been called, the spawner will have access to `cert_paths`, which means we don't have to store that information away manually by overriding `move_certs` just to know where they are. Excellent!

We need to separating two locations when we consider this:
1. KubeSpawner needed to know where the certs generated by jupyterhub were located, so it can create a secret to later be mounted by the pod.
2. KubeSpawner needs to mount the certs someplace in the pod, which we currently have hardcoded to a location.

I've mostly focused on the first part at the moment, and noted that we currently have hardcoded the location where the user pod mounts them. If a JupyterHub user wants to provide certs instead of having jupyterhub generate them, then it can do so, and we would be fine we simply create a secret from a cert provided us no matter if JupyterHub created it or similar.

Regarding if the user pods mounts them in a certain location, and regarding how the user pod is actually able to use them (because just mounting a file someplace doesn't mean they are used of course), is another matter. And, for that, I think it may be actually needed to use environment variables. And, if we ensure the spawner's `cert_paths` is updated to reflect where they will be mounted on the user pod at some point, we can rely on `get_env` to have the user environment correctly setup automatically with that information. But that assumes `get_env` is consuming `cert_paths` when we have updated it properly. But that will be during pod creation...

We need to map out the order in which things need to happen, and in the order of things currently happen.

## The order of things need to happen i think...

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcblx0Y19jZXJ0W1NwYXduZXIuY3JlYXRlX2NlcnRzXVxuICBzdGFydFtLdWJlU3Bhd25lci5zdGFydF1cbiAgc2VjcmV0W0s4cyBTZWNyZXRdXG4gIHN2Y1tLOHMgU2VydmljZV0gIFxuICBzdmNfZG5bSzhzIFNlcnZpY2U6IERvbWFpbiBuYW1lXVxuICBzdmNfaXBbSzhzIFNlcnZpY2U6IENsdXN0ZXJJUF1cbiAgcG9kW0s4cyBwb2RdXG4gIHBvZF9pcFtLOHMgUG9kIENsdXN0ZXJJUF1cbiAgcG9kX3VpZFtLOHMgUG9kIFVJRF1cbiAgcG9kX3VpZCAtLT4gc3ZjXG4gIHBvZF9pcCAmIHN2Y19pcCAmIHN2Y19kbiAgLS0-IGNfY2VydFxuICBzdGFydCAtLT4gcG9kXG4gIHN2YyAtLT4gc3ZjX2lwXG4gIHBvZCAtLT4gcG9kX2lwICYgcG9kX3VpZFxuICBjX2NlcnQgLS0-IHNlY3JldFxuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

## The order it currently happens
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcblx0Y19jZXJ0W1NwYXduZXIuY3JlYXRlX2NlcnRzXVxuICBtX2NlcnRbU3Bhd25lci5tb3ZlX2NlcnRzXVxuICBzdGFydFtTcGF3bmVyLnN0YXJ0XVxuICBcbiAgY19jZXJ0IC0tPiBtX2NlcnQgLS0-IHN0YXJ0IiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

kubespawner/objects.py
consideRatio: @sstarcher: this is the current implementation in this section, which is running in JupyterHub / KubeSpawner.

```
    secret.data = {}
    user_ssl_key = "{}/user-{}/user-{}.key".format(ssl_directory, username, username)
    with open(user_ssl_key, 'r') as file:
        encoded = base64.b64encode(file.read().encode("utf-8"))
        secret.data['ssl.key'] = encoded.decode("utf-8")
```

This section should still not rely on a environment variable, as it is within the JupyterHub pod, and this section is where we will rely on information in `self.certs_path` instead or `local_spawner_certs_path` as we may have updated `self.certs_path` to represent the location for the user.

consideRatio: I'll mark this as unresolved to let you press resolve if you agree @sstarcher!

consideRatio:
> The alternative is to add code to jupyterhub-singleuser to read cert info from environment variables, rather than Secrets mounted. This has the added advantage that we can drop these variables before launching kernels - protecting this information from user code that does not need it - with a secret, all user code can read it. It will also make things simpler, since we won't create a Secret object. What do you think?

I know you agreed on using a Secret that is mounted later, but I just want to confirm that I think this is the right call. If it is correct, that the cert needs to be created knowing the IP of the pod, the pod must be registered with the api-server, and only when that has happened can the certs be created, and injected into the secret, which is then mounted by the pod that neededto wait for that before it successfully starts the container in the pod.

In other wrods:

Pod API server registration -> Pod ClusterIP -> Cert creation -> k8s Secret creation -> K8s secret mount on pod when the container is about to start.

sstarcher:
Do we have any reason to or need to honor the defaults for 
https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/spawner.py#L765

Vs overriding them to a better location.  Those are set to where those certificates exist in the JupyterHub pod which has no file system relation to the Jupyter pod that Kubespanwer launches.

I believe it makes sense for us to override those 3 environment variables for the pod that we are launching to different values such as I am currently doing.  The overriding and masking of self.cert_paths I believe will cause confusion.

I think we should make a clean separation between the pod that gets created and the configuration from JupyterHub.  It would make sense to keep them as is if we were sharing at file system.

sstarcher:
@consideRatio the certs are and can be created before the pod.  We do not rely on the pod IP.  We use the service that gives us a DNS entry so we don't need the POD IP.

consideRatio:
> Do we have any reason to or need to honor the defaults for
https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/spawner.py#L765 vs overriding them to a better location.

I suggest we honor the use of these in the user environment to locate the certs, but that we allow ourselves to edit their values. In a post above with a heading `Regarding self.cert_paths and user server's env vars` i discussed this.

sstarcher:
@consideRatio I certainly agree we set those for the user pod as that is how we tell Jupyter where to get the certs, but I don't believe they need to be the same values as JupyterHub is setting.

consideRatio:
> @consideRatio the certs are and can be created before the pod. We do not rely on the pod IP. We use the service that gives us a DNS entry so we don't need the POD IP.

To use the service DNS name, and not cluster IP, okay. But... What happens if you as a user pod want to communicate with the hub? Then I don't think you would be able to be recognized, because you would send traffic originating from the ClusterIP of the pod.

sstarcher:
The location of the traffic is a non-issue as far as I can tell.  The issue is when calling the user pod the name/ip has to be what is in the cert.  The origination is not validated by a SSL certificate.  The current setup with DNS works and we have been using it for months in production.

sstarcher:
The change to DNS should also fix this issue - https://github.com/jupyterhub/kubespawner/issues/223

consideRatio:
> but I don't believe they need to be the same values as JupyterHub is setting.

No I agree with that also.

But whatever `cert_paths` is when `get_env` is called from `kubespawner.start`, will make `get_env` return these environment variables pre-populated with the values of `cert_paths`. So, I suggest something like.


```python
# in kuberspawner.start

local_spawner_cert_paths = self.cert_paths
self.cert_paths = # whatever we want
env = self.get_env()

# creation of k8s secret: use local_spawner_cert_paths to find the certs to put in the k8s secret
```

sstarcher:
I would prefer instead of caching and modifying cert path we instead in the portion that creates the pod we do what I'm currently doing and override those 3 environment variables.  We will rely on less underlying mechanisms if we go this route.

consideRatio:
> The location of the traffic is a non-issue as far as I can tell. The issue is when calling the user pod the name/ip has to be what is in the cert. The origination is not validated by a SSL certificate. The current setup with DNS works and we have been using it for months in production.

**About not including the pod ClusterIP's on cert creation:**
I think it may be okay to accept all traffic towards the pod need to go through the k8s service, but it a compromise to require that and it could come with consequences, I'm not sure, so I dislike taking the risk.

**About it being an issue or not:**
I lack understanding to know, but unless it is tested to establish a communication from the pod to the hub, and thereby bypassing the service, that I think it is an issue. Sometimes the hub pod is used to provide endpoints that the user pods can speak with.

I'd like for all of these situations to work out well, and perhaps 2 and 4 that doesn't involve establishing a connection is easier, but I'd like to see it work out well also for 1 and 3 and not only for 1 which can benefit from the k8s service.
1. The user pod get incoming traffic (new connection)
2. The user pod responds to incoming traffic
3. The user pod establishes outgoing traffic (new connection)
4. The user pod get a response to outgoing traffic

consideRatio:
> I would prefer instead of caching and modifying cert path we instead in the portion that creates the pod we do what I'm currently doing and override those 3 environment variables. We will rely on less underlying mechanisms if we go this route.

It sounds okay to me. It would make us hardcode the names of the env vars in kubespawner, but it would avoid our need to rely on the details of JupyterHub.

sstarcher:
If we are on agreement with that should we make the location we mount into the pod configurable or is the default hard coding that I currently have sufficient?  Can you think of a use case where we would want to move that location?

consideRatio:
> The change to DNS should also fix this issue - #223

I think it would make the proxy's configuration about where the send traffic as configured dynamically by jupyterhub more reliable, but the core issue as I understand it now is that JupyterHub didn't realize the pod was killed, and that I fails to delete it it tries, which cause the user pod to obstruct the hubs ability to restart it.

I think it is fine to use the service name's dns record, but I'm vary on relying on it in case 1-4 as mentioned in https://github.com/jupyterhub/kubespawner/pull/386#issuecomment-608560537, so I currently still want to see us use known IPs when creating certificate.

I will try read up on this more though, I'm not knowledgeable enough on how certificates are used currently. I think their main purpose is when a connection is established (1 & 3).

consideRatio:
> If we are on agreement with that should we make the location we mount into the pod configurable or is the default hard coding that I currently have sufficient? Can you think of a use case where we would want to move that location?

I'm not sure, but adding a configuration for it is quite easy, and provides a way for documentation to present the path we use which help us give insight about that without writing explicit documentation that can get outdated. So, I think we should make it configurable where we use our currently hardcoded value as a default value, because it help us with the documentation part.

sstarcher:
> > The location of the traffic is a non-issue as far as I can tell. The issue is when calling the user pod the name/ip has to be what is in the cert. The origination is not validated by a SSL certificate. The current setup with DNS works and we have been using it for months in production.
> 
> **About not including the pod ClusterIP's on cert creation:**
> I think it may be okay to accept all traffic towards the pod need to go through the k8s service, but it a compromise to require that and it could come with consequences, I'm not sure, so I dislike taking the risk.

I actually view this as the opposite using the ClusterIP is a risk.  If the pod is restarted in place then it will no longer work.  When using the DNS it gives you a static name that resolves these issues.  Kubespawner already has tickets filed against it for this exact issue that using the DNS will fix.


> 
> **About it being an issue or not:**
> I lack understanding to know, but unless it is tested to establish a communication from the pod to the hub, and thereby bypassing the service, that I think it is an issue. Sometimes the hub pod is used to provide endpoints that the user pods can speak with.
> 
> I'd like for all of these situations to work out well, and perhaps 2 and 4 that doesn't involve establishing a connection is easier, but I'd like to see it work out well also for 1 and 3 and not only for 1 which can benefit from the k8s service.
> 
> 1. The user pod get incoming traffic (new connection)
The user pod will register with the proxy using the dns name.  Anything going through the proxy should function without change.

> 2. The user pod responds to incoming traffic
This should be fully unaffected by the certificate itself.

> 3. The user pod establishes outgoing traffic (new connection)
This is also not effected by the user pods SSL cert.  This is fully tied to the SSL certs for the Proxy/Hub or whatever the pod is calling.

> 4. The user pod get a response to outgoing traffic
Similar to 2.

In this setup the certificate for the user should only effect #1 from my understanding.

consideRatio:
> If the pod is restarted in place then it will no longer work.

I think this is fine, consider:
- a container can restart, but that won't make the pod get a new ClusterIP (i think, not 100% confident, but 99%)
- a pod could be evicted, but then no new pod will be recreated, only pods backed by a replica set will be recreated if they are evicted. Also worth mentioning is, that the user pods, which are not backed by a replica set controller, will not be allowed to be relocated.

> Kubespawner already has tickets filed against it for this exact issue that using the DNS will fix.

I recall some issues, don't think they are relevant to this consideration. What I'm suggesting is that we use both the cluster ip and the service name, as compared to only using the service name.

consideRatio:
> The user pod establishes outgoing traffic (new connection)
This is also not effected by the user pods SSL cert. This is fully tied to the SSL certs for the Proxy/Hub or whatever the pod is calling.

Excellent! So, if 2-4 is irrelevant, and its only 1 left of relevance, then not using either the cluster ip of the service or the pod sounds good to me.

## Iteration 123

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcblx0Y19jZXJ0W1NwYXduZXIuY3JlYXRlX2NlcnRzXVxuICBzdGFydFtLdWJlU3Bhd25lci5zdGFydF1cbiAgc2VjcmV0W0s4cyBTZWNyZXRdXG4gIHN2Y1tLOHMgU2VydmljZV0gIFxuICBzdmNfZG5bSzhzIFNlcnZpY2U6IERvbWFpbiBuYW1lXVxuICBwb2RbSzhzIHBvZF1cbiAgcG9kX3VpZFtLOHMgUG9kIFVJRF1cbiAgc3ZjX2RuIC0tPiBjX2NlcnRcbiAgY19jZXJ0IC0tPiBzdGFydFxuICBzdGFydCAtLT4gcG9kXG4gIHBvZF91aWQgLS0-IHN2YyAmIHNlY3JldFxuICBwb2QgLS0-IHBvZF91aWRcbiIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

sstarcher:
Sounds good, so let me outline what I believe needs changed from the current code.

* remove from the _init the SSL logic and move that to create_certs
* Move service and secret creation to after pod creation and set the ownerReference information with the UID from the pod
* Remove the need for internal_ssl_directory and use cert_paths
* Add configuration option for the location the certificates are mounted into the pod

Let me know if you agree with the above and if I missed anything.

consideRatio:
@sstarcher I'm looking to acquire an overview of all certs involved, when they are needed, and what information they contain. I aim to document this properly also. Any help with understanding this better appreciated of course!

---

> remove from the _init the SSL logic and move that to create_certs

What did this refer to?

> Move service and secret creation to after pod creation and set the ownerReference information with the UID from the pod

:+1:

> Remove the need for internal_ssl_directory and use cert_paths

:+1:

> Add configuration option for the location the certificates are mounted into the pod

:+1:

sstarcher:
It refers specifically to this logic that is in the _init
https://github.com/jupyterhub/kubespawner/pull/386/files#diff-3082573191d96b79e2ec71986d4ef3e2R187

sstarcher:
@consideRatio I believe my last commit addressed your comments.  Let me know what you think.

kubespawner/objects.py
```diff
@@ -597,11 +599,25 @@ def make_ingress(
...
+            controller=False
+           )
 
 def make_secret(
     name,
     username,
-    ssl_directory,
+    cert_paths,
+    hub_ca,
+    owner,
```
consideRatio, line 620: We have had some history with failure to map the k8s naming and python naming etc before while it was possible, I think it is better to reduce confusion if you go for owner_references and then pass a list with one entry created by make_owner_reference to avoid any mismatchs of assumptions by mimicing k8s api tightly.

```suggestion
    owner_reference,
```

kubespawner/objects.py
```diff
@@ -615,8 +631,10 @@ def make_secret(
         going to be created in.
     username:
         The name of the user notebook.
-    ssl_directory:
+    cert_paths:
         Path to a directory containing all users certificates and keys.
```
consideRatio, line 635: This is later used as a dictionary, but I understand it to be a string from this docstring.

kubespawner/spawner.py
```diff
@@ -369,13 +366,12 @@ def _namespace_default(self):
...
-        None,
-        allow_none=True,
+    secret_mount_path = Unicode(
+        "/etc/jupyterhub/ssl/",
+        allow_none=False,
         config=True,
         help="""
-        Location of JupyterHub's internal_ssl directory.  This directory is managed by JupyterHub itself.  
-        See https://jupyterhub.readthedocs.io/en/stable/api/app.html
+        Location of the Spawned pod's ssl directory when JupyterHub has internal_ssl enabled.
```
consideRatio, line 374:
```suggestion
        Location to mount the spawned pod's certificates needed for internal_ssl functionality.
```

kubespawner/spawner.py
```diff
@@ -1542,6 +1542,15 @@ def is_pod_running(self, pod):
...
 
+    def is_pod_creating(self, pod):
+        """
+        Check if the given pod exists and has a UID
+
+        pod must be a dictionary representing a Pod kubernetes API object.
+        """
+
+        return pod and pod.metadata and pod.metadata.uid
+
```
consideRatio, line 1553: The name of this function doesn't really represent what it does. It sounds like it checks if it is in the process of being created.

Isn't it so that it will get an UID directly as part of the details you receive from the response when making the k8s API request to create the pod?

kubespawner/spawner.py
```diff
@@ -1796,6 +1805,14 @@ def start(self):
         self._start_future = self._start()
         return self._start_future
 
+    def create_certs(self):
+        """Set DNS alt name prior to cert creation"""
```
consideRatio, line 1809:
```suggestion
        """Overrides the base class Spawner's function to set the DNS names the
        certificate should be valid for before they are created."""
```

kubespawner/spawner.py
consideRatio:
```
        for i in range(retry_times):
            try:
                yield self.asynchronize(
                    self.api.create_namespaced_pod,
                    self.namespace,
                    pod,
                )
```

This will return the pod manifest in the end I think, which can be asserted to contain metadata.uid

kubespawner/spawner.py
```diff
@@ -1911,6 +1899,40 @@ def _start(self):
...
+                    self.api.create_namespaced_secret,
+                    namespace=self.namespace,
+                    body=self.get_secret_manifest(owner)
+                )
+            except ApiException as e:
+                self.log.info(e)
+                if e.status == 409:
+                    self.log.info("Secret " + self.secret_name + " already exists, so did not create new secret.")
+                else:
+                    raise
```
consideRatio, line 1922: Are user certificates are recreated by JupyterHub? This could cause issues. The service should be deleted when the past pod has been deleted, so I think it won't happen, but if this occurs, this will probably cause a proper failure for the user.

This should not be only self.log.info, or have a crash loop backoff attempt process.

kubespawner/objects.py
sstarcher: yes, I left the old doc string on accident good catch.

kubespawner/spawner.py
sstarcher: I don't believe we are guaranteed to get the UID returned in that manifest.  That code is set up as async could you clarify to me how I would get the result from that?

kubespawner/spawner.py
sstarcher: The self.log.info was a mistake.  As for the other stuff I'm not following what you saying could you please explain differently.  Whenever the pod is deleted both the secret and the service will be garbage collected.

sstarcher:
Updated except for the two above I still have comments on thanks again.

kubespawner/spawner.py
consideRatio: If we make kubespawner stop using the old coroutines + yeild, and the new `async` and `await`, `yeild` will be replaced with `await` - which help us better understand that we will actually from this section of code where `yeild` is called, await the response, and when we have it, return it.

So like this will work I think.

```python
        for i in range(retry_times):
            try:
                created_pod = yield self.asynchronize(
                    self.api.create_namespaced_pod,
                    self.namespace,
                    pod,
                )
```

kubespawner/spawner.py
consideRatio: It regards implementing the same kind of logic that can be found just above for attempting to create the pod multiple times and only accept failure after a while.

```python
        # If we run into a 409 Conflict error, it means a pod with the
        # same name already exists. We stop it, wait for it to stop, and
        # try again. We try 4 times, and if it still fails we give up.
        # FIXME: Have better / cleaner retry logic!
        retry_times = 4
        if self.modify_pod_hook:
            pod = yield gen.maybe_future(self.modify_pod_hook(self, pod))
        for i in range(retry_times):
            try:
                yield self.asynchronize(
                    self.api.create_namespaced_pod,
                    self.namespace,
                    pod,
                )
                break
            except ApiException as e:
                if e.status != 409:
                    # We only want to handle 409 conflict errors
                    self.log.exception("Failed for %s", pod.to_str())
                    raise
                self.log.info('Found existing pod %s, attempting to kill', self.pod_name)
                # TODO: this should show up in events
                yield self.stop(True)
                self.log.info('Killed pod %s, will try starting singleuser pod again', self.pod_name)
        else:
            raise Exception(
                'Can not create user pod %s already exists & could not be deleted' % self.pod_name)
```

consideRatio:
@sstarcher responded, thank you for the thorough work you put in on this! :heart: :star:

kubespawner/spawner.py
sstarcher: ahh got ya, I can add the retry logic.

sstarcher:
@consideRatio @yuvipanda let me know if I can do anything else to move this along.  I'll have time this week to make any additional changes.

sstarcher:
@consideRatio @yuvipanda just a reminder as I believe I have done everything that was asked for.

abduMus:
Are there any updates yet on merging this pull request?

sstarcher:
Let me know if I can do anything else to move this forward we have been using it in production for months.

kubespawner/spawner.py
```diff
@@ -1794,11 +1880,12 @@ def _start(self):
             pod = yield gen.maybe_future(self.modify_pod_hook(self, pod))
         for i in range(retry_times):
             try:
-                yield self.asynchronize(
+                created_pod = yield self.asynchronize(
```
consideRatio, line 1883: As I understand it, this created_pod variable will be the return value of the request, equivalent to `kubectl create`, and that it will contain the pod manifest as now known by the k8s api server, which will need to have a `metadata.uid` field.

So therefore, if I'm not understanding this wrong, it doesn't make sense to have the following section below:

```
            yield exponential_backoff(
                lambda: self.is_pod_creating(self.pod_reflector.pods.get(self.pod_name, None)),
                'pod/%s does not exist!' % (self.pod_name),
            )
```

(This is a continuation of https://github.com/jupyterhub/kubespawner/pull/386/files#r404452602)

consideRatio:
@sstarcher thanks for your work on this, sorry for not following up better, we are bottlenecking on the review capacity available. I left a comment regarding one thing I still consider unresolved.

manics:
Do you have any thoughts on testing this in CI, either here or as part of https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1535? Obviously you've put in a lot of work, it'd be good to ensure we don't accidentlally break it in a future release.

consideRatio:
> Do you have any thoughts on testing this in CI, either here or as part of jupyterhub/zero-to-jupyterhub-k8s#1535? Obviously you've put in a lot of work, it'd be good to ensure we don't accidentlally break it in a future release.

I'm working in revising the CI system setup in general for z2jh. I'd love to see us able to test this in z2jh. At the moment I'm failing to keep up with the workload required to give you a speedy review experience. I hope to finish the rework of the CI system - then i can suggest how to test it better, because I don't want to set something up in the old system atm.

kubespawner/spawner.py
sstarcher: My experience was that the object returned without the UID.

sstarcher:
I have approx 2 weeks to get any additional changes merged.  Afterward, I will likely not have any more time to work on this PR for months.  Let me know what I can do to assist.

manics:
@consideratio given there's only 2 weeks to finish this off do you think it's worth getting a passing Z2JH CI scenario added using the existing setup?

- If the PR isn't merged we'll at least have a reproducible working test so it's possible for someone else to finish it off
- If it's merged hopefully the changes required for the new CI system are minimal

sstarcher:
Anyone is free to take this over.  If you have any limited changes you want me to make I can probably do that, but I will no longer have time for any more in-depth changes.

consideRatio:
Thank you so much for your work on this @sstarcher :heart:

I aim to pick this up unless someone does it before me.

scivm:
@consideRatio @sstarcher  What is preventing this from being merged already?  It may not be so important for online classes but industry is also using this for datascience and need the feature.

msummers42:
> Thank you so much for your work on this @sstarcher 
> 
> I aim to pick this up unless someone does it before me.

It looks like the test failure is unrelated to me, a timeout. Some of Shane's colleagues are happy to help out on this too.

manics:
@scivm Part of the problem is that this is part of a larger body of work (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1535) and the only way to test it is by manually setting it up. Inevitably this means it's going to take more more time.

sstarcher:
@manics I don't know if I agree with that assessment.   This PR handles TLS communication for kubespawner pods.  You don't need E2E TLS for all of JupyterHub for TLS for this to be verified.

manics:
@sstarcher thanks for the clarification.

PR #387, opened 2020-01-06
Add MultiNamespacedKubeSpawner

athornton:


athornton:
This now spawns user pods correctly, but the event reporting isn't working yet, so the Hub never sees the pod move out of spawn-pending.

kubespawner/multispawner.py
```diff
@@ -417,16 +415,16 @@ def _wait_for_namespace(self, timeout=30):
 
     @gen.coroutine
     def _define_namespaced_account_objects(self):
```
lresende: Should these be moved to objects.py where all K8s actions are defined? My version of objects.py already had all the required functions defined following the existing pattern.

athornton:
Sorry it took me so long.  I think this one is ready for review.  In essence you need to make your spawner a MultiNamespacedKubeSpawner if that's the functionality you require.

I was reluctant to start messing with existing objects.py/spawner.py ; I feel like it's safer to leave the new stuff isolated in its own files initially.  Then as a subsequent PR we can, say, fold the reflector changes back into the base reflector class, and move the K8s-object-creation into objects.py.

athornton:
Or alternatively, should I move the k8s actions into a new file, multinamespaceobjects.py or some such?

Or am I just being too anxious and worrywartish and just smooshing everything together would be fine (like, the multi-namespace Reflector stuff duplicates a lot of code; if that went into reflector.py with the occasional if statement if omit_namespace is set, it would reduce that duplication a lot).  The problem here is that I don't have a convenient test rig and so what I'm testing is basically "does it work for me" which means I am very wary of breaking the original functionality.

athornton:
Asking again: *I* think it's a better idea to leave as much of the original stuff untouched as possible for an initial implementation.  I'm OK with doing a multiobjects.py or whatever to break manipulation of k8s objects out, and then once that's merged, doing a separate PR to recombine multi and single-namespace spawners.

Would anyone care to offer some guidance?

manics:
@athornton I think your idea makes sense- get a working implementation with minimal risk to existing users, then refactor and get more testers are involved. Have you looked into adding tests in this repo?

@consideRatio could this be (pre?)released as a new experimental feature, marked as "experimental, needs testing, subject to breaking changes"?

consideRatio:
@manics @athornton I have not god time to work on thinking this through properly myself, but I really appreciate the thinking that you convey. My concern for merging something would mainly be the risk of breaking something for someone, or misleading someone that something being in a very alpha state was more robust than it was.

But if the focus is early stage development only, I'd suggest a fork and a custom hub image on z2jh to work quicker. A custom hub image is currently very easy to make and keep up to date with development in z2jh because we mount the configuration that is important and relatively often changing to the hub container anyhow. The hub image itself only contains the packages etc.

What are the pro/cons of forking?
con: travis-ci not setup
con: some may loose track of ongoing development, but i for example would start watching the fork. Or wait... the fork. Eh? Wait. I just realize this is already developing in a fork...
pro: possible to make quick and wild modifications and quite simply.
neutral: I think it makes even more sense to inject a installation of current master before starting jupyterhub in the hub container instead of rebuilding the image etc to avoid slow development of kubespawner changes.

Hmmm... Thinking of it, I think something very immature belongs in a fork, and with some maturity it can benefit of being available officially to everyone and documented alongside the rest. Hmmm... it is all quite novel considerations to me.

Yikes. This became a written ramble of my thoughts developing. I now finish by concluding I'm very novel to all of these considerations and I believe in you to figure out what makes sense! :)

athornton:
So, here, in a nutshell, is my issue with doing a z2jh fork: I was asked to carry the PR for this feature, because I am certainly using multinamespace support in production (well, as much production as the Vera Rubin Observatory and its Legacy Survey of Space And Time (formerly LSST) has at the moment--we've just started taking images from the Auxiliary Telescope so we're still a couple years from survey operation, but I have...I dunno, a couple dozen users relying on it on a daily basis).

For various reasons, mostly historical, mostly not great, the Rubin Observatory is not using z2jh.  They mostly boil down to, "back in the day, I needed Jupyterlab support before z2jh had it."  But I don't have the time to maintain a test z2jh instance to try it out in (or at least, I don't _think_ I have time; if someone can convince me that this would be very low-effort, maybe I do?).

I was asked to carry the PR for the feature because there *are* other installations that would find per-user namespaces useful.  The version I'm actually running is a little different (it's at https://github.com/lsst-sqre/jupyterhubutils ) because I'm trying to encapsulate all the LSST (er, Rubin Observatory)-specific behavior into its own class hierarchy so I can leverage that to spawn Argo Workflows (as well as JupyterLab instances) based on my (custom) Lab images.  That approach, however, does nothing to help places that have their own "business logic" (ugh, hate that term) but would like to use per-user namespaces.

The point is, having *another* fork to make this easily testable with z2jh represents time and energy I don't believe I have (see above: AuxTel is coming on line--things at the summit are really heating up fast).  And maybe it's not another fork.  If someone with z2jh experience would like to run up an instance with this as its Kubespawner and verify that it doesn't break in single-namespace mode, I would be very grateful; I just don't think that I have the time. attention, and computational resources to invest in setting up another test stand at the moment, particularly since I have never actually used z2jh.

However, I don't really believe this code is extremely alpha.  LSST (now the Rubin Observatory) has been using dynamically-created-and-destroyed user workspaces for...well over a year now, anyway.  Granted we're probably not exercising all of the functionality, but the part we _are_ exercising (creating quota-ed namespaces and associated accounts and roles for user spawns, and tearing them down when the user leaves) is pretty robust.

So in short, that's why I think doing this (moving object support to multiobjects.py sounds sensible to me, and if people agree I can do that in the next week or two) inside kubespawner, but touching as little as possible of the single-namespace use case, is probably the right approach.  If we merge it, and people use it for a month or two without anything _outside_ the multinamespace support breaking (meaning we managed to not screw up any existing functionality), then (and only then) do we look at DRYing out the code by collapsing reflector and object support back into single source files with, presumably, a boolean selector to determine whether you want the single or multiple namespace behavior.

If it's catastrophic, we revert, increment the release version, and retrench.  But I don't think it's likely to be.

athornton:
Sorry, been pulled in other directions.  Will try to get back to this soonish.

manics:
Creating a fork for testing is really easy, especially with Travis. See for example
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1584

If you create your own Z2JH fork and enable travis on it you'll get all the CI testing for free, there's no need to maintain your own test infrastructure. For example, the Travis builds for my fork are on https://travis-ci.com/manics/zero-to-jupyterhub-k8s/branches

You will of course need to add tests for `MultiNamespacedKubeSpawner`, ideally in both this repo (unit tests) and Z2JH (integration tests). Without these it's likely the multi-namespace support will be inadvertently broken by a future commit.

athornton:
Thanks for the example!

I finally got a couple free hours to break out the K8s object support.  I hope that I can add some reasonable test cases in the next week or two, but anyone with spare cycles to help would be greatly appreciated.

praveenkanamarlapudi:
@athornton any updates?

#416 will be addressed with this PR.

athornton:
I never did get to testing it in a z2jh context; just about that time, COVID-19 ramped up, we all moved to working-from-home, and a bunch of other stuff came up at work.  I apologize for dropping the ball.

However, tomorrow I'm giving a presentation to my team about the State Of The Science Platform, so this will be brought up.  One of our goals is to reduce the bus factor of that work, so maybe we can put in some effort that isn't coming directly from me.

I've been using a minor variant quite happily for a while, although probably with minor changes.  My current state of things is in: https://github.com/lsst-sqre/jupyterhubutils/tree/master/jupyterhubutils/spawner

We'd need to pull out the eliot logging changes to make it generally consumable, but that's basically remove-a-line-and-outdent.

athornton:
If anyone's interested, that presentation I'm about to give is at: https://github.com/lsst-sqre/jupyter-overview-ajt-202007

The Jupyter-relevant thing in there besides "we need to test the user namespace support in z2jh" is that, if this gets successfully upstreamed, I'm going to start arguing for a more formal object management framework in the spawner.  We've been using this internally for a while, because our volume needs are complex, our options form can differ per-user, and once we introduced namespaces we needed serviceaccount/role/rolebindings in the user namespaces, and quotas on those namespaces.

Now certainly most sites don't need that level of complexity--for a classroom setting, it's superfluous.  But any place or project using a K8s Hub/Lab/Persistent-Storage setup as a major component of a Science Platform is eventually going to discover they want this capability, and breaking it out into an optional component of the KubeSpawner framework and upstreaming it would presumably mean that not every large project ends up rolling its own.

athornton:
Closing this one in favor of https://github.com/jupyterhub/kubespawner/pull/458 which sits atop current JupyterHub, includes a few unit tests, and works with z2jh with a little tweaking.

PR #389, opened 2020-01-23
log a warning if unrecognized user_options are provided

minrk:
should help debugging if folks aren't seeing the effects they expect

/cc @ankitladhania


consideRatio:
LGTM, the CI error is unrelated, feel free to self-merge.

PR #391, opened 2020-01-27
[MRG] Update contributing guide

betatim:
I wanted to work on kubespawner and first couldn't find the contributing guide because I looked in `CONTRIBUTING.md` which was empty. I've edited `CONTRIBUTING.md` to be similar to what we use for repo2docker https://repo2docker.readthedocs.io/en/latest/contributing/contributing.html and contain the contents of `SETUP.md`.

Once I found `SETUP.md` I had to make a few changes to the `jupyterhub_config.py` to get things to work locally (with jhub 1.1). These changes are included in this PR but I am unsure if they are needed/correct.

What do you think of these changes/re-organisation?


consideRatio:
It is a bit tricky for me to spot what you updated among what was just moved. Can you point it out with code review comments perhaps?

CONTRIBUTING.md
```diff
@@ -1,3 +1,189 @@
...
+    pods inside minikube.
+
+    ```bash
+    # Linux
+    sudo ip route add 172.17.0.0/16 via $(minikube ip)
+
+    # MACOS
+    sudo route -n add -net 172.17.0.0/16 $(minikube ip)
+    # later on you can undo this it with
+    sudo route delete -net 172.17.0.0
```
betatim: added this command

CONTRIBUTING.md
```diff
@@ -1,3 +1,189 @@
...
+   ```sh
+   git clone https://github.com/jupyterhub/kubespawner.git
+   ```
+
+1. Setup a virtual environment. After cloning the repository, you should set up an
+   isolated environment to install libraries required for running / developing
+   kubespawner.
+
+   There are many ways of doing this: conda envs, virtualenv, pipenv, etc. Pick
+   your favourite. We show you how to use venv:
```
betatim: New sentence

jupyterhub_config.py
```diff
@@ -4,8 +4,8 @@
 
 c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
```
betatim: All changes in this file are new

CONTRIBUTING.md
betatim: if someone knows what the linux equivalent is we can add that above as well

CONTRIBUTING.md
```diff
@@ -1,3 +1,189 @@
...
+We use Sphinx to build the kubespawner documentation. You can make changes to
+the documentation with any text editor and directly through the GitHub website.
+
+For small changes (like typos) you do not need to setup anything locally. For
+larger changes we recommend you build the documentation locally so you can see
+the end product in its full glory.
+
+To make edits through the GitHub website visit https://github.com/jupyterhub/kubespawner/tree/master/docs/source, open the file you would like to edit and then click "edit". GitHub will
+walk you through the process of proposing your change ("making a Pull Request").
+
```
betatim: everything above here is "new" in the sense that it didn't use to be anywhere in the kubespawner repo. it is "old" in the sense that it is copied more or less directly from repo2docker's contrib guide.

CONTRIBUTING.md
GeorgianaElena: The man page says:
```
ip route { add | del | change | append | replace } ROUTE
```
So, I believe the "undo" linux equivalent would be
```
sudo ip route del 172.17.0.0/16
```

betatim: Thanks for finding out, I added it.

consideRatio:
This LGTM! Thank you @betatim!

PR #393, opened 2020-01-27
Let kubespawner_override merge environment config instead of replacing it

scivm:
current behavior is that the kubespawner profile completely overrides the spawner environment.  I have set the AD_TOKEN into the spawner environment in pre_spawn_start so I don't want my kubespawner profile environment override to remove this which it currently does.

This PR changes the behavior from override to merge for dict() overrides.  Now the kubespawner environment dict is combined with the spawner environment dict where the kubespawner environment values overrides any existing spawner environment variables.

This also fixes an inconsistent behavior where after the first failed spawn the spawner and kubespawner are combined on the second try.  (not really sure how that was happening)

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1993,7 +1993,7 @@ class KubeSpawner(Spawner):
             else:
                 self.log.debug(".. overriding KubeSpawner value %s=%s", k, v)
             # if the override value is a dict and a similar dict exists in spawner do a shallow copy with 
-            # kubespawner key/value overriding simlar key in spawner.  This is important for example if you
+            # kubespawner key/value overriding similar key in spawner.  This is important for example if you
             # have set environment key/values in pre_spawn_start to prevent them from being removed.
             if isinstance(v,dict): 
                 self.log.debug(".. The value for key %s is a dict so we check if similar dict exists on spawner", k)
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1997,8 +1997,8 @@ class KubeSpawner(Spawner):
             if isinstance(v,dict): 
                 self.log.debug(".. The value is a dict so we will do shallow copy")
                 if (hasattr(self,k)):
-                    self.log.debug("... shallow copy of KubeSpawner key %s values %s over spawner %s", k,**self[k], **v)
-                    v = {**self[k], **v}
+                    self.log.debug("... shallow copy of KubeSpawner key %s values %s over spawner %s", k,getattr(self,k), v)
+                    v = {**getattr(self,k), **v}
             setattr(self, k, v)
 
     # set of recognised user option keys
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1992,6 +1992,10 @@ class KubeSpawner(Spawner):
                 self.log.debug(".. overriding KubeSpawner value %s=%s (callable result)", k, v)
             else:
                 self.log.debug(".. overriding KubeSpawner value %s=%s", k, v)
+                if (k == "environment"):
+                    if self.environment:
+                        self.log.debug(".. joining Spawner environment with KubeSpawner environment")
+                        v = {**self.environment, **v}
             setattr(self, k, v)
 
     # set of recognised user option keys
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1995,9 +1995,10 @@ class KubeSpawner(Spawner):
             # if the override value is a dict and a similar dict exists in spawner do a shallow copy with 
             # kubespawner values overriding values for existing keys in spawner. 
             if isinstance(v,dict): 
+                self.log.debug(".. The value is a dict so we will do shallow copy")
                 if (hasattr(self,k)):
-                    self.log.debug("... shallow copy of KubeSpawner key %s values %s over spawner %s", k,**self.k, **v)
-                    v = {**self.k, **v}
+                    self.log.debug("... shallow copy of KubeSpawner key %s values %s over spawner %s", k,**self[k], **v)
+                    v = {**self[k], **v}
             setattr(self, k, v)
 
     # set of recognised user option keys
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1993,11 +1993,11 @@ class KubeSpawner(Spawner):
             else:
                 self.log.debug(".. overriding KubeSpawner value %s=%s", k, v)
             # if the override value is a dict and a similar dict exists in spawner do a shallow copy with 
-            # kubespawner values overriding values for existing keys in spawner. 
+            # kubespawner key/value overriding simlar key in spawner. 
             if isinstance(v,dict): 
-                self.log.debug(".. The value is a dict so we will do shallow copy")
+                self.log.debug(".. The value for key %s is a dict so we check if similar dict exists on spawner", k)
                 if (hasattr(self,k)):
-                    self.log.debug("... shallow copy of KubeSpawner key %s values %s over spawner %s", k,getattr(self,k), v)
+                    self.log.debug("... shallow copy of KubeSpawner key %s values %s over spawner %s", k,v,getattr(self,k))
                     v = {**getattr(self,k), **v}
             setattr(self, k, v)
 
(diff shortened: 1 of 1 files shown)


kubespawner/spawner.py
```diff
@@ -1992,6 +1992,10 @@ def _load_profile(self, profile_name):
                 self.log.debug(".. overriding KubeSpawner value %s=%s (callable result)", k, v)
             else:
                 self.log.debug(".. overriding KubeSpawner value %s=%s", k, v)
+                if (k == "environment"):
```
consideRatio: The lines above have some conditional statements to log what goes on, and ensure that callable functions are evaluated for their value. This logic needs restructured in these. It should be allowed this environment key to be callable as well.

I think:
1. if we have a callable, evaluate it and log doing so in a log.debug
2. if we are merging/updating a dictionary such as we probably want to do in this override, log.debug doing so
3. if we are overriding an existing value, log.debug doing so

scivm: 1) My code now lets the callable do it's job and that sends info to log.debug.  It has no effect on existing values that are not dict so existing log.debuf is sufficient for those.
2) once we find a dict that may need merging, log.debug is sent prior to the attempt
3) If a similar dict for that key exists on spawner, we do shallow copy of kubespawner values over the spawner values.   This is also logged through log.debug.
4) I no longer hardcode environment.  Any similar dict on Kubespawner and Spawner will get the same functionality and logging.

manics:
This changes the default behaviour from overriding to merging. This makes sense when it's a dict of environment variables. Does this apply generally to all dict properties in KubeSpawner?

scivm:
Current dict in spawner and kubespawner objects:

spawner dict()
c.Spawner.environment
c.LocalProcessSpawner.environment 
c.LocalProcessSpawner.popen_kwargs

and kubespawner dict()
c.KubeSpawner.common_labels
c.KubeSpawner.environment
c.KubeSpawner.extra_annotations
c.KubeSpawner.extra_container_config
c.KubeSpawner.extra_labels
c.KubeSpawner.extra_pod_config
c.KubeSpawner.extra_resource_guarantees
c.KubeSpawner.extra_resource_limits
c.KubeSpawner.lifecycle_hooks
c.KubeSpawner.node_selector
c.KubeSpawner.storage_extra_labels
c.KubeSpawner.storage_selector

The dict name between spawner and kubespawner need to match so currently only the environment() dict will match and get the merge behavior.

Either way to make general case or hardcode "environment" is fine for me.

consideRatio:
I think using kubespawner_overrides where some will act as overrides and some will act as merges is problematic. I can imagine wanting to actually override, and actually wanting to merge, for the same dictionary.

I think I'd prefer an approach where we add kubespawner_merge or similar and do deep merges from it. Hmmm

manics:
If there's no simple solution perhaps this is a documentation issue- make it clearer how to use a `Callable` to do a merge?

scivm:
Now I have noticed that extra_containers that are defined in kubespawner_override have an empty env section.   Would we want extra_containers to get the environment variables we have defined in kubespawner_override?   Otherwise, there is no way to put environment variables to the extra_containers?

I am using extra_container sidecar to avoid putting everything in one container to avoid many GB size images and some users would not need this functionality anyway so no point to load it for them.

manics:
Another alternative, make the code that does the merging into a configurable:
https://github.com/jupyterhub/kubespawner/blob/86f1a0c2847e8ad874864efca95603ee9efc2b84/kubespawner/spawner.py#L1989-L1995

e.g. (untested)
```python
def default_properties_merge(self, overrides):
    for k, v in overrides.items():
        if callable(v):
            v = v(self)
            self.log.debug(".. overriding KubeSpawner value %s=%s (callable result)", k, v)
        else:
            self.log.debug(".. overriding KubeSpawner value %s=%s", k, v)
        setattr(self, k, v)

properties_merger = Callable(
        default_properties_merge,
        config=True,
)

...

self.properties_merger(kubespawner_override)
```

And you configure your own `properties_merger` to merge the keys in any way you want (e.g. replace, shallow, deep, append).

consideRatio:
Thank you for your work on this @scivm :heart:! I think this PR address a real problem, but that the suggested API for the user isn't good enough to merge. I will below suggest a workaround and suggest a path forward.

The PR as it currently is specifically _merges_ instead of _overrides_ environment configuration if part of `kubespawner_overrides`. I think this is problematic because its not obvious from the naming of `kubespawner_overrides`, and it is a fixed choice that may not be wanted at all time, and that it is specifically just for the environment configuration we do it in this way.

## Towards a suggestion to resolve the problem
For a path forward to resolve the real problem for users of KubeSpawners about wanting to merge the `environment` configuration, we need to come up with a concrete change proposal. If you have a proposal about this, please open an issue and feel free to ping me about it.

## Current available solutions
It is possible to configure this through a `KubeSpawner.pre_spawn_hook`.

```python
async def my_pre_spawn_hook(spawner):
    spawner.environment.update({
        "MY_ADDED_ENV": "WIEE",
    })

c.KubeSpawner.pre_spawn_hook = my_pre_spawn_hook
```

But, the key is probably to make use of `kubespawner_override` in a profile list, so then what? Something like this is possible.

```python
async def my_pre_spawn_hook(spawner):
    is_option_1 = "my_display_name_value" in spawner.user_options.get("profile") # {'profile': 'display_name of chosen profile'}

    spawner.environment.update({
        "MY_ADDED_ENV": "WIEE" if is_option_1 else "DOH",
    })

c.KubeSpawner.pre_spawn_hook = my_pre_spawn_hook
```

## Closing PR
I'll close this PR for now, but please respond if you think it should be reopened for further discussion. :sunflower:

consideRatio:
Hmmm... I think what I'd want is a way to configure logic to execute based on what profile was chosen in a sensible way, but the concept of a profile is specific to KubeSpawner, and I'd most of all like to see this kind of feature be implemented in the Spawner base class instead.

manics:
@consideRatio Do you think we could move the profiles support in KubeSpawner to [WrapSpawner](https://github.com/jupyterhub/wrapspawner), which would effectively make it portable to other Spawners?

consideRatio:
@manics perhaps, I know too little about wrapspawner to evaluate that technically, but I think it would be very useful to extract profile_list and similar out of KubeSpawner as its not really related to kubespawner but spawners in general.

consideRatio:
@manics btw, perhaps you could onboard me and potentially others briefly in the next jupyterhub team meeting about wrapspawner and profilespawner? I didn't get what wrapspawner did from the readme.

rkdarst:
> @consideRatio Do you think we could move the profiles support in KubeSpawner to [WrapSpawner](https://github.com/jupyterhub/wrapspawner), which would effectively make it portable to other Spawners?

I think that definitely would make sense.  wrapspawner will need some
love to make it work completely reliably (but it does work good enough
for most common use cases, there have mainly been some traitlets
quirks lately).

There is also this:
  https://github.com/jupyterhub/jupyterhub/issues/2724
which suggests to move this core function into JupyterHub itself,
which is probably better to work on than improving wrapspawner.

I am a big fan of both of these ideas, and would like to say I will
help do it.  Realistically, that won't happen in the near future, so
all I can do is help test and cheer you on.  But if I can help
somehow, let me know!

- Richard

rkdarst:
> @manics btw, perhaps you could onboard me and potentially others briefly in the next jupyterhub team meeting about wrapspawner and profilespawner? I didn't get what wrapspawner did from the readme.

wrapspawner is basically the "profile selection" option that exists in
kubespawer, but decoupled from the spawner itself (as one might argue
makes the most sense).  You define a child spawner, id, and a dict of
attributes to override on the child spawner.  When it's spawned, it
starts the child spawner class and then passes all the stuff through
to it.  Sounds a lot like something familiar.... (the kubespawner
profile list)

Best way to describe it is "exactly the profile_list built into
kubespawner".  I think the kubespawner one is slightly more advanced
because it has things like callable for profile list and other
configuration points.  It's also simpler because the spawner class is
fixed, so you don't have to deal with a proxy layer between spawners.

Combining kubespawer features into wrapspawer should in theory not be
too hard... will take some time to make the interface compatible with
kubespawner.  But wrapspawner is the recommend way of profiles in
batchspawner, so really the jump is not that big.

consideRatio:
Thank you @rkdarst for your help to make me understand! Grasping that wrapspawner is getting configured with a child-spawner was an important piece of understanding!

PR #403, opened 2020-05-03
[KubeIngressProxy] allow singleuser pods to use IPv6 addresses

stv0g:
This change replaces the IPv4 regex used in the proxy/ingress code by Python's ipaddress module which is included in Python since 3.3.

It allows us to add ingress resources for IPv6 reachable singleuser pods.

This PR is currently WIP since it requires more testing

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -2,6 +2,7 @@
 Helper methods for generating k8s API objects.
 """
 import base64
+import ipaddress
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


kubespawner/objects.py
```diff
@@ -511,46 +512,43 @@ def make_ingress(
...
 
     target_parts = urlparse(target)
-
-    target_ip = target_parts.hostname
     target_port = target_parts.port
 
-    target_is_ip = re.match('^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$', target_ip) is not None
+    try:
+        # Try to parse as an IP address
+        target_ip = ipaddress.ip_address(target_parts.hostname)
```
minrk, line 519: This is catching a possible ValueError in several places and treating them as invalid ip. Maybe just catch the parsing of the address?

Or am I missing which things might raise ValueError and why?

```suggestion
    try:
        # Try to parse as an IP address
        target_ip = ipaddress.ip_address(target_parts.hostname)
    except ValueError:
        target_is_ip = False
    else:
        target_is_ip = True
```

stv0g: Hi @minrk 

I addressed this in 2c3d9f5

kubespawner/objects.py
```diff
@@ -511,46 +512,48 @@ def make_ingress(
...
+    except ValueError:
+        target_is_ip = False
+    else:
+        target_is_ip = True
 
-    # Make endpoint object
-    if target_is_ip:
+    if target_is_ip
```
minrk, line 525: typo:

```suggestion
    if target_is_ip:
```

stv0g: Ahh sorry. I should test more often..

yuvipanda:
@minrk do you think this is ready to go?

yuvipanda:
uh oh, I think I screwed up the merge...

yuvipanda:
I just undid my work and let this be where it was. Sorry!

stv0g:
@consideRatio @yuvipanda 

I rebased my PR to the latest master branch and force-pushed it.

stv0g:
Hi @remche @yuvipanda @consideRatio 

Are there any objections to merge this?

yuvipanda:
Thanks for the ping, @stv0g!

stv0g:
Thanks :)

PR #406, opened 2020-05-20
[KubeIngressProxy] add a way to specify host for ingress

### LINKED ISSUE #404: Ingress rules w/o host wont match
### Bug description
Ingress rules for singleuser pods wont match if a rule with host field for hub server exists : 

```
NAME                                                      HOSTS                ADDRESS   PORTS   AGE
jupyter--2fuser-2fgpu-2f-route                            *                              80      42m
jupyterhub                                                gpu-hub.host.tld                80      46m
```

#### Workaround
Manually adding the host to the ingress fixes the mismatch. 

#### Expected behaviour
The proxy implementation should add the host field to the ingress rule, or at least optionally allow it.

### How to reproduce
I did test with Traefik and nginx-ingress controller.
Note that kubernetes-ingress (NGINXinc implementation) does not even support a rule with no host  : nginxinc/kubernetes-ingress#209.

### Your personal set up
- K8s v1.15.11 (RKE)
- nginx-0.25.1-rancher1 or Traefik 2.0.6

remche:
This PR fixes #404 by adding a way to specify host for ingress objects :
```
c.KubeIngressProxy.ingress_host = 'host.tld'
```
I wonder if we should allow an array here ?


manics:
Are there any other parameters that may be needed for creating the ingress, for instance SSL configuration? If so it'd be good to consider those other use-cases now even if they're not yet implemented, so that we can minimise future confusion if they're added later.

consideRatio:
Thanks for working on improving this project!

I lack a lot of background about KubeIngressProxy and don't understand it yet, but a key idea that I and others have partially strayed from in the past with negative consequences have been to implement something that doesn't map so directly to the Kubernetes API specification, but instead for example makes an array field of strings a single string.

Does this change map well to a K8s api specification?

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#ingress-v1beta1-networking-k8s-io

kubespawner/objects.py
```diff
@@ -490,11 +490,17 @@ def make_ingress(
...
+        path = routespec
+    else:
+        host, path = routespec.split('/', 1)
+
     meta = V1ObjectMeta(
         name=name,
         annotations={
             'hub.jupyter.org/proxy-data': json.dumps(data),
-            'hub.jupyter.org/proxy-routespec': routespec,
+            'hub.jupyter.org/proxy-routespec': path,
```
minrk, line 503: what's the reason to make the proxy-routespec not contain the routespec anymore, and only the path component? I think this should probably stay as it is, but if the path and/or host should *also* be included, those should be additional fields.

remche: If we keep the full routespec here, it does not match the hub `_service_map` and get destroyed and recreated at each `check_routes` call.

chancez:
This should probably support multiple hostnames since jupyterhub supports accessing notebook servers on a per-user subdomain basis, meaning you would need to create an ingress with two `hosts`, eg: `jupyterhub.example.org` and `*.jupyterhub.example.org`.

kubespawner/proxy.py
```diff
@@ -68,6 +68,13 @@ class KubeIngressProxy(Proxy):
         """
     )
 
+    ingress_host = Unicode(
+        config=True,
+        help="""
+        Hostname for ingress rules
+        """
+    )
+
```
consideRatio, line 77: I agree with https://github.com/jupyterhub/kubespawner/pull/406#issuecomment-666666196 that we should consider if we have multiple hosts.

This is a bit complicated, because the k8s Ingress resource expose `spec.rules[].host` and `spec.tls.hosts`, so both singular and plural at the same time in a way. I'm not sure what configuration is in scope to define for this though =/

[K8s Ingress resource reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingress-v1beta1-networking-k8s-io)

txu0393:
This looks awesome! Is there a timeline for this?
It's probably out of scope for this PR but I find myself having to override the `make_ingress` function so that I can add custom annotations on the Ingress resources as well.
We make use of `kubernetes.io/ingress.class` since we have multiple ingress controllers on our k8s clusters

remche:
Now that `Ingress` is in GA I will resume the work on this PR. I did not plan to support annotation in this PR, especially since `kubernetes.io/ingress.class` is deprecated in favor of `ingressClassName` field, but it should not be very difficult to add it afterwards.

kubespawner/proxy.py
```diff
@@ -56,6 +59,17 @@ class KubeIngressProxy(Proxy):
...
 
+    ingress = List(
+        trait=Dict,
+        config=True,
+        help="""
+        List of ingress in the form of a dict {'host': 'host.example.fr,
+        'tlsSecret': 'tlsSecretName'}.
+
+        Wildcard might not work, refer to your ingress controler documentation.
+        """,
```
Id2ndR, line 70:
```suggestion
        List of ingress in the form of a dict {'host': 'host.example.fr,
        'secret_name': 'host-tls'}.

        Wildcard might not work, refer to your ingress controler documentation.
        """,
```
The tls secret is defined as [secret_name](https://github.com/jupyterhub/kubespawner/pull/406/files#diff-4c2e260a4476d246feddf1d239084673283d54250a7df9cecb79afa5eea459f5R779)

remche:
KubeIngressProxy seems broken in master since the recent since recent changes in `ResourceReflector`. I'm not comfortable enough with reflector mechanism, I will need time to dig deeper into the changes.

droctothorpe:
Any updates on this @remche?

droctothorpe:
FYI, https://github.com/jupyterhub/kubespawner/pull/515 should fix the reflector issues, @remche.

consideRatio:
Closing as addressed via #667!

PR #407, opened 2020-05-29
Add a unit test for the spawner

kpaschen:
This adds a test for the case where a pod is unschedulable and never leaves 'pending'.
Also a test for what happens when the pod reflector is added twice, and a basic test about creating a pvc (I only test the easy bit here where it raises an ApiException; for testing non-exception case would need to use a mock or a fixture that ensures the pvc is deleted afterwards).

I made two changes to the spawner that need review from someone who knows this code -- both are in places where a TimeoutException has been caught, and the spawner prints 'restarting pods'. The original code calls 'raise' right after that, and that seemed odd, so I changed it. If you think that's wrong, please let me know!

This brings coverage to 78.19%.

I've also noticed that proxy.py appears unused -- I wasn't sure of the status of that code though.

### DIFF
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -1812,7 +1812,7 @@ async def test_spawn_pending_pods(kube_ns, kube_client):
     c.KubeSpawner.start_timeout = 5  # Do not wait very long.
     spawner = KubeSpawner(hub=Hub(), user=MockUser(), config=c)
     with pytest.raises(TimeoutError) as te:
-      await spawner.start()
+        await spawner.start()
     assert 'pod/jupyter-fake did not start' in str(te.value)
 
     pods = kube_client.list_namespaced_pod(kube_ns).items
@@ -1838,7 +1838,7 @@ async def test_spawn_watcher_reflector_started_twice(config):
 
     with pytest.raises(ValueError) as ve:
         spawner.pod_reflector.start()
-    assert ('Thread watching for resources is already running' in str(ve.value))
+    assert 'Thread watching for resources is already running' in str(ve.value)
     await spawner.stop()
 
 
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -1798,3 +1798,79 @@ async def test_ipv6_addr():
     )
     url = spawner._get_pod_url({"status": {"podIP": "cafe:f00d::"}})
     assert "[" in url and "]" in url
+
+
+async def test_spawn_pending_pods(kube_ns, kube_client):
+    """Spawner can deal with pods that are pending."""
+    c = Config()
+    # Make a config where the fake-jupyter pod cannot be scheduled so it
+    # will always be in a pending state.
+    # You can use any setting here that will be impossible to satisfy on the
+    # minikube cluster.
+    c.KubeSpawner.node_selector = {'disktype': 'ssd'}
+    c.KubeSpawner.namespace = kube_ns
+    c.KubeSpawner.start_timeout = 5  # Do not wait very long.
+    spawner = KubeSpawner(hub=Hub(), user=MockUser(), config=c)
+    with pytest.raises(TimeoutError) as te:
+      await spawner.start()
+    assert 'pod/jupyter-fake did not start' in str(te.value)
+
+    pods = kube_client.list_namespaced_pod(kube_ns).items
+    assert pods[0].status.phase == 'Pending'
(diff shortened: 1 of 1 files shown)


kubespawner/spawner.py
```diff
@@ -1842,7 +1842,8 @@ def _start(self):
                     self.pod_name,
                 )
                 self._start_watching_pods(replace=True)
-            raise
+            else:
+                raise
```
betatim, line 1846: It could be that we reraise the exception because even after restarting the reflector this particular attempt at spawning the pod won't recover. That is speculation though.

Maybe @yuvipanda or @manics know why the code is the way it is?

betatim:
Looks great. Let's see if someone remembers why/if the `raise` should be where it was before this PR or if that was a bug.

Otherwise I think merging this is a no brainer :D Thanks for taking the time to increase the coverage by adding some nice tests

kubespawner/spawner.py
manics: Sorry, I've no idea. The original code is from @minrk https://github.com/jupyterhub/kubespawner/commit/42a48acdbe4471562549307d0d0dff91e8a1b582

consideRatio: I think the original code make sense, but the change doesn't.

My understanding is that _start is supposed to fail with an error if it fails, hence `raise`. The logic about restarting the pod reflector here is to ensure that the Singleton reflector (keeps track of pods in the namespace for all spawner instances) isn't stuck in a disfunctional state by restarting it.

consideRatio, line 1846: As an example

```suggestion
            # a failure to start the server should result in a raised exception
            raise
```

kubespawner/spawner.py
```diff
@@ -1895,7 +1896,6 @@ def stop(self, now=False):
         except TimeoutError:
             self.log.error("Pod %s did not disappear, restarting pod reflector", self.pod_name)
             self._start_watching_pods(replace=True)
-            raise
```
consideRatio, line 1898: I'm not sure if it makes sense to raise an error or not in this case. Worth knowing is that in [this case](https://github.com/jupyterhub/jupyterhub/blob/0317fd63fa1dac4d709b4f95738e74c5c408da6c/jupyterhub/spawner.py#L1572-L1575) only a warning is logged.

consideRatio: I think it makes sense to keep this as well. I think doing so results in a clear indication of an issue in juppyterhub logs etc and that may be the right call. I do wonder what the outcome is for jupyterhub, it wont crash and restart right?

kubespawner/spawner.py
```diff
@@ -1895,7 +1896,6 @@ def stop(self, now=False):
         except TimeoutError:
             self.log.error("Pod %s did not disappear, restarting pod reflector", self.pod_name)
             self._start_watching_pods(replace=True)
```
consideRatio, line 1898:
```suggestion
            self._start_watching_pods(replace=True)
            raise
```

This is to revert a change that I think is problematic. By doing so, we can get this test merged.

tests/test_spawner.py
```diff
@@ -125,6 +127,64 @@ async def test_spawn(kube_ns, kube_client, config):
...
+async def test_spawn_pending_pods(kube_ns, kube_client):
+    """Spawner can deal with pods that are pending."""
+    c = Config()
+    # Make a config where the fake-jupyter pod cannot be scheduled so it
+    # will always be in a pending state.
+    # You can use any setting here that will be impossible to satisfy on the
+    # minikube cluster. 
+    c.KubeSpawner.node_selector = {'disktype': 'ssd'}
+    c.KubeSpawner.namespace = kube_ns
+    c.KubeSpawner.start_timeout = 30  # Do not wait very long.
```
consideRatio, line 140:
```suggestion
    c.KubeSpawner.start_timeout = 5  # Do not wait very long.
```

This duration will increase the time to complete the tests significantly, so I suggest to make this duration quite low.

yuvipanda:
image

Well, am here looking at this now :)

I 'solved' the merge conflicts and pushed this up, and will merge this if the tests pass.

yuvipanda:
Unfortunately it looks like the internals of kubespawner have changed enough in the last 4 years that these tests no longer apply :( I'm going to close this PR for now, with apologies for the delay that caused this - am sorry @kpaschen!

PR #408, opened 2020-06-03
handle pod url changes in poll

minrk:
this doesn't happen often, but seems to be possible in rare cases

this can wait for finishing up #386 (EDIT: #409) which also changes the URL returned by start

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1952,11 +1952,11 @@ class KubeSpawner(Spawner):
         if not self.pod_reflector.first_load_future.done():
             await asyncio.wrap_future(self.pod_reflector.first_load_future)
         ref_key = "{}/{}".format(self.namespace, self.pod_name)
-        data = self.pod_reflector.pods.get(ref_key, None)
-        if data is not None:
-            if data["status"]["phase"] == 'Pending':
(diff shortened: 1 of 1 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/proxy-establishing-route-to-wrong-pod-ip/4571/4

minrk:
@consideRatio this is now much smaller, since #409 added get_pod_url. This PR now only adds the extra check during poll instead of also defining get_pod_url itself.

kubespawner/spawner.py
```diff
@@ -1969,6 +1969,18 @@ async def poll(self):
                             await self.stop(now=True)
                         return c["state"]["terminated"]["exitCode"]
                     break
+
+            # pod running. Check and update server url if it changed!
+            # only do this if fully running, not just starting up
+            if self.is_pod_running(pod):
+                pod_url = self._get_pod_url(pod)
+                if self.server.url != pod_url:
```
consideRatio, line 1977: It appears this can fail like this

image

minrk: I suspect this is related to our mocking and not a 'real' possibility, but added an `if self.server` just to be safe, since it only makes sense to do this check when both are defined and can be compared.

Issue #413, opened 2020-06-30
Add ability to override component label 'singleuser-server'


harsimranmaan:
### Proposed change
The current component label is hard coded and does not allow for need based customization.
Add an option to kubespawner to allow setting `component` label to a value other than 'singleuser-server'.

### Who would use this feature?

Would be really helpful in running multiple Jupyterhub instances in the same namespace.

### (Optional): Suggest a solution

Add a config option called `component-label` that defaults to 'singleuser-server' but can be set to another value if needed. Maybe add some validation on what values are accepted like format and length.

Happy to send a PR.

consideRatio:
Closed by #418

PR #418, opened 2020-07-17
Add component_label property to support multiple hub instances in the…

harsimranmaan:
… same namespace
When multiple jupyterhub instances are run the same namespace, the spawner reflector is hard-wired to listen for `singleuser-server` events. This makes every JH instance listen to events for pods that are not managed by that JH. The change is to ensure that the spawner could be configured to listen to pods as specified in the config rather than hard-code it. The change is backwards compatible. 

Addresses #413

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -463,6 +463,7 @@ def make_ingress(
         name,
         routespec,
         target,
+        labels,
         data
 ):
     """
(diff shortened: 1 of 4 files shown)


consideRatio:
@harsimranmaan please provide a proper description of what you have done and why. Also please self-review if you introduce any breaking change or similarly.

harsimranmaan:
@consideRatio  Updated description to include more info. I have tested this to ensure that it is backwards compatible and also works when a component_label is specified.

harsimranmaan:
I have been this change in the application for over a month now.

yuvipanda:
Thanks a lot for the PR, @harsimranmaan :)

PR #420, opened 2020-07-21
Log thread pool worker count on init

mriedem:
This simply logs the number of workers in the executor
thread pool when we create it.


yuvipanda:
Thanks, @mriedem!

Issue #423, opened 2020-08-14
High CPU usage in the reflector


rmoe:
The Python Kubernetes client by default deserializes API responses into objects. This takes a lot of time when you're processing events for thousands of pods. During a test run where we created 2000 users and servers the hub process spent a large percentage of its CPU time just deserializing these responses. Here's a flamegraph from [py-spy](https://github.com/benfred/py-spy) that was captured during the entire scale test:

![image](https://user-images.githubusercontent.com/26524/90191483-d1dc2880-dd75-11ea-9125-94f3c59d10ab.png)
During that run 54% of the hub's CPU time was in `_watch_and_update` and within that method it was dominated by calls to the api deserialization methods.

If the reflector deals in JSON instead the graph from an identical run looks like this:
![image](https://user-images.githubusercontent.com/26524/90264268-af91eb80-de05-11ea-810b-212f231cc995.png)

(Github won't let me upload the SVGs py-spy generates but I can provide those another way if anybody wants to drill into them).

During these scale tests the wall time between the runs doesn't actually vary that much. However the reduced CPU load makes the hub API much more respsonsive. With the default kubespawner when scaling from 1000 to 2000 (and beyond) new servers take progressively longer to spawn and the API becomes unresponsive. Those runs are littered with errors like:

```
2020-08-13 14:38:13,454 WARNING [urllib3.connectionpool] Retrying (Retry(total=9, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='host', port=443): Read timed out. (read timeout=30)")': /hub/api/users/hub-stress-test-1361/server
```
Those errors go away when we don't deserialize the responses.

### Proposed change
The watcher can pass `_preload_content=False` to the method in `list_method_name` which will return JSON from the API. The reflector can just keep that data as dictionaries instead of deserialized objects. 

### Who would use this feature?

Everybody

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/identifying-jupyterhub-api-performance-bottleneck/1289/6

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/core-component-resilience-reliability/5433/6

consideRatio:
[...]

PR #424, opened 2020-08-14
Breaking change / performance: don't make kubernetes-client deserialize k8s events into objects

rmoe:
Turning API responses into Python objects take a significant
amount of CPU time when dealing with a large number of events.

Fixes: #423

---

## Breaking changes

- The Kubernetes EventsReflector, which is providing the KubeSpawner instances with information about [Kubernetes Events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#event-v1-core) describing events for other resources, is now exposing events as python dictionaries rather than `V1Event` objects. `V1Event` is defined in the kubernetes-client/python library as a representation of a Kubernetes Event.
- KubeSpawner's `.progress` method implementation (https://github.com/jupyterhub/jupyterhub/pull/1771) which is generating a formatted `message` as well as a KubeSpawner specific `raw_event` entry now returns the `raw_event` as a Python dictionary with entries formatted in `camelCase` where the keys were formatted in `snake_case`.

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -152,12 +152,12 @@ class NamespacedResourceReflector(LoggingConfigurable):
 
         Overwrites all current resource info.
         """
-        method = partial(getattr(self.api, self.list_method_name), _preload_content=False)
-        initial_resources = method(
+        initial_resources = getattr(self.api, self.list_method_name)(
             self.namespace,
             label_selector=self.label_selector,
             field_selector=self.field_selector,
             _request_timeout=self.request_timeout,
+            _preload_content=False,
         )
         # This is an atomic operation on the dictionary!
         initial_resources = json.loads(initial_resources.read())
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -2,6 +2,8 @@
 # asyncio Futures cannot be used across threads
 from concurrent.futures import Future
 
+from functools import partial
+import json
 import time
 import threading
 
@@ -150,16 +152,18 @@ class NamespacedResourceReflector(LoggingConfigurable):
 
         Overwrites all current resource info.
         """
-        initial_resources = getattr(self.api, self.list_method_name)(
+        method = partial(getattr(self.api, self.list_method_name), _preload_content=False)
+        initial_resources = method(
             self.namespace,
             label_selector=self.label_selector,
             field_selector=self.field_selector,
             _request_timeout=self.request_timeout,
         )
         # This is an atomic operation on the dictionary!
-        self.resources = {p.metadata.name: p for p in initial_resources.items}
+        initial_resources = json.loads(initial_resources.read())
+        self.resources = {p["metadata"]["name"]: p for p in initial_resources["items"]}
         # return the resource version so we can hook up a watch
-        return initial_resources.metadata.resource_version
+        return initial_resources["metadata"]["resourceVersion"]
 
     def _watch_and_update(self):
         """
@@ -219,10 +223,11 @@ class NamespacedResourceReflector(LoggingConfigurable):
                 if self.timeout_seconds:
                     # set watch timeout
                     watch_args['timeout_seconds'] = self.timeout_seconds
+                method = partial(getattr(self.api, self.list_method_name), _preload_content=False)
                 # in case of timeout_seconds, the w.stream just exits (no exception thrown)
                 # -> we stop the watcher and start a new one
                 for watch_event in w.stream(
-                    getattr(self.api, self.list_method_name),
+                    method,
                     **watch_args
                 ):
                     # Remember that these events are k8s api related WatchEvents
@@ -236,10 +241,10 @@ class NamespacedResourceReflector(LoggingConfigurable):
                     resource = watch_event['object']
                     if watch_event['type'] == 'DELETED':
                         # This is an atomic delete operation on the dictionary!
-                        self.resources.pop(resource.metadata.name, None)
+                        self.resources.pop(resource["metadata"]["name"], None)
                     else:
                         # This is an atomic operation on the dictionary!
-                        self.resources[resource.metadata.name] = resource
+                        self.resources[resource["metadata"]["name"]] = resource
                     if self._stop_event.is_set():
                         self.log.info("%s watcher stopped", self.kind)
                         break
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -106,7 +106,7 @@ class EventReflector(NamespacedResourceReflector):
         #   timestamp without and the other is a higher resolution timestamp.
         return sorted(
             self.resources.values(),
-            key=lambda event: event.last_timestamp or event.event_time,
+            key=lambda event: event["lastTimestamp"] or event["eventTime"],
         )
 
 
@@ -1492,10 +1492,10 @@ class KubeSpawner(Spawner):
         # FIXME: Validate if this is really the best way
         is_running = (
             pod is not None and
-            pod.status.phase == 'Running' and
-            pod.status.pod_ip is not None and
-            pod.metadata.deletion_timestamp is None and
-            all([cs.ready for cs in pod.status.container_statuses])
+            pod["status"]["phase"] == 'Running' and
+            pod["status"]["podIP"] is not None and
+            "deletionTimestamp" not in pod["metadata"] and
+            all([cs["ready"] for cs in pod["status"]["containerStatuses"]])
         )
         return is_running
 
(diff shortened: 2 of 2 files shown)


kubespawner/reflector.py
```diff
@@ -150,16 +152,18 @@ def _list_and_update(self):
...
+        initial_resources = method(
             self.namespace,
             label_selector=self.label_selector,
             field_selector=self.field_selector,
             _request_timeout=self.request_timeout,
         )
         # This is an atomic operation on the dictionary!
-        self.resources = {p.metadata.name: p for p in initial_resources.items}
+        initial_resources = json.loads(initial_resources.read())
+        self.resources = {p["metadata"]["name"]: p for p in initial_resources["items"]}
```
mriedem, line 164: Technically this changes the value on the `resources` dict from an object to a dict so the help here should probably be updated as well:

https://github.com/jupyterhub/kubespawner/pull/424/files#diff-2f4d4efedf466e7bf4f8f083d0c722adR52

One thing I'm not sure about is how much this might impact out-of-tree reflectors that are expecting `self.resources` to return objects rather than dicts, i.e. this could be a breaking change for them. One could maybe mitigate that by wrapping the dicts in simple objects that pass through `__getattr__` to `__getitem__` but as we can see below with `resource_version` we'd also have to camel-case-ify some of the attributes, like [this](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/fcf9c2e92a450e175a9d53abbe6dbbb2936df096/jupyterhub/files/hub/jupyterhub_config.py#L39).

I'm not sure what the ABI contract on something like this is within the KubeSpawner. Would a release note be sufficient?

kubespawner/spawner.py
```diff
@@ -106,7 +106,7 @@ def events(self):
         #   timestamp without and the other is a higher resolution timestamp.
         return sorted(
             self.resources.values(),
-            key=lambda event: event.last_timestamp or event.event_time,
+            key=lambda event: event["lastTimestamp"] or event["eventTime"],
```
mriedem, line 109: This is a good example of the kind of issue I'm talking about above. Handling this for the known event and pod reflectors within this repo is sufficient but I worry about breaking out of tree reflectors, but like I said I also don't know what contract is there.

kubespawner/spawner.py
```diff
@@ -1492,10 +1492,10 @@ def is_pod_running(self, pod):
         # FIXME: Validate if this is really the best way
         is_running = (
             pod is not None and
-            pod.status.phase == 'Running' and
-            pod.status.pod_ip is not None and
-            pod.metadata.deletion_timestamp is None and
-            all([cs.ready for cs in pod.status.container_statuses])
+            pod["status"]["phase"] == 'Running' and
+            pod["status"]["podIP"] is not None and
+            "deletionTimestamp" not in pod["metadata"] and
```
mriedem, line 1497: Hmm, maybe this should be:

`not pod["metadata"].get("deletionTimestamp")`

Because `deletionTimestamp` not being in the dict vs being in the dict with a None value are different things, unless that's not possible with how the kube API works?

kubespawner/spawner.py
```diff
@@ -1656,16 +1657,13 @@ async def progress(self):
                     # objects within it, and we need what we pass back to be
                     # serializable to it can be sent back from JupyterHub to
                     # a browser wanting to display progress.
```
mriedem, line 1659: This comment about the object not being serializable can go away.

kubespawner/spawner.py
```diff
@@ -1642,10 +1642,11 @@ async def progress(self):
                 # pod_id may change if a previous pod is being stopped
                 # before starting a new one
                 # use the uid of the latest event to identify 'current'
-                pod_id = events[-1].involved_object.uid
+                pod_id = events[-1]["involvedObject"]["uid"]
                 for i in range(next_event, len_events):
                     event = events[i]
                     # move the progress bar.
+
```
mriedem, line 1649: nit: unrelated change (try to avoid these because they can cause merge conflicts and bloat the size of the review)

kubespawner/spawner.py
```diff
@@ -1559,20 +1559,20 @@ def poll(self):
             yield self.pod_reflector.first_load_future
         data = self.pod_reflector.pods.get(self.pod_name, None)
```
mriedem, line 1560: The docstring on the `PodReflector.pods` attribute should probably be updated because it talks about the value being an object representation which it no longer is.

https://github.com/jupyterhub/kubespawner/pull/424/files#diff-3082573191d96b79e2ec71986d4ef3e2R69

mriedem:
>The bigger concern is the impact to out of tree reflectors

By the way, I'm assuming you can have out of tree reflectors but I'm not entirely sure if you can. The KubeSpawner docs don't mention anything about entry points to load your own reflectors:

https://jupyterhub-kubespawner.readthedocs.io/en/latest/reflector.html

kubespawner/spawner.py
rmoe: `deletionTimestamp` is only set by the server when a deletion is requested. It shouldn't exist at any other time.

kubespawner/spawner.py
rmoe: Good catch. I'll get all of these comments and docstrings fixed.

kubespawner/reflector.py
```diff
@@ -150,16 +152,18 @@ def _list_and_update(self):
...
         Overwrites all current resource info.
         """
-        initial_resources = getattr(self.api, self.list_method_name)(
+        method = partial(getattr(self.api, self.list_method_name), _preload_content=False)
+        initial_resources = method(
             self.namespace,
             label_selector=self.label_selector,
             field_selector=self.field_selector,
             _request_timeout=self.request_timeout,
         )
```
consideRatio, line 161: Is it correct that introducing partial accomplished the same thing as just adding _preload_content=False to the list of arguments? If so, is there a motivation to using partial over just adding the argument to the list?

```suggestion
        initial_resources = getattr(self.api, self.list_method_name)(
            self.namespace,
            label_selector=self.label_selector,
            field_selector=self.field_selector,
            _request_timeout=self.request_timeout,
            _preload_content=False,
        )
```

consideRatio:
## Notification / Question

@clkao this PR will impact the users of `raw_event` returned by `progress()`. I think the difference will be that you get camelCase instead of snake_case, but still get a python dictionary. @rmoe do you think I described this change correctly?

## My conclusion

Overall, I think this PR makes a lot of sense given the excellent report in #423. I think we should accept it, verify the breaking changes and report them in a release which needs to be minor rather than patch.

## Breaking changes
This is what I pick up to be breaking changes.

- Progress() returned `raw_event` changed from python dictionary formatted with snake_case to a python dictionary formatted with camelCase (I think).
- A spawner's events property populated by the Singleton events reflector now returns python dictionaries instead of V1Event objects from kubernetes-client/python.

mriedem:
FWIW I agree.

>Progress() returned raw_event changed from python dictionary formatted with snake_case to a python dictionary formatted with camelCase (I think).

`raw_event` was a dict because the V1Event object was converted to a dict [here](https://github.com/jupyterhub/kubespawner/pull/424/files#diff-3082573191d96b79e2ec71986d4ef3e2L1659). I'm not sure what `event.to_dict()` does regarding the field names but assuming they were snake-case because that's [what they were](https://github.com/jupyterhub/kubespawner/pull/424/files#diff-3082573191d96b79e2ec71986d4ef3e2L1666) on the `event` object and [now they are camel-case](https://github.com/jupyterhub/kubespawner/pull/424/files#diff-3082573191d96b79e2ec71986d4ef3e2R1660).

>A spawner's events property populated by the Singleton events reflector now returns python dictionaries instead of V1Event objects from kubernetes-client/python.

Yup, the docstring for the `events` property function was updated to reflect that [here](https://github.com/jupyterhub/kubespawner/pull/424/files#diff-3082573191d96b79e2ec71986d4ef3e2R91).

I [mentioned](https://github.com/jupyterhub/kubespawner/pull/424/files#r470732969) that we could try to still return an object with some massaging of the field names but that could get complicated. We'd probably know if that worked if we didn't have to make some of the changes in this patch, but it's hard to say for sure since there are no tests (that I could find) for the EventReflector code.

kubespawner/reflector.py
rmoe: There was no reason. I just copied what I did in `_watch_and_update` (where partial application is required).

rmoe:
> @clkao this PR will impact the users of `raw_event` returned by `progress()`. I think the difference will be that you get camelCase instead of snake_case, but still get a python dictionary. @rmoe do you think I described this change correctly?

That's correct. They were snake case because of how those OpenAPI objects got converted to dictionaries. Now we just have the response from the Kubernetes API as a dictionary and everything the API returns is camel cased.

consideRatio:
@rmoe thank you soo much for your thorough investigation and this PR to resolve the found performance issue! Thank you @mriedem for your review work! :heart: :100: :tada:

@jupyterhub/jupyterhubteam this LGTM but it would be nice to get another LGTM before I press merge. It solves a real problem when scaling to a very large amount of users so it motivates some breaking changes I think.

consideRatio (approved):
Excellent work! :heart:

yuvipanda:
\o/ Thanks a lot for this, @rmoe and @mriedem! 

And thanks for your review, @consideRatio :)

yuvipanda:
I was *just* dealing with an outage caused by slow response times cascading because of high CPU usage... :)

PR #426, opened 2020-09-01
Support EnvVar's with 'valueFrom' as well as with 'value'

### LINKED ISSUE #306: Regarding c.KubeSpawner.environment and the EnvVar structure's valueFrom field
I tried
```yaml
hub:
  extraConfig:
    jupyterlab: |
      c.KubeSpawner.extra_container_config = {
        "env": [{
          "name": "MY_POD_IP",
          "valueFrom": {
            "fieldRef": {
              "fieldPath": "status.podIP"
            }
          }
        }]
      }
```
But this will overwrite the default env, and cause error. Any way to append env instead of replace all env?

consideRatio:
## Background

The c.KubeSpawner.environment configuration can only accept a dictionary with keys and values being strings, because they will be passed as name and value to the name and value parameters of the EnvVar resource. But, Kubernetes EnvVar resources can have a name / valueFrom combo as well.

Due to this limitation, users ended up forced to do something very obscure like described https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1103#issuecomment-546139103.

## Solution

In this PR I enable users to pass other kinds of structures as well that will allow for any kind of valid k8s EnvVar structure to be passed. The format is best documented in the test currently.

## References
- Z2JH PR that relies on this PR to work: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1757
- Issue in Z2JH related to this limitation: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1103
- Issues in KubeSpawner related to this limitation: https://github.com/jupyterhub/kubespawner/issues/306#issuecomment-474934945 (Closes #306)
- [The base class definition of the `environment` traitlet](https://github.com/jupyterhub/jupyterhub/blob/cc8e7806530466dce8968567d1bbd2b39a7afa26/jupyterhub/spawner.py#L453) that KubeSpawner leaves untouched.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -274,12 +274,24 @@ def make_pod(
     if all([e is None for e in container_security_context.to_dict().values()]):
         container_security_context = None
 
+    # Transform a dict into valid Kubernetes EnvVar Python representations. This
+    # representation shall always have a "name" field as well as either a
+    # "value" field or "value_from" field. For examples see the
+    # test_make_pod_with_env function.
+    prepared_env = []
+    for k, v in (env or {}).items():
+        if type(v) == dict:
+            if not "name" in v:
+                v["name"] = k
+            prepared_env.append(get_k8s_model(V1EnvVar, v))
+        else:
+            prepared_env.append(V1EnvVar(name=k, value=v))
     notebook_container = V1Container(
         name='notebook',
         image=image,
         working_dir=working_dir,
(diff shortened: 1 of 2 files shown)


yuvipanda:
The more we can just pass through to k8s, the better. Thanks for the PR, @consideRatio

consideRatio:
Thanks @yuvipanda! :heart: :tada:

PR #431, opened 2020-09-11
Support for 'release' kubespawner config parameter. 

kristofmartens:
…n be used for templating multiple named kubernetes resouces to avoid name collisions when deploying jupyterhub multiple times in the same namespace

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -311,11 +311,11 @@ class KubeSpawner(Spawner):
         allow_none=True,
         config=True,
         help="""
-        The name of the helm release that is used for this jupyterhub
+        The name of the helm release that is used for this jupyterhub deployment.
         
-        When set to `None` (the default), releasename can not be expanded in various naming templates. 
+        When set to `None` (the default), releasename will not be expanded in various naming templates. 
         Being able to distinguish between releases is important to allow the same user to work on 
-        multiple jupyterhub deployments in the same namespace 
+        multiple jupyterhub deployments in the same namespace.
         """
     )
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -306,6 +306,19 @@ class KubeSpawner(Spawner):
         """
     )
 
+    release = Unicode(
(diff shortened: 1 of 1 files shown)


kristofmartens:
Currently working on allowing multiple jupyterhub releases in the same namespace. There are quite some named resources created in the kubespawner code that are not taking this into account. Adding support for a "releasename" template provides me the tools to overcome this issue

consideRatio:
I understand the use case, but I think the implementation needs to be more generic. Helm releases is a Helm specific concept, and they have for example updated their recommendations on labels to use. Now I believe app.kubernetes.io/instance is meant to be used instead in according to Helm's new best practices, and I think we should use configurations such as common_labels or something to configure what labels we want to use.

If we want to make it function well with specific labels, we also need kubespawner to inspect and manage pods with the right labels, and that is based on the PodReflector.

The PodReflector will look for `component: singleuser-server` labelled pods by default, and will be able to change this to `component: some-other-label-value` if you configure `c.KubeSpawner.component_label`, but it is in my mind a quite bad implementation at this point.

I think what makes sense is for kubespawner to have, is match_labels, and let that set be what it scans for, and then we can configure KubeSpawner in various different Helm releases to care for `app.kubernetes.io/instance: jupyterhub-2,app.kubernetes.io/component=singleuser-server` for example.

This kind of change would be nice to get, but such PR needs to be accompanied with careful consideration of what will break for users and how to migrate from an old version to a new version.

yuvipanda:
https://github.com/jupyterhub/kubespawner/pull/418 is related?

consideRatio:
Yepp @yuvipanda! Also https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1791 where I track work to mitigate this from z2jh.

By using the component label configuration, a user can install kubespawner in a single namespace, but I look to make kubespawner work not with a single label as match label, but instead allow for a set of labels just like a Deployment monitors its pods with a set of matchLabels.

I'll close this issue in favor of the z2jh issue for now.

PR #433, opened 2020-09-14
Add timeout and retry to create_namespaced_pod

gravenimage:
We've found persistent intermittent failures to spawn on Azure AKS (multiple k8s versions).   On a rancher cluster created on Azure VMs (k8s 1.18) we do not see this.   It appears that this particular POST call to the create_namespaced_pod API occasionally just doesn't return (within 5 minutes at least).   Adding a timeout and using the existing retry logic fixes this.  

Improvements that could be made:
 - explaining why k8s fails sometimes! 
 - could this happen on other k8s APIs? 
 - should this use a better backoff mechanism
 - how to avoid pulling in urllib3 dependency (used by kubernetes package) - or just accept it.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -221,14 +221,16 @@ class KubeSpawner(Spawner):
         """
     )
 
-    k8s_api_request_timeout_retries = Integer(
-        5,
+    k8s_api_request_retry_timeout = Integer(
+        30,
         config=True,
         help="""
-        Number of times kubernetes API calls shoulbe retried before giving up
+        Total timeout, includign retry timeout, for kubernetes API calls
 
-        When a k8s API request connection times out, we retry it. This lets you
-        configure the number of times a request is retried before giving up.
+        When a k8s API request connection times out, we retry it while backing
+        off exponentially. This lets you configure the total amount of time
+        we will spend trying an API request - including retries - before
+        giving up.
         """
     )
 
@@ -1882,8 +1884,7 @@ class KubeSpawner(Spawner):
                 partial(self._make_create_pvc_request, pvc, self.k8s_api_request_timeout),
                 f'Could not create pod {self.pvc_name}',
                 # Each req should be given k8s_api_request_timeout seconds.
-                # FIXME: We should instead add a timeout_times property to exponential_backoff instead
-                timeout=self.k8s_api_request_timeout * self.k8s_api_request_timeout_retries
+                timeout=self.k8s_api_request_retry_timeout
             )
         # If we run into a 409 Conflict error, it means a pod with the
         # same name already exists. We stop it, wait for it to stop, and
@@ -1896,9 +1897,7 @@ class KubeSpawner(Spawner):
         yield exponential_backoff(
             partial(self._make_create_pod_request, pod, self.k8s_api_request_timeout),
             f'Could not create pod {self.pod_name}',
-            # Each req should be given k8s_api_request_timeout seconds.
-            # FIXME: We should instead add a timeout_times property to exponential_backoff instead
-            timeout=self.k8s_api_request_timeout * self.k8s_api_request_timeout_retries
+            timeout=self.k8s_api_request_retry_timeout
         )
 
         # we need a timeout here even though start itself has a timeout
@@ -1985,8 +1984,7 @@ class KubeSpawner(Spawner):
         yield exponential_backoff(
             partial(self._make_delete_pod_request, self.pod_name, delete_options, grace_seconds, self.k8s_api_request_timeout),
             f'Could not delete pod {self.pod_name}',
-            # FIXME: We should instead add a timeout_times property to exponential_backoff instead
-            timeout=self.k8s_api_request_timeout * self.k8s_api_request_timeout_retries
+            timeout=self.k8s_api_request_retry_timeout
         )
 
         try:
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -208,7 +208,7 @@ class KubeSpawner(Spawner):
     )
 
     k8s_api_request_timeout = Integer(
-        1,
+        3,
         config=True,
         help="""
         API request timeout (in seconds) for all k8s API calls.
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1853,7 +1853,7 @@ class KubeSpawner(Spawner):
                 self.log.info('Killed pod %s, will try starting singleuser pod again', self.pod_name)
             except ReadTimeoutError:
                 if i < (retry_times - 1):
-                    self.log.warn(f'create_namespaced_pod read timeout on attempt {i+1} of {retry_times}')
+                    self.log.warn(f'create_namespaced_pod for {self.user.name} timeout on attempt {i+1} of {retry_times}')
                 else:
                     raise
         else:
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1602,6 +1602,7 @@ class KubeSpawner(Spawner):
 
     @run_on_executor
     def asynchronize(self, method, *args, **kwargs):
+        self.log.info(f"Asynchronously calling {method} with {args} and {kwargs}")
         return method(*args, **kwargs)
 
     @property
@@ -1834,6 +1835,7 @@ class KubeSpawner(Spawner):
             pod = yield gen.maybe_future(self.modify_pod_hook(self, pod))
         for i in range(retry_times):
             try:
+                self.log.info(f"Attempting to create pod {pod['metadata']['name']}, with timeout {self.k8s_api_request_timeout}")
                 yield self.asynchronize(
                     self.api.create_namespaced_pod,
                     self.namespace,
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -225,7 +225,7 @@ class KubeSpawner(Spawner):
         30,
         config=True,
         help="""
-        Total timeout, includign retry timeout, for kubernetes API calls
+        Total timeout, including retry timeout, for kubernetes API calls
 
         When a k8s API request connection times out, we retry it while backing
         off exponentially. This lets you configure the total amount of time
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1784,6 +1784,40 @@ class KubeSpawner(Spawner):
     _last_event = None
 
     @gen.coroutine
+    def _make_create_pod_request(self, pod, request_timeout):
+        """
+        Make an HTTP request to create the given pod
+
+        Designed to be used with exponential_backoff, so returns
+        True / False on success / failure
+        """
+        try:
+            self.log.info(f"Attempting to create pod {pod.metadata.name}, with timeout {self.k8s_api_request_timeout}")
+            # Use tornado's timeout, _request_timeout seems unreliable?
+            yield gen.with_timeout(timedelta(seconds=request_timeout), self.asynchronize(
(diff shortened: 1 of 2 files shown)


yuvipanda:
This is great! I was *just* running into this issue, I think 👍 

We already use exponential backoff with timoutes in various other places - see https://github.com/jupyterhub/kubespawner/blob/b209ce169cec441169131bd0e0d6d43d1e3ca2df/kubespawner/spawner.py#L1851 for example. Can you modify this to use that pattern? That would be awesome <3

We already use the kubernetes package, so `urllib3` isn't a new dependency, right? Should be ok.

Thanks a lot for working on this!

yuvipanda:
I totally misunderstood the backoff comment you made, please ignore what I said 👍

I'm testing this in production now :)

kubespawner/spawner.py
```diff
@@ -34,6 +35,7 @@
 from jupyterhub.utils import exponential_backoff
 from jupyterhub.traitlets import Command
 from kubernetes.client.rest import ApiException
+from urllib3.exceptions import ReadTimeoutError
```
gravenimage, line 38: (Consider my comments a project-learning experience!) 

Should you specify urllib3 in the setup.py as well?  Of course it's going to come in as a transitive dep of 'kubernetes' - I don't know whether the policy would be to explicitly require it.

yuvipanda:
Since I'm running into this right now in production, I've been hacking away trying to get it fixed. You can see my work in my branch: https://github.com/yuvipanda/kubespawner/tree/timeout.

Are you running this in production right now, @gravenimage? 

Again, thank you very much for finding this issue and contributing :)

yuvipanda:
Basically, it looks like the AKS master we are using is very flaky - lots of ReadTimeouts everywhere. So we have to do the right thing and have timeouts / retries *everywhere* we talk to the k8s API. Our start / stop timeouts time out before the requests themselves time out, causing very weird & hard to debug issues.

In my branch I now have it for create and delete, since they both were being quite flaky. Needs a bit more of a systematic approach.

gravenimage:
I've been running my PR in a very low-use production environment for a few days.  I see a 5-second delay as expected quite a lot, but haven't had any failures so by that mark it is a success.  In the past I have also seen pod shutdown failures but could never get a reproduction,  and havent seen them recently.  I can well believe that AKS is flaky all over the place; at my work we have no real insight into the master plane :-(

gravenimage:
I'm on mobile at the moment so can't look at code very well,  but I'd thought of making a more failure-aware exponential-backoff ala Polly in .net where I could specify strategies for ReadTimeout without changing the logic of the rest of the code

yuvipanda:
I've added timeout and retry logic to pod / pvc creations / deletions. I ended up using tornado's timeout mechanism - the k8s API libraries was very iffy, and I couldn't quite figure out why. This also means we don't have to import urllib3 anymore, so one less implicit dependency!

Without this, we had *extremely* high spawn failure rates. This seems to have placated that a little, and I'm continuing to run this in production.

I hope this is useful to your production users too, @gravenimage. I am sorry I sort of took over your PR - I've sent you a private note about that too.

kubespawner/spawner.py
consideRatio: It appears that ReadTimeoutError is used in reflector.py still, but not in spawner.py. I personally think it should be an explicit dependency for kubespawner if we actually have it explicitly imported still.

> This also means we don't have to import urllib3 anymore, so one less implicit dependency!

```suggestion
```

kubespawner/spawner.py
```diff
@@ -1757,6 +1783,82 @@ def start(self):
...
     _last_event = None
 
+    @gen.coroutine
+    def _make_create_pod_request(self, pod, request_timeout):
+        """
+        Make an HTTP request to create the given pod
+
+        Designed to be used with exponential_backoff, so returns
+        True / False on success / failure
+        """
```
consideRatio, line 1793: The function also relies on raising errors to signal an outcome. I'm not confident on how JupyterHub which is using a KubeSpawner object throwing an error will react. @yuvipanda do you know?

To have this docstring or similar describe a bit about that would be relevant for me at least, I've asked this myself many times while inspecting this code base and failed to spot the associated JupyterHub try/except logic as I assume this wouldn't bring the hub down.

kubespawner/spawner.py
```diff
@@ -1757,6 +1783,82 @@ def start(self):
...
+                # We only want to handle 409 conflict errors
+                self.log.exception("Failed for %s", pod.to_str())
+                raise
+            self.log.info(f'Found existing pod {pod_name}, attempting to kill')
+            # TODO: this should show up in events
+            yield self.stop(True)
+
+            self.log.info(f'Killed pod {pod_name}, will try starting singleuser pod again')
+            # We tell exponential_backoff to retry
+            return False
```
consideRatio, line 1818: I found the following [list of k8s api-server response status codes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#http-status-codes) to be useful. For a _create resource_ request, the 409 status means the resource already exists.

I think this logic here is invalid or needs clarifications. Should we raise an error, return True / False, and/or take actions like deleting the pod etc in various situations?

kubespawner/spawner.py
```diff
@@ -1757,6 +1783,82 @@ def start(self):
...
+            self.log.info(f"Attempting to create pvc {pvc.metadata.name}, with timeout {request_timeout}")
+            yield gen.with_timeout(timedelta(seconds=request_timeout), self.asynchronize(
+                self.api.create_namespaced_persistent_volume_claim,
+                namespace=self.namespace,
+                body=pvc
+            ))
+            return True
+        except gen.TimeoutError:
+            # Just try again
+            return False
```
consideRatio, line 1840: Its a bit of pr-review-creep perhaps, but I feel like I want to mention that there are various others situations that will make sense to retry that will be listed as ApiException errors.

Here are those recommended to retry according to [this kubernetes community documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#http-status-codes).

429 StatusTooManyRequests
500 StatusInternalServerError
503 StatusServiceUnavailable
504 StatusServerTimeout

kubespawner/spawner.py
```diff
@@ -1880,32 +1941,55 @@ def _start(self):
...
+        else:
+            grace_seconds = self.delete_grace_period
+
+        delete_options.grace_period_seconds = grace_seconds
+
+
+        yield exponential_backoff(
+            partial(self._make_delete_pod_request, self.pod_name, delete_options, grace_seconds, self.k8s_api_request_timeout),
+            f'Could not delete pod {self.pod_name}',
+            # FIXME: We should instead add a timeout_times property to exponential_backoff instead
```
minrk, line 1989: FWIW, I don't agree that this `timeout_times` logic belongs in exponential_backoff. exponential_backoff is well defined for retrying something up to a time limit, letting unhandled errors raise. Handling a separate, shorter timeout as one of the cases that should be retried rightly belongs in the internal function without any knowledge of that in the outer exponential_backoff logic.

kubespawner/spawner.py
minrk: That's a great comment! I think it's okay to address this here if desired by the author, but also fine to improve the retry conditions in a later PR, since this one doesn't change the PVC retry logic, just reorganizes it.

kubespawner/spawner.py
```diff
@@ -1774,72 +1876,31 @@ def _start(self):
...
-            raise Exception(
-                'Can not create user pod %s already exists & could not be deleted' % self.pod_name)
+
+        # If there's a timeout, just let it propagate
+        yield exponential_backoff(
+            partial(self._make_create_pod_request, pod, self.k8s_api_request_timeout),
+            f'Could not create pod {self.pod_name}',
+            # Each req should be given k8s_api_request_timeout seconds.
+            # FIXME: We should instead add a timeout_times property to exponential_backoff instead
+            timeout=self.k8s_api_request_timeout * self.k8s_api_request_timeout_retries
```
minrk, line 1902: See above suggestion for making this timeout directly configurable instead of trying to calculate it from countable retries

kubespawner/spawner.py
minrk: Letting unhandled errors raise is normal error behavior, i.e. spawn failed with an error, resulting in logged exception and 500 error. Using exponential backoff has three cases:

1. success (return True/truthy)
2. handled failure that should be retried (return False)
3. unhandled error (let it raise, will propagate up to the caller)

I don't think this PR changes the outcomes of different errors other than these short timeouts.

kubespawner/spawner.py
minrk: This logic is unchanged from before - if a pod is already created (i.e. not by us!), delete it and recreate it.

The very-short timeouts in this PR, though, make it a bit more likely that we:

1. create it successfully, but timeout is hit
2. next attempt fails with 409, but success was ours!
3. delete our first creation and try again

kubespawner/spawner.py
```diff
@@ -207,6 +208,31 @@ def __init__(self, *args, **kwargs):
...
+        This is the total amount of time a request might take before the connection
+        is killed. This includes connection time and reading the response.
+
+        NOTE: This is currently only implemented for creation and deletion of pods,
+        and creation of PVCs.
+        """
+    )
+
+    k8s_api_request_timeout_retries = Integer(
+        5,
```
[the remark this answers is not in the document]
yuvipanda: Yeah, I agree - both on having a timeout here and on the naming. Will amend.

kubespawner/spawner.py
yuvipanda: Yeah, I think setting an overall number_of_tries is the way to go.

kubespawner/spawner.py
consideRatio: I opened https://github.com/jupyterhub/kubespawner/issues/436 to represent this matter.

kubespawner/spawner.py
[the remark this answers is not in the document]
yuvipanda: Done!

kubespawner/spawner.py
yuvipanda: I opened https://github.com/jupyterhub/kubespawner/pull/437 for that

yuvipanda:
Thanks for the comments, @consideRatio @minrk. I've made the changes suggested I think. I agree the retry and error handling logic could be improved further, but hope that can be made a separate PR to keep this one small.

kubespawner/spawner.py
minrk: Follow-up to clarify the last point: this case where we created the pod, API call raised but creation actually happened, and then we delete & retry was already technically a possibility that could occur. This PR adds one more situation where it can happen that can be triggered by load on *either* kubernetes or the Hub.

minrk (approved):

minrk:
Awesome, thank you!

kubespawner/spawner.py
```diff
@@ -1774,72 +1877,28 @@ def _start(self):
...
-
-                    self.log.info("PVC " + self.pvc_name + " already exists, possibly have reached quota though.")
-
-                else:
-                    raise
-
+            # If there's a timeout, just let it propagate
+            yield exponential_backoff(
+                partial(self._make_create_pvc_request, pvc, self.k8s_api_request_timeout),
+                f'Could not create pod {self.pvc_name}',
```
mriedem, line 1885: Should be something like "Could not create PVC" right? Copy/paste error?

consideRatio: Ah! Good catch!

PR #438, opened 2020-09-17
Allow pod to spawn if the PVC specified already exists

gravenimage:
I think this is just fallout from the recent timeout PR, if I understand correctly.


yuvipanda:
Yep, great catch! Thanks, @gravenimage!

Issue #440, opened 2020-09-29
KeyError: 'containerStatuses' in poll()


mriedem:
## Bug description
During an unplanned load event in our hub some pods were timing out and I noticed a KeyError in the logs:

```
Sep 29 15:12:00hub-6788d79c4-mct9jhubERRORERROR 2020-09-29T20:12:00.742Z [JupyterHub user:645] Unhandled error starting 5f5118929d1395001a6aa4df's server: pod/jupyter-5f5118929d1395001a6aa4df did not start in 300 seconds!
Sep 29 15:12:00hub-6788d79c4-mct9jhubERRORERROR 2020-09-29T20:12:00.778Z [JupyterHub user:657] Failed to cleanup 5f5118929d1395001a6aa4df's server that failed to start
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 651, in spawn
        await self.stop(spawner.name)
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 748, in stop
        status = await spawner.poll()
      File "/usr/local/lib/python3.6/dist-packages/kubespawner/spawner.py", line 1572, in poll
        ctr_stat = data["status"]["containerStatuses"]
    KeyError: 'containerStatuses'
    
Sep 29 15:12:00hub-6788d79c4-mct9jhubERRORERROR 2020-09-29T20:12:00.779Z [JupyterHub gen:599] Exception in Future <Task finished coro=<BaseHandler.spawn_single_user.<locals>.finish_user_spawn() done, defined at /usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py:845> exception=TimeoutError('pod/jupyter-5f5118929d1395001a6aa4df did not start in 300 seconds!',)> after timeout
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 593, in error_callback
        future.result()
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py", line 852, in finish_user_spawn
        await spawn_future
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 661, in spawn
        raise e
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/user.py", line 560, in spawn
        url = await gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
[...]

PR #441, opened 2020-10-01
Access containerStatuses key with get()

rmoe:
There are some scenarios in which a pod will have a status
value but not a containerStatuses value. This change prevents
a KeyError in that case.

Closes: #440


yuvipanda:
Thanks for the quick fix, @rmoe!

Issue #221, opened 2018-07-31
Rename setting: `image_pull_secrets` to `image_pull_secret_name`


consideRatio:
I think it would be an improvement to have `image_pull_secrets` renamed to `image_pull_secret_name`, but it probably isn't worth our time to deprecate and introduce the new name, or write this issue... Well now that I have done it eh...

---

https://github.com/jupyterhub/kubespawner/blob/27056a79580df957d0cc2d7b61ecfa4ce694e4fd/kubespawner/spawner.py#L417-L431

---

A Kubernetes Pod can look like...

```yaml
spec:
  imagePullSecrets: [{'name': 'name-of-k8s-secret-resource-1'}, {'name': 'name-of-k8s-secret-resource-2'}]
```

But KubeSpawner can only be configured to add one single secret in another format.

What the user sets `KubeSpawner.image_pull_secrets` to will passed to the `make_pod` function...

https://github.com/jupyterhub/kubespawner/blob/27056a79580df957d0cc2d7b61ecfa4ce694e4fd/kubespawner/spawner.py#L1185-L1191

... and in it will set the pod specification according to the Kubernetes API like this ...

https://github.com/jupyterhub/kubespawner/blob/27056a79580df957d0cc2d7b61ecfa4ce694e4fd/kubespawner/objects.py#L173-L177

PR #442, opened 2020-10-04
Allow image_pull_secrets config to be specified the k8s native way

consideRatio:
## Motivation

I'm working on [a feature in jupyterhub/zero-to-jupyterhub-k8s](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1794) to configure all
the Helm chart's pods' `spec.imagePullSecrets` configurations at once while also
being able to configure it for individual pods. The chart global configuration
should be allowed to combine with the pod specific configuration, which require
KubeSpawner to support the Kubernetes matching variant where a list of strings
rather than a single string can be provided.

To conclude, `c.KubeSpawner.image_pull_secrets` passed a string will get a soft
deprecation, and passing a list of either strings or dicts like `{"name": "string"}` is
the new recommended syntax, the latter represent the k8s native syntax.

## Overview

- [x] Add support for KubeSpawner configuration `image_pull_secrets` to accept a
  list of Kubernetes Secret references. Either as strings with Secret names, or
  dicts like `{"name": "secret-name"}`, where the latter mimics the k8s
  configuration available in a Pod.
- [x] Add a _soft deprecation_ of passing a string to `image_pull_secrets`,
  providing a warning while automatically adjusting the set value to a single
  entry in a list in the k8s native format.
  - [x] Add a test of the soft deprecation's indirect configuration.
- [x] Make a **breaking change** where `make_pod` in objects.py no longer can accept
  `image_pull_secret` (singular), and instead expects `image_pull_secrets`
  (plural), that assumes a list is passed.

## Related

- https://github.com/jupyterhub/kubespawner/issues/221 - An old discussion about KubeSpawner's k8s mismatching image_pull_secrets configuration.
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1794 - The z2jh PR that depends on this PR.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -28,7 +28,7 @@ def make_pod(
     port,
     image,
     image_pull_policy,
-    image_pull_secret=None,
+    image_pull_secrets=None,
     node_selector=None,
     run_as_uid=None,
     run_as_gid=None,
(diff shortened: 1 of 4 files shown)


yuvipanda:
Thanks a lot, @consideRatio

consideRatio:
That feeling of waking up and seeing notifications and the purple color of PRs ❤️❤️❤️!!

Thank you for so quickly helping out with review work @yuvipanda 🥳🥰

PR #444, opened 2020-10-04
Python 3.6+ migration: async in 3.5 and async with yeild in 3.6

consideRatio:
Since KubeSpawner already require Python 3.6+ according to setup.py, and async support was introduced in Python 3.5 and augmented for generators using yeild in Python 3.6, this is plausible to do at this point.

Review greatly appreciated!

## Notes
- [x] @gen.coroutine is replaced by appending `async` to the function definition
- [x] At times where `yield` is used not to return a value, but as something to wait for in the old system with `@gen.coroutine`, then I've replaced it with `await`.
- [x] At times where `gen.yield_` is used to actually return a value as a async generator, then I've replaced it with `yield`
- [x] I've made use of `async.ensure_future(...)` on functions where situations like `my_resp = my_async_func()` occur, in order to ensure the `my_resp` becomes a [`Task` object](https://docs.python.org/3/library/asyncio-task.html#task-object) which behaves like a [`Future` object](https://www.tornadoweb.org/en/stable/concurrent.html#tornado.concurrent.Future) which has the functions `.done()` and `.result()` for example, while the response of a `my_async_func()` is a [`coroutine` object](https://docs.python.org/3/glossary.html#term-coroutine) that won't have `.done` etc.

## Review questions
- [ ] Does it make sense that the `start()` function isn't async, but the `stop()` function is?

### DIFF
--- a/kubespawner/proxy.py
+++ b/kubespawner/proxy.py
@@ -106,18 +106,16 @@ class KubeIngressProxy(Proxy):
         )
         return safe_name
 
-    @gen.coroutine
-    def delete_if_exists(self, kind, safe_name, future):
+    async def delete_if_exists(self, kind, safe_name, future):
         try:
-            yield future
+            await future
             self.log.info('Deleted %s/%s', kind, safe_name)
         except client.rest.ApiException as e:
             if e.status != 404:
                 raise
             self.log.warn("Could not delete %s/%s: does not exist", kind, safe_name)
 
-    @gen.coroutine
-    def add_route(self, routespec, target, data):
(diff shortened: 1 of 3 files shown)


consideRatio:
Not sure who has experience with async stuff @manics, perhaps you do and some time to review this PR? :bow:

yuvipanda:
This actually looks pretty good to me, @consideRatio!

`start` not being async is fine - we return a Future from it (via `_start`) so in practice it behaves exactly same as an `async def` function.

PR #451, opened 2020-10-22
KubeSpawner.image_pull_secrets malfunctions in 0.14.0 - this fixes it

johnhoman:
The existing KubeSpawner.image_pull_secrets doesn't seem to be working. The validation method returns `None` when the input type is not a string e.g. a list causes errors. The existing help text claims that a list is the supported type but passing a list will result in image_pull_secrets being `NoneType` . From the traitlets docs `If this function does not have any return statement, then the returned value will be None, instead of what we wanted (which is proposal['value']).` https://traitlets.readthedocs.io/en/stable/using_traitlets.html#validation-and-coercion

I made a `help` change to `KubeSpawner.image_pull_secrets` to constrain the input type to a list of strings. I'm not sure if this is the right thing. The existing help section claims that is should be `A list of references to Kubernetes Secret resources with credentials to pull images from image registries`. This is somewhat unclear so I'm not sure how to correct this. Was the intention for this to be a list of `V1LocalObjectReference`?

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -547,6 +547,8 @@ class KubeSpawner(Spawner):
             )
             return [{"name": proposal['value']}]
 
+        return proposal['value']
+
     node_selector = Dict(
         config=True,
         help="""
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -330,3 +330,21 @@ def test_pod_name_collision():
     assert spawner.pod_name != named_spawner.pod_name
     assert named_spawner.pvc_name == "claim-user-2dhas--2ddash"
     assert spawner.pvc_name != named_spawner.pvc_name
+
+
+def test_spawner_can_use_list_of_image_pull_secrets():
+    secrets = ["ecr", "regcred", "artifactory"]
+
+    c = Config()
+    c.KubeSpawner.image_spec = "private.docker.registry/jupyter:1.2.3"
+    c.KubeSpawner.image_pull_secrets = secrets
+    spawner = KubeSpawner(hub=Hub(), config=c, _mock=True)
+    assert spawner.image_pull_secrets == secrets
+
+    secrets = [dict(name=secret) for secret in secrets]
+    c = Config()
+    c.KubeSpawner.image_spec = "private.docker.registry/jupyter:1.2.3"
+    c.KubeSpawner.image_pull_secrets = secrets
+    spawner = KubeSpawner(hub=Hub(), config=c, _mock=True)
+    assert spawner.image_pull_secrets == secrets
+


consideRatio:
@johnhoman thank you for investigating this thoroughly and describing it clearly! I assume you are fully correct about this being a bug, and you have found the issue. I'm the culprit and is confused this wasn't caught 

We have in the past chosen to not mirror the k8s definitions, but come up with our own, or mimic the JupyterHub Spawner configurations etc. At this point in time, I think we should make sure to support what mirrors k8s conventions exactly, while perhaps also being backward compatible.

For this configuration of `image_pull_secrets`, I suggest that we support:
1. The deprecated syntax of `"a single string"`
2. The convinience syntax of `["a list of strings"]`
3. And the k8s native syntax of `"[{'name': 'a list of objects with a name key with a string value'}"`

I think at this point, let's go with supporting all of them to not need to think about breaking backward compatibility.

It was me who broke this configuration, because I note that I only tested the logic of `make_pod`, and not the configuration logic. I suggest that we let the validation only emit a warning, but not modify the value at all, but instead let the make_pod logic do that on its end.

https://github.com/jupyterhub/kubespawner/blob/147f2603cabf89a4930b506318824283f5d36ba8/kubespawner/objects.py#L235-L244

This means that I think this PR can settle for or a one-liner change, which is to update the validation function to always just return the proposed value without changing it, independent if it is a string or list of strings or list of objects with a name key.

johnhoman:
> @johnhoman thank you for investigating this thoroughly and describing it clearly! I assume you are fully correct about this being a bug, and you have found the issue. I'm the culprit and is confused this wasn't caught
> 
> We have in the past chosen to not mirror the k8s definitions, but come up with our own, or mimic the JupyterHub Spawner configurations etc. At this point in time, I think we should make sure to support what mirrors k8s conventions exactly, while perhaps also being backward compatible.
> 
> For this configuration of `image_pull_secrets`, I suggest that we support:
> 
> 1. The deprecated syntax of `"a single string"`
> 2. The convinience syntax of `["a list of strings"]`
> 3. And the k8s native syntax of `"[{'name': 'a list of objects with a name key with a string value'}"`
> 
> I think at this point, let's go with supporting all of them to not need to think about breaking backward compatibility. It was me who broke this configuration in a recent PR in #442, because I note that I only tested the logic of `make_pod`, and not the configuration logic.
> 
> https://github.com/jupyterhub/kubespawner/blob/147f2603cabf89a4930b506318824283f5d36ba8/kubespawner/objects.py#L235-L244
> 
> This means that I think this PR can settle for or a one-liner change, which is to update the validation function to just add a return value with the proposed value without changing it unless it is a string, independent if it is a string or list of strings or list of objects with a name key. That would resolve your issue.

Thanks for the review! If I understood you correctly then I made the appropriate changes. Please let me know if I missed something.

consideRatio:
I think this is spot on! Thank you for your work to investigate and fix the mistake I introduced :) :tada: :heart:

As seen on https://www.traviscistatus.com/, it is under heavy load and has been the last week so the builds are lagging behind. But, let's get this merged when it succeeds, and then I want to also create a 0.14.1 release of kubespawner fixing this.

consideRatio (approved):
LGTM! Thank you @johnhoman :tada: :heart:

consideRatio:
I some some tests succeed, so then it must be good for all k8s versions. Thanks!

Issue #307, opened 2019-03-10
matrix of profile_list options


rabernat:
I would like to give users an option to select _independently_ between machine options (e.g. `cpu_limit`) and notebook image.

I imagine that this may be possible already, either through some `profile_form_template` magic or by supplying a callable as profile_list. But I thought I would ask for suggestions here before diving too deep into either of these. Perhaps the devs already have a good idea about the easiest way to accomplish this.

consideRatio:
Ah so like a set of independent choices that would then be merged together?

- Current situation: we are using N sets of predefined options.
- Your Matrix suggestion  as I understand it: is having sets of sets of options that then is merged.
- Related to this: allowing for multiple choice for individual options within an item within a profile_list set of options.

rabernat:
Yes exactly.

Symbolically, I would specify the profile list like something like this
```python
{'memory_options': [
    {
        'display_name': 'Low Memory',
        'kubespawner_override': {
            'mem_limit': '512M',
        }
    }, {
        'display_name': 'High Memory',
        'kubespawner_override': {
            'mem_limit': '2G',
        }
    } ]
 'image_options' [
    {
        'display_name': 'Image A',
        'kubespawner_override': {
            'image': 'training/datascience:A',
        }
    }, {
        'display_name': 'Image B',
        'kubespawner_override': {
            'image': 'training/datascience:B',
        }
    } ]
}
```

I would like to see two different multiple-choice areas on the spawner form: one for `memory_options`, one for `image_options`.

rabernat:
If someone can give me some advice about how to get started, I could consider starting a PR to implement this. I think we would want to discuss and agree on the needed data structures first.

keniseli:
I'd love to help with the implementation. Is the desired outcome already clear enough?

rabernat:
Hi @consideRatio -- @keniseli is a new contributor to Pangeo and we identified this issue as a good place to start.

It would be great if you or some of the other devs could give him some pointers on how to get started.

consideRatio:
Thanks for the ping @rabernat and welcome @keniseli! I think you are asking the perfect question right away =)

Some steps that I think makes sense to take initially:
- Clarify a proposals impact for the end user experience
- Clarify a proposals way to configure Kubespawner
- Try get a local development environment up and running according to the docs here, or perhaps the docs in jupyterhub/zero-to-jupyterhub-k8s that has modern docs on setting up at k8s cluster with some tools that was not around when this docs were written.

[...]

PR #452, opened 2020-10-23
Add options_from_form as configurable

cbanek:
Without this part as being configurable, it doesn't make sense to
be able to configure the options_form part since you can't easily
get the values back to permute them into options the kubespawner
will understand.  I think this meant to be decorated since it is mentioned
in the documentation in the options_form section.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2035,7 +2035,8 @@ class KubeSpawner(Spawner):
         else:
             return self._render_options_form(self.profile_list)
 
-    def options_from_form(self, formdata):
+    @default('options_from_form')
+    def _options_from_form(self, formdata):
         """get the option selected by the user on the form
 
         This only constructs the user_options dict,


consideRatio:
KubeSpawner is a class derived from Spawner, and I understand @default to be a decorator for traitlets only.

Inspecting the base class I see that [Spawner is defining options_from_form as a function](https://github.com/jupyterhub/jupyterhub/blob/082f6516a15de1c74657a57b5f9fee5a082c0600/jupyterhub/spawner.py#L357) rather than as a traitlet of type `Callable()`, like [options_form is defined for example](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L311).

Due to that, it doesn't seem to make sense to me to decorate the function with @default, as there is no such traitlet defined, and I suspect the kind of change you are looking for should be done in JupyterHub rather than KubeSpawner.

---

If you want to configure the options_from_form function, you are seemingly currently required so create a custom class that derives from KubeSpawner and override it. And, if there is a usecase for that, then it would be better to make it configurable I guess, which would be something done in the Spawner base class.

cbanek:
Sounds good - thanks for the double check.

To back up about a million miles, I like the profile part of the kubespawner, but I want to have two independent sets of boxes on the options form: one for image, and one for size (cpu/memory).  So right now I could do this with a cross product and make every possible permutation, but our list of images is really pretty large, so it'd make the page cumbersome.  So right now, I need to be able to look at the returned form data, since the only data that is saved (and defaulted) is profile.

So it sounds like there's two options:
1. Make changes to spawner in jupyterhub and have this function be configurable (which is partially here, but mostly not)
2. Deriving from KubeSpawner and overriding the functions.


I'm very easily confused by the traitlets part I will admit, although  when adding the annotation here, I was able to get the function override to work (otherwise I'd get a warning about it not being configurable).  In general, I'm not sure when I should be making things configurable versus the base class, although being able to inject just a function over inheritance seems good (composition over inheritance)?

Thanks much for the help!

consideRatio:
@cbanek to make KubeSpawner's profile_list give you two independent kinds of choices is not possible at the moment, and I think it is related to discussion in #307.

Btw, regarding traitlets, if you want to invest some time to learn more about them I can recommend https://www.youtube.com/watch?v=_gYEVTaNuKU - for me, it has been the most complicated tech part of the Jupyter ecosystem to understand, so I became very happy to get to listen to this talk.

Regarding overriding or configuring, I think you should aim to configure using traitlets if possible, and override as a last resort. Overriding functions etc will expose you to far more breaking changes when you update versions of the base class you are overriding functions in as compared to if you configured it someway using its mechanisms intended to be configured.

---

About the options_from_form - I'm generally confused about these functions, I don't know for sure what they do actually. Looking it up in JupyterHub's documentation for Spawner's I find [this](https://jupyterhub.readthedocs.io/en/latest/api/spawner.html#jupyterhub.spawner.Spawner.options_from_form). They explicitly suggest it can make sense to override. I guess it will need to be very tightly coupled with whatever goes on in the [options_form](https://jupyterhub.readthedocs.io/en/latest/api/spawner.html#jupyterhub.spawner.LocalProcessSpawner.options_form) HTML.

It sounds to me, that you are looking to do something beyond KubeSpawner, which may motivate some change in the default Spawner base class, but I'm not sure. But I want to avoid making a change to KubeSpawner itself unless it is relevant specifically to KubeSpawner. Are you okay to close this PR as a change suggestion not verified to make enough sense for the KubeSpawner project?

cbanek:
Sure no worries, thanks for the help and the links on traitlets, I'll be sure to give that a watch.  As if python isn't sometimes confusing enough. :)

For the spawner page, yes this is where I started, and it looked like it should be override-able with the trailetls part, but I guess since it doesn't have the python recipe next to it (and it's below all the others) that it isn't.  I also feel like the options_form and the options_from_form are tightly coupled, but in the base class it basically just returns all the form data, whereas the kubespawner limits the returned data to what it was expecting.  If the kubespawner options_from_form defaulted profile and returned everything else in the formdata, I think that would also possibly work?

consideRatio:
> If the kubespawner options_from_form defaulted profile and returned everything else in the formdata, I think that would also possibly work?

Hmmmm, if you don't generate your own options_form HTML, but instead use profile_list which is a KubeSpawner feature to generate the options_form HTML (and perhaps also options_from_form?), then you will need to manually create a "profile" for each permutation of combinations because the user will only be able to choose one of the profiles with the generated options_form HTML when using profile_list.

It may be relevant for you to inspect the following [configuration](https://github.com/neurohackademy/nh2020-jupyterhub/blob/da35049a93d372fe501b7968687e703a5e048e30/deployments/hub-neurohackademy-org/config/prod.yaml#L213-L289) I did of KubeSpawner before for a deployment. I put some notes which I think can help you piece together some parts of how things work better.

PR #453, opened 2020-10-24
Let uid/gid/fs_gid default to None instead of 0

consideRatio:
This PR will introduce a change for those that use KubeSpawner without configuring .gid, .uid, or .fs_gid. This is because those fields then have default to the value of 0 instead of None, which makes them become set to zero instead of simply not be set.﻿

I think it is sensible to introduce this breaking change, which only will be breaking for those that use the KubeSpawner default for these settings, for two reasons.

1. The KubeSpawner default values are not secure defaults - granting root access by default doesn't seem to make much sense to me.
2. The most sensible default in my mind is to not override the Docker image configuration, no matter if it is secure or not.

## Review notes
- Union traitlets must have an explicit default_value keyword argument passed, we cannot use a positional argument to set the default value.
- I updated the tests for uid/gid/fs_gid to ensure that it does the right thing when it is set to 0 rather than None correctly.

## Breaking changes
- In standalone use of KubeSpawner, the defaults change, which may break users assumptions.
- In z2jh we don't explicitly set `gid` on KubeSpawner, which means that while we previously in Z2JH got `c.KubeSpawner.gid=0` as default, we now get `c.KubeSpawner.gid=None` as default. Users may have relied on `gid=0`.

---

Closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1733.
Closes #453.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -567,6 +567,7 @@ class KubeSpawner(Spawner):
             Integer(),
             Callable(),
         ],
+        default_value=None,
         allow_none=True,
         config=True,
         help="""
@@ -592,6 +593,7 @@ class KubeSpawner(Spawner):
             Integer(),
             Callable(),
         ],
+        default_value=None,
         allow_none=True,
         config=True,
         help="""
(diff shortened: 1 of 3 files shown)


yuvipanda:
I like this! I think it defaulting to 0 is definitely unintended behavior, and this is a good fix for it.

consideRatio:
@yuvipanda thank you for the quick review and merge!

PR #455, opened 2020-10-25
Cleanup JS patch of JupyterHub 0.8 HTML not needed in 0.9+

### LINKED ISSUE #225: profile_form_template - code clean up when we stop supporting JupyterHub 0.8
From `spawner.py`:

```python
    profile_form_template = Unicode(
        default_value="""
        <script>
        // JupyterHub 0.8 applied form-control indisciminately to all form elements.
        // Can be removed once we stop supporting JupyterHub 0.8
        $(document).ready(function() {
            $('#kubespawner-profiles-list input[type="radio"]').removeClass('form-control');
        });
        </script>
        <style>
        /* The profile description should not be bold, even though it is inside the <label> tag */
        #kubespawner-profiles-list label p {
            font-weight: normal;
        }
        </style>

        <div class='form-group' id='kubespawner-profiles-list'>
        {% for profile in profile_list %}
        <label for='profile-item-{{ loop.index0 }}' class='form-control input-group'>
            <div class='col-md-1'>
                <input type='radio' name='profile' id='profile-item-{{ loop.index0 }}' value='{{ loop.index0 }}' {% if profile.default %}checked{% endif %} />
            </div>
            <div class='col-md-11'>
                <strong>{{ profile.display_name }}</strong>
                {% if profile.description %}
                <p>{{ profile.description }}</p>
                {% endif %}
            </div>
        </label>
        {% endfor %}
        </div>
        """,
        help="""
        Jinja2 template for constructing profile list shown to user.

        Used when `profile_list` is set.

        The contents of `profile_list` are passed in to the template.
        This should be used to construct the contents of a HTML form. When
        posted, this form is expected to have an item with name `profile` and
        the value the index of the profile in `profile_list`.
        """
    ).tag(config=True)
```

consideRatio:
Closes #225 which was a reminder to do this cleanup.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1133,13 +1133,6 @@ class KubeSpawner(Spawner):
 
     profile_form_template = Unicode(
         """
-        <script>
-        // JupyterHub 0.8 applied form-control indisciminately to all form elements.
-        // Can be removed once we stop supporting JupyterHub 0.8
-        $(document).ready(function() {
-            $('#kubespawner-profiles-list input[type="radio"]').removeClass('form-control');
-        });
-        </script>
         <style>
         /* The profile description should not be bold, even though it is inside the <label> tag */
         #kubespawner-profiles-list label p {


yuvipanda:
w00p w00p, thanks @consideRatio!

PR #458, opened 2020-11-12
Add user namespace support

athornton:
I'm going to update the initial description; the original is preserved at the bottom.

This PR adds the ability to spawn user pods into their own namespaces.  This is done with a new Configurable, `enable_user_namespaces`.  A sane default is provided for individual user namespaces, `{hubnamespace}-{username}`, but this too is a Configurable.  There are also corresponding test cases for user-namespace spawning.

If you are using this feature, the Hub will need ClusterRoles analogous to the Roles it already has, since it needs to watch, create, and destroy objects in namespaces that are not its own.  That will require a PR to z2jh, which has not been written yet.

Because reflecting across multiple namespaces requires changes to the watch method names and method arity for the Reflector classes, the Reflectors have been refactored to allow this and avoid unnecessary duplication.  There is now just a ResourceReflector, for which NamespacedResourceReflector and MultiNamespaceResourceReflector are thin wrappers, and behavior is gated by the omit_namespace Configurable.

There are some whitespace changes, in some cases induced by my editor, and in others to provide consistent formatting of our docstrings.  Many docstrings have been updated to point to current (1.20) K8s documentation, as the referenced 1.11 documentation is no longer online.

A future PR will format everything with Black and add a pre-commit hook to enforce formatting, which ought to keep this problem from recurring.

At the moment, cleaning up a namespace on pod exit is the responsibility of the user (by replacing the post_stop hook in the spawner).  A future PR will introduce a way to set this as a Boolean or Callable.  There is not a one-size-fits-all model for whether to delete the namespace on pod stop or not--if one is using pre-created namespaces, you wouldn't want to do this, and in my own use-case it's possible that the user wants to stop the interactive pod but leave batch workload running within the same namespace.

[original description follows]

Apologies about the editor-induced whitespace changes.

This one passes the unit tests and works in z2jh modulo changes to the configmap to use the new subclass and change the internal Hub URL to be hub.`<namespace>`, and an appropriate ClusterRole/CRBinding to give the Hub power to do cluster operations.

[end original description]


meeseeksmachine:

https://discourse.jupyter.org/t/500-failed-to-connect-to-my-hub-at-http-hub-8081-hub-hub-api/6785/2

athornton:
What do I need to do to move this forward?

kubespawner/spawner.py
```diff
@@ -2158,3 +2310,22 @@ async def load_user_options(self):
...
+                self.log.exception("Failed to create namespace %s",
+                                   self.namespace)
+                raise
+
+
+class MultiNamespaceKubeSpawner(KubeSpawner):
+    """
+    MultiNamespaceKubeSpawner is the all-namespaces analog of KubeSpawner.
+    """
+    enable_user_namespaces = True
```
minrk, line 2331: By having `enable_user_namespaces` be a configurable trait, as you've defined above, I think we can skip the definition of this additional class and refer folks to:

```python
c.KubeSpawner.enable_user_namespaces = True
```

kubespawner/spawner.py
```diff
@@ -2158,3 +2310,22 @@ async def load_user_options(self):
                     )
                 )
             )
+
+    async def _ensure_namespace(self):
+        ns = make_namespace(self.namespace)
+        api = self.api
+        try:
+            api.create_namespace(ns)
```
minrk, line 2318: let's keep our API requests async:

```suggestion
            await gen.with_timeout(
                timedelta(seconds=self.k8s_api_request_timeout),
                self.asynchronize(api.create_namespace, ns),
            )
```

kubespawner/__init__.py
```diff
@@ -11,8 +11,11 @@
 # We export KubeSpawner specifically here. This simplifies import for users.
 # Users can simply import kubespawner.KubeSpawner in their applications
 # instead of the more verbose import kubespawner.spawner.KubeSpawner.
+#
+# We're also exporting the convenience class MultiNamespaceKubeSpawner,
+# for the same reason.
 
-from kubespawner.spawner import KubeSpawner
+from kubespawner.spawner import KubeSpawner, MultiNamespaceKubeSpawner
```
minrk, line 18: Given how MultiNamespaceKubeSpawner is implemented, I think we can lose the additional class and instead enable this behavior with:

```python
c.KubeSpawner.enable_user_namespaces = True
```

kubespawner/objects.py
```diff
@@ -169,31 +171,35 @@ def make_pod(
         https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/.
 
         Pass this field an array of "Toleration" objects.*
-        * https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#nodeselectorterm-v1-core
+        #nodeselectorterm-v1-core
```
minrk, line 174: Looks like whatever auto-wrapped these lines mangled the URLs :/

kubespawner/reflector.py
```diff
@@ -152,16 +184,26 @@ def _list_and_update(self):
...
-        initial_resources = getattr(self.api, self.list_method_name)(
-            self.namespace,
-            label_selector=self.label_selector,
-            field_selector=self.field_selector,
-            _request_timeout=self.request_timeout,
-            _preload_content=False,
-        )
+        initial_resources = None
+        if self.omit_namespace:
+            initial_resources = getattr(self.api, self.list_method_name)(
```
minrk, line 189: Can probably avoid this duplication with something like:

```python
kwargs = dict(
    _preload_content=False,
)
if not self.omit_namespace:

initial_resources = getattr(self.api, self.list_method_name)(**kwargs)
```

which I see you are already doing below with `watch_args`

kubespawner/spawner.py
```diff
@@ -71,6 +73,22 @@ def pods(self):
         return self.resources
 
 
+class MultiNamespacePodReflector(MultiNamespaceResourceReflector):
```
minrk, line 76: Since the Multi base class is really one line, I think this might be simpler with:

```python
class MultiNamespacePodReflector(PodReflector):
    omit_namespace = True
    list_method_name = 'list_pod_for_all_namespaces'
```

and avoid duplication of the PodReflector.

Same goes for the other cross-namespace reflectors

kubespawner/spawner.py
```diff
@@ -107,9 +125,27 @@ def events(self):
...
 
 
+class MultiNamespaceEventReflector(MultiNamespaceResourceReflector):
+    """
+    MultiNamespaceEventReflector is the all-namespaces analog of
+    EventReflector.
+    """
+
+    kind = "events"
+    list_method_name = "list_event_for_all_namespaces"
```
minrk, line 135: In general, what do you think about refactoring the reflectors to use `kind` to construct `list_method_name`. Do these all have the form `list_{kind}_for_all_namespaces` or `list_namespaced_{kind}` (if kind were singular, which it is not currently). Then we could just have two parameters: `kind` and `omit_namespace` to configure most values?

kubespawner/spawner.py
```diff
@@ -244,14 +292,55 @@ def __init__(self, *args, **kwargs):
...
+        This comes with some caveats.  The Hub must run with significantly
+        more privilege (must have ClusterRoles analogous to its usual Roles)
+        and can therefore do heinous things to the entire cluster.
+
+        It will also make the Reflectors aware of pods and events across
+        all namespaces.  This will have performance implications, although
+        using labels to restrict resource selection helps somewhat.
+
+        If you use this, consider cleaning up the user namespace in your
+        post_stop_hook.
```
minrk, line 310: Should cleaning up the user namespace be something we do by default?

kubespawner/spawner.py
```diff
@@ -645,7 +734,8 @@ def _validate_image_pull_secrets(self, proposal):
         upgrades to break.
 
         You'll *have* to set this if you are using auto-provisioned volumes with most
-        cloud providers. See `fsGroup <https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#podsecuritycontext-v1-core>`_
+        #podsecuritycontext-v1-core>`_
```
minrk, line 737: Same comment about the broken links. Looks like an auto-line-wrapper doesn't respect docstring contents.

kubespawner/spawner.py
```diff
@@ -1773,9 +1892,12 @@ def _start_watching_events(self, replace=False):
         If replace=True, a running pod reflector will be stopped
         and a new one started (for recovering from possible errors).
         """
+        event_reflector_class = EventReflector
+        if self.enable_user_namespaces:
+            event_reflector_class = MultiNamespaceEventReflector
```
minrk, line 1897: Depending on the resolution of the above discussion, this may be a single class adding `kwargs["omit_namespace"] = True`

minrk (commented):
This is great! Thanks for sticking with this for so long. I left some comments about how to implement the multiple behaviors, but I think the main issue to resolve is whether namespace cleanup belongs in the default implementation and on-by-default (unconditional?).

kubespawner/spawner.py
athornton: OK.  This is clearly a matter of taste, but I'm happy to go with that.

kubespawner/spawner.py
athornton: Oooh, I like that.

kubespawner/spawner.py
athornton: I don't think so.  One of the early comments here wanted to do this with pre-created namespaces.  The existing code here works fine with that, because if the namespace is extant, you get a 409 back and keep going.

But if you have pre-created namespaces, then presumably you want them to stick around in the absence of a user's running pod.

There are also some subtleties to the teardown: I can imagine a case in which the user wants to stop their running pod but have the namespace sticking around doing batch work (we're currently using Argo Workflows for something like this).  Thus I think it's better to tell the user where they can clean it up (or I could add another configurable).

kubespawner/spawner.py
athornton: I'll patch these up as I implement the other suggestions.

kubespawner/spawner.py
minrk: Makes sense. thanks! I think it will be worth making it a bool configurable rather than requiring generic logic in a hook, but that can be its own PR.

athornton:
Working on rebasing against current master; should have something working relatively soon.

athornton:
OK.  We're back to the tests passing, so I will start with your suggestions.

I ran `black -l 79 .` over it because I got tired of fighting with the editor and the existing code; hopefully that will satisfy everyone, but if not, it is a separate commit, so we can cherry-pick it out if you feel strongly about it.

athornton:
So, at this point it's done except for the reflector refactor, which I hope to have in a state to review by tomorrow.

athornton:
I think it is now ready for review.

athornton:
Actually, let me do a bit more tweaking.  The stuff about letting you set the kind by calling __init__ on ResourceReflector with that in kwargs is dumb, because these are all singletons, so you need to subclass them once per reflector type anyway.

athornton:
OK.  *Now* I think it's ready for review.

manics:
I'm a very big fan of auto-formatting python packages using black, but I think it's best done in a separate PR so we can ensure everyone's agreed, along with enabling pre-commit to ensure the code remains formatted. For an example pre-commit configuration from the JupyterHub repo see:
https://github.com/jupyterhub/jupyterhub/blob/166b00867f2c4fff0862565d882fe1c376fe4af2/.pre-commit-config.yaml#L6-L9
https://github.com/jupyterhub/jupyterhub/blob/166b00867f2c4fff0862565d882fe1c376fe4af2/.github/workflows/test.yml#L40-L41

Putting it in a separate PR also means it'll be easier to track down any problems should they occur in future.

athornton:
All right, I'll revert the blacken commit and then do what tidying up I can.

athornton:
I've unblackened it and it passes my tests locally.  I missed a suggestion in reflector.py so I'm gonna do that now.

athornton:
Now I need to unmangle the URLs....

athornton:
The 1.11 API isn't even published anymore.  I got...most of the docstring refs, and replaced trailing double-underscore with trailing underscore, which I think is how RST wants it?

athornton:
I think this is ready now.

kubespawner/reflector.py
```diff
@@ -152,16 +207,21 @@ def _list_and_update(self):
...
         )
+        if not self.omit_namespace:
+            kwargs["namespace"] = self.namespace
+
+        initial_resources = getattr(self.api, self.list_method_name)(**kwargs)
         # This is an atomic operation on the dictionary!
         initial_resources = json.loads(initial_resources.read())
-        self.resources = {p["metadata"]["name"]: p for p in initial_resources["items"]}
+        self.resources = {
+            f'{p["metadata"]["namespace"]}/{ p["metadata"]["name"]}': p for p in initial_resources["items"]}
```
manics, line 224:
```suggestion
            f'{p["metadata"]["namespace"]}/{p["metadata"]["name"]}': p for p in initial_resources["items"]}
```
Presumably a typo?

athornton: Good catch.  Yes and thanks.

kubespawner/spawner.py
athornton: Done!

kubespawner/spawner.py
athornton: Done, generally got it up to referencing 1.20, tried to make the formatting consistent within a section.  Honestly refactoring all the docstrings is probably its own PR, to impose some kind of consistent style here, but...another day.

kubespawner/spawner.py
athornton: I agree with both those sentiments.  I'd like to land this one, then I'll make that a configurable (Bool | Callable, maybe?)  or maybe just Bool and if you want to do something fancy, make it False and add the hook.

Then (or maybe first) I'll do a PR to blacken it and to add blacken-pre-commit.

kubespawner/spawner.py
athornton: Done!

kubespawner/spawner.py
athornton: I've gotten rid of the multinamespace classes entirely.

kubespawner/reflector.py
athornton: Done!

kubespawner/objects.py
athornton: Demangled.  I think I got them all?

kubespawner/__init__.py
athornton: Done!

kubespawner/spawner.py
athornton: Done!

kubespawner/spawner.py
athornton: Done!

athornton:
Is there anything else I need to do in order to get this one merged?

consideRatio:
@athornton thank you so much for your work on this! It is a big feature and I've personally been afraid of taking on the work effort to review it due to it. I do like to help though.

To help for example me to contribute as a reviewer, could you update the original topic where you summarize a bit what has changed and such? It is quite an effort to find the lines related to non-black stuff that are actual changes without knowing what to look for so clearly, but it is also great having such summary as part of the original post of the PR after merge as well.

athornton:
@consideRatio The initial comment has been updated.  Is that what you were looking for?

consideRatio:
@athornton yes! Thank you!!!

Note that the suggestion of cleaning up the namespace resource from the in-namespace-pod's lifecycle hook won't work, but its plausible to do it from a post_stop hook of the spawner I think.

---

@yuvipanda we have no tests for the KubeIngressProxy etc, and I think we are at a point where we will cause breaking changes to it by modifying the reflectors a bit. I think we can fix it, but I wonder if you have inputs with regards to what to do about the coupling between KubeSpawner and the less mature KubeIngressProxy project both residing in this github project.

consideRatio:
@yuvipanda we have no tests for the KubeIngressProxy etc, and I think we are at a point where we will cause breaking changes to it by modifying the reflectors a bit. I think we can fix it, but I wonder if you have inputs with regards to what to do about the coupling between KubeSpawner and the less mature KubeIngressProxy project both residing in this github project.

athornton:
Good point, I'll change that to spawner post_stop hook.

minrk (commented):
This is really close, thank you @athornton!

> replaced trailing double-underscore with trailing underscore, which I think is how RST wants it?

Unfortunately, no. Those were correct before to use `__` for anonymous links, and changing them to `_` has created one named link with several targets, resulting in warnings and possibly incorrect links. They were correct before, and it would be good to revert these changes.

I've tried to find all the actual code changes to review and they are looking really good to me! I did notice a few duplicated blocks, possibly caused by resolving merge conflicts? I highlighted the ones I found below.

Then we can make the standalone black PR so we don't have so much style churn to sift through.

In general, it's a really good principle that if an autoformatter would make a lot of changes to a file in any project, its changes shouldn't be included in a PR that does anything else. That is true for contributing to ~any open source project.

I don't usually do this, but because of including both the blacken and unblacken changes in this PR, I would propose we squash & merge this PR when it's ready rather than the usual merge commit, if that's okay with you @athornton?

athornton:
Thanks for the feedback.  I will get started on it as soon as I can, although there are a couple things ahead of it in the queue.  I have no issue with any of the requested changes, and no issue with a squash-and-merge.

kubespawner/spawner.py
```diff
@@ -1972,8 +2103,16 @@ def _start_watching_pods(self, replace=False):
         If replace=True, a running pod reflector will be stopped
         and a new one started (for recovering from possible errors).
         """
-        PodReflector.labels.update({'component': self.component_label})
-        return self._start_reflector("pods", PodReflector, replace=replace)
+        pod_reflector_class = PodReflector
+        if self.enable_user_namespaces:
+            reflector_omit_namespace = True
```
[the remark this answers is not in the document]
athornton: Removed.

kubespawner/spawner.py
```diff
@@ -1477,6 +1582,20 @@ def _expand_user_properties(self, template):
         # k8s object names cannot have trailing -
         return rendered.rstrip("-")
 
+    def _expand_user_namespace_name(self):
```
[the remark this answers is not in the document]
athornton: Sure.

athornton:
OK, @minrk , I think that's the lot.

minrk:
Awesome, thank you so much @athornton!

minrk:
code autoformatting PR in #473, adopting the same tooling used elsewhere in jupyterhub

meeseeksmachine:

https://discourse.jupyter.org/t/how-to-create-their-own-namespace-for-each-user/8039/3

meeseeksmachine:

https://discourse.jupyter.org/t/the-release-cycle-of-ztjh/8055/2

viktoriaas:
Hello everyone,
I would like to ask whether this feature is already incorporated in any version of KubeSpawner AND z2jh (as stated by @consideRatio in https://discourse.jupyter.org/t/the-release-cycle-of-ztjh/8055/2 from Feb'21.) If yes, how can it be turned on? I went through whole z2jh documentation and haven't found any mention of this feature. 

Any help is much appreciated 🫶🏼

consideRatio:
z2jh documentation wont mention this, it will be kubespawner documentation.

In z2jh, hub.extraConfig can be used to configure kubespawner for use with this feature.

PR #409, opened 2020-06-04
Support internal_ssl

minrk:
Continuing @sstarcher's great work in #386 to hopefully get this finished (thanks for working on it!)

todo:

- [x] implement move_certs to make sure secret mount matches load path
- [x] double-check that the hub CA needs to be added to the secret (according to spec, it shouldn't, I think)
- [x] (maybe) consolidate some repeated retry logic that's getting pretty long
- [x] test that dns_name routing is correct when internal-ssl is not enabled (maybe pull in get_pod_url from #408)
- [x] add a couple integration tests that
  - [x] launch
  - [x] and talk to pods with ssl

cc @consideRatio who mentioned an intention to pick this up, but is focused on the CI/CD rework at the moment. I'm hoping to help get this ready and also use it to test the CI/CD update PR so we can land it soon.

PR #460, opened 2020-11-25
Support pod IP template

dtaniwaki:
My notebook pods in calico-installed EKS with Amazon VPC networking changes the IP address when it's restarted on failure. Then, JupyterHub will lose the connection and cannot find the pod although it knows the pod is running by Kubernetes API.

I tried to make a workaround by using a headless service and pod subdomain, but it was not possible to customize the pod IP, so I'm sending this PR.

I will use the following manifest and `jupyterhub_config.py` to utilize this feature.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: notebooks
  namespace: jupyterhub
spec:
  selector:
    app: jupyterhub
    component: singleuser-server
  clusterIP: None
  ports: []
```

```python
c.KubeSpawner.pod_ip = "jupyter-{username}--{servername}.notebooks.jupyterhub"

def modify_pod_hook(spawner, pod):
    pod.spec.subdomain = "notebooks"
    pod.spec.hostname = spawner._expand_user_properties("jupyter-{username}--{servername}")
    return pod

c.KubeSpawner.modify_pod_hook = modify_pod_hook
```

Could you consider merging this change?

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -355,6 +355,24 @@ class KubeSpawner(Spawner):
         """
     )
 
+    pod_ip_template = Unicode(
+        config=True,
+        help="""
+        Template to use to form The IP address (or hostname) of user's pods.
+
+        e.g. 'jupyter-{username}--{servername}.notebooks.jupyterhub.svc.cluster.local',
+
+        `{username}` is expanded to the escaped, dns-label-safe username.
+        `{servername}` is expanded to the escaped, dns-label-safe server name, if any.
+
+        Trailing `-` characters in each domain level are stripped for safe handling of empty server names (user default servers).
+
+        This must be unique within the namespace the pods are being spawned
+        in, so if you are running multiple jupyterhubs spawning in the
+        same namespace, consider setting this to be something more unique.
+        """
+    )
+
     storage_pvc_ensure = Bool(
         False,
         config=True,
@@ -1955,7 +1973,13 @@ class KubeSpawner(Spawner):
                     ]
                 ),
             )
-        return (pod["status"]["podIP"], self.port)
+
+        if self.pod_ip_template:
+            ip = ".".join([s.rstrip("-") for s in self._expand_user_properties(self.pod_ip_template).split(".")])
+        else:
+            ip = pod["status"]["podIP"]
+
+        return (ip, self.port)
 
     async def _make_delete_pod_request(self, pod_name, delete_options, grace_seconds, request_timeout):
         """
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -358,7 +358,8 @@ class KubeSpawner(Spawner):
     pod_connect_ip = Unicode(
         config=True,
         help="""
-        Template to use to form The IP address (or hostname) of user's pods.
+        The IP address (or hostname) of user's pods which KubeSpawner connects to.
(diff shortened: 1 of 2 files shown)


kubespawner/spawner.py
```diff
@@ -355,6 +355,24 @@ def _namespace_default(self):
         """
     )
 
+    pod_ip_template = Unicode(
```
dtaniwaki, line 358: Or, `notenook_ip_template`?

kubespawner/spawner.py
```diff
@@ -1955,7 +1973,13 @@ async def _start(self):
                     ]
                 ),
             )
-        return (pod["status"]["podIP"], self.port)
+
+        if self.pod_ip_template:
+            ip = ".".join([s.rstrip("-") for s in self._expand_user_properties(self.pod_ip_template).split(".")])
```
yuvipanda, line 1978: Adding a comment here on why the rstrip happens would be great!

kubespawner/spawner.py
yuvipanda: pod_ip_template seems good, to match other traitlet names here.

yuvipanda:
Thanks for opening @dtaniwaki! Do you think you can add a test like https://github.com/jupyterhub/kubespawner/blob/master/tests/test_spawner.py#L102? Then we can merge that.

dtaniwaki:
@yuvipanda I added a unit test. Could you review it?

kubespawner/spawner.py
consideRatio: Thank you for your work on this @dtaniwaki! Could you also try to:
- communicate that this is how KubeSpawner will try to access the pod after creation (running in the hub pod), rather than what IP the pod will be created with
- communicate that this is optional, and we otherwise default to accessing the pod using its by k8s automatically assigned IP

@yuvipanda and @dtaniwaki I also suggest we name this `pod_connect_ip_template`, it would mimic the naming of [c.JupyterHub.hub_connect_ip](https://jupyterhub.readthedocs.io/en/stable/api/app.html#jupyterhub.app.JupyterHub.hub_connect_ip). What do you think?

dtaniwaki: @yuvipanda Which do you think better?

yuvipanda: @dtaniwaki I think we can just call it `pod_connect_ip`, and do template expansion, since the template expansion is the same thing we do for many other properties here.

yuvipanda:
@dtaniwaki this LGTM! Once the rename is done and the merge conflict resolved, I'll merge this. Apologies that it took so long!

yuvipanda:
I opened #464 to fix the test failures.

yuvipanda:
Thanks, @dtaniwaki!

minrk:
Sorry, I wasn't paying enough attention! I should have pulled out the get_pod_url from #409 when I realized it was going to take a while, which is also a fix for this issue and what I think we should do instead. #409 changes everything to use the dns name to connect by default, and `get_pod_url` us overrideable with a callable. When I figure out that one, I think we should probably not have two separate ways to set the connect url, and setting the ip is not generally enough (we definitely need to be able to set the protocol for internal_ssl as well).

dtaniwaki:
@minrk That makes sense!

PR #461, opened 2020-11-25
Support extra env

dtaniwaki:
As it's discussed in https://github.com/jupyterhub/kubespawner/issues/306, we can't add extra environment variables easily and intuitively.

I implemented `extra_env` config so we can easily add extra environment variables.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1274,6 +1274,17 @@ class KubeSpawner(Spawner):
         """
     )
 
+    extra_env = Dict(
+        config=True,
+        help="""
+        Extra environment variables to set on the single-user server containers.
+
+        The keys and values must both be strings.
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
Hello @dtaniwaki! Thanks for the PR.

My understanding right now is that you can use `c.KubeSpawner.environment` to set either key-value pair env vars, or full kubernetes style env vars that reference secrets / configmap / etc (since https://github.com/jupyterhub/kubespawner/pull/426 landed). 

This PR adds username / property expansion to environment variables. Is that the sole intent of `extra_env`?

dtaniwaki:
I used `extra_container_config` and that's why it didn't work.

> This PR adds username / property expansion to environment variables. Is that the sole intent of extra_env?

This is an extra feature for cutsomizability, but it's not necessary for me for now.

yuvipanda:
Thanks, @dtaniwaki!

PR #463, opened 2020-12-11
Expand storage selector

dtaniwaki:
I'd like to use pre-created persistent volume for each user, but currently, we don't have any method to expand username for storage selector.


yuvipanda:
This LGTM! Thanks, @dtaniwaki!

yuvipanda:
Hmm, the test failures seem to be from the new pip solver.

yuvipanda:
Opened #464 to address the test failures.

PR #467, opened 2020-12-14
Wrap concurrent.futures Future in polling function

### LINKED ISSUE #466: Polling fails on hub restart with TypeError: object Future can't be used in 'await' expression
On hub restart, poll for existing servers fails with:

```
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 2032, in check_spawner
        status = await spawner.poll()
      File "/usr/local/lib/python3.8/dist-packages/kubespawner/spawner.py", line 1610, in poll
        await self.pod_reflector.first_load_future
    TypeError: object Future can't be used in 'await' expression
```

I am no expert on the various async libraries, but think this may be because we are using a asyncio await with a concurrent.futures Future?

jupyterhub-1.3.0
jupyterhub-kubespawner-0.14.1
tornado-6.1

ondave:
Cannot await a concurrent.futures Future, so easiest fix is to wrap with asyncio.wrap_future

Closes #466 - Edit by @consideRatio.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1642,7 +1642,7 @@ class KubeSpawner(Spawner):
         """
         # have to wait for first load of data before we have a valid answer
         if not self.pod_reflector.first_load_future.done():
-            await self.pod_reflector.first_load_future
+            await asyncio.wrap_future(self.pod_reflector.first_load_future)
         data = self.pod_reflector.pods.get(self.pod_name, None)
         if data is not None:
             if data["status"]["phase"] == 'Pending':


consideRatio:
@minrk could you help review this one liner? I think could be a quite severe bug with a simple fix.

I'm not so confident about these things even though I learned a lot working with the async in some PRs before.

minrk:
Yup, this is right!

Background: tornado `@gen.coroutine` coroutines have a lot more 'awaitables' than stdlib asyncio (concurrent.futures.Futures, tornado's own concurrent.Future which is now an alias for asyncio.Future, internal objects like YieldPoints). A lot of this variety has fallen away with recent convergence of asyncio/tornado, but one distinction remains: asyncio doesn't support concurrent.futures.Future without this explicit wrapper step, but tornado still does. So our recent switch to `async def` coroutines means we had to make sure that we always ensure our tornado awaitables are cast to asyncio awaitables where necessary, and this one was missed.

A bit about *why* we have this: `concurrent.futures.Future` objects are threadsafe, so can be used to communicate a via a standard Future API across threads. asyncio Futures **must not** be used across threads. So one good way to signal from one thread to another is to use a concurrent.futures.Future. This `wrap_future` is the adaptation to make the threadsafe Future awaitable in asyncio.

It bothers me perhaps more than it should that:

1. asyncio thinks this is enough of an issue to provide the `asyncio.wrap_future` utility, but
2. asyncio does not bother to call this itself when it's clear and unambiguous that it's always the right thing to do (and/or concurrent.futures.Future doesn't implement this within its own `__await__` methods to make them awaitable by default)

consideRatio:
Thank you @ondave for reporting and fixing this issue!!! :tada: :heart:!

@minrk is this a consequence of https://github.com/jupyterhub/jupyterhub/pull/3242, which also mean it would only influence jupyterhub>=1.3.0 ? Should something be patched in JupyterHub itself following this?

minrk:
> is this a consequence of jupyterhub/jupyterhub#3242,

Close. It should only be relevant within the given function, so I'm pretty sure this was caused by #444, which made the same switch on this particular method. Take for example:

```python
from concurrent.futures import ThreadPoolExecutor
from tornado import gen, ioloop


pool = ThreadPoolExecutor(1)


@gen.coroutine
def gen_coroutine():
    return (yield pool.submit(lambda: "tornado ok"))


async def async_def():
    return (await pool.submit(lambda: "asyncio ok"))


async def main():
    print("tornado?")
    print(await gen_coroutine())
    print("asyncio?")
    print(await async_def())

if __name__ == "__main__":
    ioloop.IOLoop.current().run_sync(main)
```

In this script, the tornado implementation can yield a c.f.Future and the async def can't, but the outer `async def main()` doesn't need to know about it, so it's not a far-reaching compatibility issue.

JupyterHub has its own compatibility layer in `utils.maybe_future` that already calls `wrap_future` where necessary. I did review https://github.com/jupyterhub/jupyterhub/pull/3242 with that in mind, i.e. making sure we have `utils.maybe_future` where we have user-overrideable APIs that could be allowed to return these Futures (never officially supported in the docs, but technically would have worked prior to 1.3). That doesn't mean I caught everything, of course :)

PR #469, opened 2020-12-15
Changelog for 0.15.0

consideRatio:
What made me want to cut a release was the bugfixes, but as there are many enhancements awaiting a release i suggest we cut 0.15.0 instead of 0.14.2.

A bugfix I affecting 0.14.0 and 0.14.1 worth highlighting: https://github.com/jupyterhub/kubespawner/pull/467, thank you @ondave! :tada:


yuvipanda:
Thanks, @consideRatio!

Issue #398, opened 2020-03-26
Tokens should be put in `kind: Secret` instead of env vars


elgalu:
https://github.com/jupyterhub/kubespawner/blob/892780ae86c6e74172065ab984bf2d059d00c31f/kubespawner/spawner.py#L1498

https://github.com/jupyterhub/jupyterhub/blob/914a3eaba5c83f639b4b5d8873c6e9d93e809f31/jupyterhub/spawner.py#L722

consideRatio:
This relates a bit to #110.

tirumerla:
Hi @consideRatio - I was also wondering about this today as i found two env. variables `JUPYTERHUB_API_TOKEN` & `JPY_API_TOKEN` when i did `kubectl get pod -o yaml <singleuser_pod_name>`. Here are my concerns with this approach rather than using k8s secrets 

- "Plain Text" env. variables don't have basic security measures like providing encryption at rest or having some sort of authorization policies. 
- You can enforce policies/permissions by giving read-only access to pods but not secrets. Even if you want to give access to secrets you can have granular policies to give read-only access to specific secrets.

Would be great if we can use k8s secrets for these variables as opposed to "plain text" env. variables.

Thanks.

yuvipanda:
This makes sense. The complexity here will be creating & cleaning up the secret objects with pod start / stop. We could explore doing this with [kubernetes garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/), although since the secret would need to be created before the pod - not sure if that'll work for that. 

We already create PVCs before creating pods, I think we can extend that logic to secrets. While we're there, we could possibly just put all env vars there instead of just 'sensitive' ones.

Happy to review any PRs :)

minrk:
I think tackling this will give us a chance to generally improve logic around "objects created to go with the pod". Using consistent labels for *any* object created by kubespawner associated with a user pod should make it manageable to create and clean up everything without lifecycle concerns when e.g. gc is unavailable due to ordering.

tirumerla:
Thanks @yuvipanda and @minrk for the response. This is currently a security concern for us, and would love to help. Unfortunately I'm not sure where to start, however I'll try helping as much as I can.

yuvipanda:
[...]

PR #471, opened 2020-12-23
read env from secrets

tirumerla:
what
---

- Create a secret with all the env. variables and bind this to the pod

why
---

- Currently single user pod manifest show all the env. variables in plain text. To fix this issue and to follow the security standards and best practices all the env. variables are pulled into a secret which gives more flexibility for access policies.

references
---
See #398 for details.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -7,6 +7,7 @@ implementation that should be used by JupyterHub.
 
 import asyncio
 import json
+import base64
 import multiprocessing
 import os
 import string
@@ -1624,7 +1625,7 @@ class KubeSpawner(Spawner):
             supplemental_gids=supplemental_gids,
             run_privileged=self.privileged,
             allow_privilege_escalation=self.allow_privilege_escalation,
-            env=self.get_env(),
+            env_from=self.secret_name,
             volumes=self._expand_all(self.volumes),
             volume_mounts=self._expand_all(self.volume_mounts),
             working_dir=self.working_dir,
@@ -1665,12 +1666,11 @@ class KubeSpawner(Spawner):
         annotations = self._build_common_annotations(
             self._expand_all(self.extra_annotations)
         )
-
+        
         return make_secret(
             name=self.secret_name,
+            str_data=self.get_env(),
             username=self.user.name,
-            cert_paths=self.cert_paths,
-            hub_ca=self.internal_trust_bundles['hub-ca'],
             owner_references=[owner_reference],
             labels=labels,
             annotations=annotations,
@@ -1771,6 +1771,24 @@ class KubeSpawner(Spawner):
         # deprecate image
         env['JUPYTER_IMAGE_SPEC'] = self.image
         env['JUPYTER_IMAGE'] = self.image
+        if self.internal_ssl:
+            with open(self.cert_paths['keyfile'], 'r') as file:
+                env['ssl.key'] = file.read()
+
+            with open(self.cert_paths['certfile'], 'r') as file:
+                env['ssl.crt'] = file.read()
+
+            with open(self.cert_paths['cafile'], 'r') as file:
+                env["notebooks-ca_trust.crt"] = file.read()
+
+            with open(self.internal_trust_bundles['hub-ca'], 'r') as file:
+                encoded = base64.b64encode(file.read().encode("utf-8"))
+                env["notebooks-ca_trust.crt"] = env[
+                    "notebooks-ca_trust.crt"
+                ] + file.read()
+            env['JUPYTERHUB_SSL_KEYFILE'] = self.secret_mount_path + "ssl.key"
+            env['JUPYTERHUB_SSL_CERTFILE'] = self.secret_mount_path + "ssl.crt"
+            env['JUPYTERHUB_SSL_CLIENT_CA'] = (self.secret_mount_path + "notebooks-ca_trust.crt")
 
         return env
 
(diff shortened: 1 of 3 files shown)


consideRatio:
Quick note before i forget: there is work done in #409 about a k8s secret resource to go with the pod.

sai-tirumerla-leapyear:
> Edit: aaah nice you have seen!

Thanks @consideRatio, Yep made changes on top of this.

> I want us to still allow for env set through refs to other secrets etc. I have not yet inspected the code (on mibile atm), but id like some discussion on the choice for the API you suggest kubespawner should implement.

Sounds good, let me know once you review the code. We can take it from there based on your feedback. :)

tirumerla:
> Edit: aaah nice you have seen!

Thanks @consideRatio, Yep made changes on top of this.

> I want us to still allow for env set through refs to other secrets etc. I have not yet inspected the code (on mibile atm), but id like some discussion on the choice for the API you suggest kubespawner should implement.

Sounds good, let me know once you review the code. We can take it from there based on your feedback. :)

tirumerla:
Hey @consideRatio @yuvipanda, Happy new year :) Did you guys get a chance to look at this?

consideRatio:
@tirumerla thank you for your work. Is it okay if I suggest what could cause issues, and ask you to comment if you think there will be issues?

Main concern, will this PR stop supporting usage of the feature developed in https://github.com/jupyterhub/kubespawner/pull/426 which was a long standing feature requested? Imagine that the user wants to set environment variables, and have them be set through the secret, but also be able to set any kind of environment variable directly - such as a reference to a configmap or another secret? Will this PR cause such conflicts?

I hope you can excuse me for being a bit lazy and not investigating this in detail by code inspection!

tirumerla:
> Will this PR cause such conflicts?

@consideRatio Thank you for your response. This is a good point and a great catch - I will have to make some changes, any environment variable that is passed needs to be validated in `get_env` and construct the environment dict which is then passed to `make_secret` as opposed to validating in `make_pod`

> I hope you can excuse me for being a bit lazy and not investigating this in detail by code inspection!

Not a problem :)

tirumerla:
@consideRatio I have updated my code to allow the feature developed in #426. Let me know what you think about it.

tests/test_objects.py
```diff
@@ -577,29 +558,12 @@ def test_make_pod_with_env():
...
+                    "envFrom": [
                         {
-                            'name': 'TEST_KEY_3',
-                            'valueFrom': {
-                                'secretKeyRef': {
-                                    'name': 'my-k8s-secret',
-                                    'key': 'password',
-                                },
-                            },
-                        },
```
consideRatio, line 602: Can you ensure old tests are retained and only update what changed? I believe the output 581-584 is what really change, but not the others, right?

```
                        {
                            'name': 'TEST_KEY_1',
                            'value': 'TEST_VALUE',
                        },
```

tirumerla: > Can you ensure old tests are retained and only update what changed?

Yes good point, fixing.

tirumerla: I changed the tests, added one more test to spawner to verify 'valueFrom' env. variables.

kubespawner/spawner.py
```diff
@@ -1768,11 +1771,51 @@ def get_env(self):
...
+
+            with open(self.internal_trust_bundles['hub-ca'], 'r') as file:
+                env["notebooks-ca_trust.crt"] = env[
+                    "notebooks-ca_trust.crt"
+                ] + file.read()
+            env['JUPYTERHUB_SSL_KEYFILE'] = self.secret_mount_path + "ssl.key"
+            env['JUPYTERHUB_SSL_CERTFILE'] = self.secret_mount_path + "ssl.crt"
+            env['JUPYTERHUB_SSL_CLIENT_CA'] = (self.secret_mount_path + "notebooks-ca_trust.crt")
+
+        return env, prepared_env
```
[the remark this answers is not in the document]
tirumerla: > I consider introducing get_env_from alongside get_env here, I dislike the complexity we embed in this by returning a tuple.

Yes, this is a good point. I thought about this, creating a new `get_env_from` just for env. variables using "valueFrom" but in the moment i went with getting something up quicker rather than simpler :). You're right, i will take a look at this to make it simpler.

kubespawner/spawner.py
```diff
@@ -1768,11 +1770,57 @@ def get_env(self):
...
+                env["notebooks-ca_trust.crt"] = env[
+                    "notebooks-ca_trust.crt"
+                ] + file.read()
+            env['JUPYTERHUB_SSL_KEYFILE'] = self.secret_mount_path + "ssl.key"
+            env['JUPYTERHUB_SSL_CERTFILE'] = self.secret_mount_path + "ssl.crt"
+            env['JUPYTERHUB_SSL_CLIENT_CA'] = (self.secret_mount_path + "notebooks-ca_trust.crt")
 
         return env
+    
+    def get_env_value_from(self):
```
tirumerla, line 1806: @consideRatio I've created a separate `get_env_value_from` to filter only the env. variables which has `value_from`, hence the name :). Now `get_env` will filter only the env. variables with values but not `value_from`. Also i'm using `V1EnvVar` in this since we already have the right format for it to process.

For example.  
This `"TEST_KEY_2": {'valueFrom': {'secretKeyRef': {'name': 'my-test-secret', 'key': 'password'}}}` being formatted as `{"name": "TEST_KEY_2", "value_from": {'secretKeyRef': {'name': 'my-test-secret', 'key': 'password'}}}`. I'm creating list of these env. variables containing value_from which are then passed to V1Container object in `env` variable.

tirumerla:
I have also tested this internally creating configmap of changed files `spawner.py` and `objects.py` , mounting these on the hub pod, logged in as a single user and spawned the single user pod. I saw the secret with all env. variables getting created, I was also able to test executing `env` from my notebook cell on the single-user pod to verify if i can see all the env. variables from secret and also "valueFrom".

tirumerla:
hi @consideRatio any updates on this?

diego-leapyear:
Hi @consideRatio , I was wondering if the Jupyter team had the chance to look into this PR. Thank you!

consideRatio:
Hi @tirumerla! Sorry I have no update from my end.

I think this feature adds value but it also add quite significant complexity related to a not so well tested feature (internal_ssl, and the recent addition of a k8s secret). Due to this, I've not personally given it much priority.

tirumerla:
> Hi @tirumerla! Sorry I have no update from my end.
> 
> I think this feature adds value but it also add quite significant complexity related to a not so well tested feature (internal_ssl, and the recent addition of a k8s secret). Due to this, I've not personally given it much priority.

Thanks @consideRatio. This makes sense. We will use this feature internally for now then.

sstarcher:
This would be a nice security improvement

SarunasAzna:
Good day!. Any news about implemetation of this feature?

PR #474, opened 2021-01-18
fix some spurious additions in tests

minrk:
some blocks added due to merge conflicts in #458

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2298,12 +2298,12 @@ class KubeSpawner(Spawner):
                 # required for creating owner reference
                 await exponential_backoff(
                     lambda: self.pod_has_uid(
-                        self.pod_reflector.pods.get(self.pod_name, None)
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
LGTM!

Merge away if you also think the test failures are unrelated.

minrk:
Unfortunately, this reveals a test failure that appears to have been broken by #458 (reflected pods never get uid, despite having them assigned), but a typo in the PR prevented the affected test from running. I'm going to try to figure that one out for a bit.

minrk:
I think I got it - it was a couple missing updates from pod_name -> ref_key in the internal_ssl block, missed because the internal_ssl test was accidentally being skipped. Should be fine, but let's wait for CI to confirm. Then I'll rebase #473 .

consideRatio:
@minrk thank you for following up thoroughly on this!

minrk:
tests passing

PR #475, opened 2021-01-19
add method to delete namespaced PVC in spawner base class

### LINKED ISSUE #446: Deleting named servers should delete the corresponding PVC
## Current state

In the current implementation of z2jh on k8s, when a named server is started, a PVC is created, which is what we want to happen. When a user *stops* their server, this PVC persists, which we also want to happen. Currently, when a user *deletes* their named server, the corresponding PVC is not deleted or released.

Persisting the PVC after the deletion of the servers leaves stranded claims that have to be manually cleaned up. Additionally, if a user creates an identically-named server sometime later, their seemingly newly created server will be attached to this stranded claim since the k8s resource names will point at each other. Thus a "newly" created server will not be a clean environment.

## Proposed change

The PVC for a named server should be deleted when the named server is deleted.

nsshah1288:
## PR Summary by Erik

This PR implements a `delete_forever` hook for KubeSpawner. The `delete_forever` hook will be respected by JupyterHub 1.4.0 and later and was implemented in https://github.com/jupyterhub/jupyterhub/pull/3337. The idea of this hook is to allow the Spawners to perform cleanup of a users associated resources at the time a JupyterHub user is deleted.

As KubeSpawner can create PVCs for users, this KubeSpawner implementation checks for a PVC that it could have created for a user and tries to delete it. This behavior should be documented properly in the CHANGELOG.md for the next KubeSpawner release, I'm not sure if I think it would be labelled as a "breaking" change, but it is a change in the user experience that is quite important to be aware about.

This PR closes #446.

## Original PR message

have tested with sqlite-pvc

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2329,7 +2329,7 @@ class KubeSpawner(Spawner):
         delete = getattr(self.api, "delete_namespaced_{}".format(kind))
         read = getattr(self.api, "read_namespaced_{}".format(kind))
 
-        # first, attempt to delete the resouce
+        # first, attempt to delete the resource
         try:
             self.log.info(f"Deleting {kind}/{name}")
             await gen.with_timeout(
@@ -2577,9 +2577,7 @@ class KubeSpawner(Spawner):
             else:
                 raise
 
-    async def _make_delete_pvc_request(
-        self, pvc_name, delete_options, grace_seconds, request_timeout
-    ):
+    async def _make_delete_pvc_request(self, pvc_name, request_timeout):
         """
         Make an HTTP request to delete the given PVC
 
@@ -2594,8 +2592,6 @@ class KubeSpawner(Spawner):
                     self.api.delete_namespaced_persistent_volume_claim,
                     name=pvc_name,
                     namespace=self.namespace,
-                    body=delete_options,
-                    grace_period_seconds=grace_seconds,
                 ),
             )
             return True
@@ -2815,7 +2811,7 @@ class KubeSpawner(Spawner):
                 self.log.exception("Failed to create namespace %s", self.namespace)
                 raise
 
-    async def delete_forever(self, now=False):
+    async def delete_forever(self):
         """Called when a user is deleted.
 
         This can do things like request removal of resources such as persistent storage.
@@ -2824,21 +2820,10 @@ class KubeSpawner(Spawner):
         This will only be called once on the user's default Spawner.
         Supported by JupyterHub 1.4.0+.
         """
-        delete_options = client.V1DeleteOptions()
-
-        if now:
-            grace_seconds = 0
-        else:
-            grace_seconds = self.delete_grace_period
-
-        delete_options.grace_period_seconds = grace_seconds
-
         await exponential_backoff(
             partial(
                 self._make_delete_pvc_request,
                 self.pvc_name,
-                delete_options,
-                grace_seconds,
                 self.k8s_api_request_timeout,
             ),
             f'Could not delete pvc {self.pvc_name}',
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2577,6 +2577,41 @@ class KubeSpawner(Spawner):
             else:
                 raise
 
+    async def _make_delete_pvc_request(
+        self, pvc_name, delete_options, grace_seconds, request_timeout
+    ):
+        """
+        Make an HTTP request to delete the given PVC
+
+        Designed to be used with exponential_backoff, so returns
+        True / False on success / failure
+        """
+        self.log.info("Deleting pvc %s", pvc_name)
+        try:
+            await gen.with_timeout(
+                timedelta(seconds=request_timeout),
+                self.asynchronize(
+                    self.api.delete_namespaced_persistent_volume_claim,
+                    name=pvc_name,
+                    namespace=self.namespace,
+                    body=delete_options,
+                    grace_period_seconds=grace_seconds,
+                ),
+            )
+            return True
+        except gen.TimeoutError:
+            return False
+        except ApiException as e:
+            if e.status == 404:
+                self.log.warning(
+                    "No pvc %s to delete. Assuming already deleted.",
+                    pvc_name,
+                )
+                # If there isn't a PVC to delete, that's ok too!
+                return True
+            else:
+                raise
+
     async def stop(self, now=False):
         delete_options = client.V1DeleteOptions()
 
@@ -2780,7 +2815,7 @@ class KubeSpawner(Spawner):
                 self.log.exception("Failed to create namespace %s", self.namespace)
                 raise
 
-    async def delete_forever(self):
+    async def delete_forever(self, now=False):
         """Called when a user is deleted.
 
         This can do things like request removal of resources such as persistent storage.
@@ -2789,18 +2824,23 @@ class KubeSpawner(Spawner):
         This will only be called once on the user's default Spawner.
         Supported by JupyterHub 1.4.0+.
         """
-        try:
-            self.api.read_namespaced_persistent_volume_claim(
-                self.pvc_name, self.namespace
-            )
-        except ApiException as e:
-            if e.status == 404:
-                self.log.warning(
-                    "Could not delete %s. This PVC does not exist.", self.pvc_name
-                )
-            else:
-                raise
+        delete_options = client.V1DeleteOptions()
+
+        if now:
+            grace_seconds = 0
         else:
-            self.api.delete_namespaced_persistent_volume_claim(
-                self.pvc_name, self.namespace
-            )
+            grace_seconds = self.delete_grace_period
+
+        delete_options.grace_period_seconds = grace_seconds
+
+        await exponential_backoff(
+            partial(
+                self._make_delete_pvc_request,
+                self.pvc_name,
+                delete_options,
+                grace_seconds,
+                self.k8s_api_request_timeout,
+            ),
+            f'Could not delete pvc {self.pvc_name}',
+            timeout=self.k8s_api_request_retry_timeout,
+        )
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2779,3 +2779,28 @@ class KubeSpawner(Spawner):
                 # It's fine if it already exists
                 self.log.exception("Failed to create namespace %s", self.namespace)
                 raise
+
+    async def delete_forever(self):
+        """Called when a user is deleted.
+
+        This can do things like request removal of resources such as persistent storage.
+        Only called on stopped spawners, and is likely the last action ever taken for the user.
+
+        This will only be called once on the user's default Spawner.
+        Supported by JupyterHub 1.4.0+.
+        """
+        try:
+            self.api.read_namespaced_persistent_volume_claim(
+                self.pvc_name, self.namespace
+            )
+        except ApiException as e:
+            if e.status == 404:
+                self.log.warning(
+                    "Could not delete %s. This PVC does not exist.", self.pvc_name
+                )
+            else:
+                raise
+        else:
+            self.api.delete_namespaced_persistent_volume_claim(
+                self.pvc_name, self.namespace
+            )
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -249,10 +249,10 @@ class KubeSpawner(Spawner):
         config=True,
         help="""
         Location (absolute filepath) for CA certs of the k8s API server.
-        
-        Typically this is unnecessary, CA certs are picked up by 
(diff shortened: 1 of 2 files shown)


nsshah1288:
@consideRatio @minrk If either of you have time, could you please review this PR? 
The PR related to this has been merged in the jupyterhub repo: https://github.com/jupyterhub/jupyterhub/pull/3337

Thanks!

kubespawner/spawner.py
```diff
@@ -2633,3 +2633,27 @@ async def _ensure_namespace(self):
...
+            if e.status == 404:
+                self.log.warning(
+                    "Could not delete %s. This PVC does not exist.", self.pvc_name
+                )
+            else:
+                raise
+        else:
+            self.api.delete_namespaced_persistent_volume_claim(
+                self.pvc_name, self.namespace
+            )
```
consideRatio, line 2659: Thanks for your work on this @nsshah1288! I think this is a feature to support now that it is part of JupyterHub 1.4.0+, which btw can be worth mentioning in the docstring.

@nsshah1288 can you elaborate on why this deletion logic make sense to structure in try/except/else and what the difference between `read_namespaced_persistent_volume_claim_with_http_info` and without `http_info` is?

nsshah1288: Thanks @consideRatio ! I have updated the docstring and also removed `_with_http_info` from the first API call. `read_namespaced_persistent_volume_claim` actually calls `read_namespaced_persistent_volume_claim_with_http_info`  in `CoreV1Api`, and the  `_with_http_info` I believe just indicates that the method is making an HTTP request to check if the specified PVC exists. I have not yet tested this change, but I will update here when I do.

I was thinking that when `delete_forever` is called, I first check if the specified PVC exists. If it doesn't exist, I throw an exception, and only if the PVC exists do I execute the `delete_namespaced_persistent_volume_claim`, hence the try/except/else. Does that sound OK?

nsshah1288: hi @consideRatio , I tested this change and it works well. When clicking 'delete user', it deletes the PVC and PV of the user if it exists. (tested with sqlite and 'dynamic' storage setting in values.yaml). Deleting users that don't have existing PVCs or PVs also works fine.

kubespawner/spawner.py
```diff
@@ -2633,3 +2633,27 @@ async def _ensure_namespace(self):
...
                 self.log.exception("Failed to create namespace %s", self.namespace)
                 raise
+
+    async def delete_forever(self):
+        """Called when a user is deleted.
+
+        This can do things like request removal of resources such as persistent storage.
+        Only called on stopped spawners, and is likely the last action ever taken for the user.
+
+        Will only be called once on the user's default Spawner. Will be a part of JupyterHub 1.4.0+
```
manics, line 2643:
```suggestion
        This will only be called once on the user's default Spawner.
        Supported by JupyterHub 1.4.0+.
```

manics (commented):
Do you think it's practicle to add a test for this new feature, especially as it may not be tested routinely?
There's an existing test which checks that secrets are deleted
https://github.com/jupyterhub/kubespawner/blob/c215e24c3816e06b87409a50ffef2125465d1c44/tests/test_spawner.py#L278-L299
so maybe that structure could be copied?

kubespawner/spawner.py
consideRatio: Thanks for verifying that it worked!

With regards to "_with_http_info" or not, I suggest we align with the method for deleting other resources with kubespawner. If we learn that it can make sense to use `_with_http_info` or not we update all our calls in kubespawner.

nsshah1288:
@manics @consideRatio I'm having some trouble trying to write the test case. I am trying to call `spawner.start()`, have the PVC created, verify it's created, and then delete the PVC and verify that it's gone. However, with my current test, I keep getting the exception _'KubeSpawner' object has no attribute 'api'_ when it is calling the method to create a PVC.
image

Do you have any suggestions on how to fix this? Thanks!

nsshah1288:
Hi @manics @consideRatio I have fixed the test case. Can you please review it?
I apologize for the delay on this. Thanks!

kubespawner/spawner.py
nsshah1288: Hi @consideRatio, could you please let me know if anything else is required for this PR? I tried to resolve your comments and also wrote a test case. Any feedback would be great, thanks!

octavd: Hello @consideRatio , did you had some time to review this change? It would be amazing to have it in the official version. At the moment i'm using a modified version of JH with those changes and works like a charm in Openshift.

consideRatio (commented):
Thank you @nsshah1288 so much for working to make this supported in JupyterHub itself which was the proper way of implementing this in KubeSpawner!

This PR looks good to me as it is content wise!

Could you as a final change also squash/fixup the 36 commits into one or two commits so it becomes less complicated to inspect the history of changes later?

nsshah1288:
Hi @consideRatio, thank you much for the review!
I apologize, but I couldn't for the life of me figure out how to squash my commits, so I made a new PR here with the same content: https://github.com/jupyterhub/kubespawner/pull/504

If this is not sufficient, please let me know and I am willing to try whatever you need. Thank you!

consideRatio:
I gave it a try as I was eager to learn more about this. It becomes complicated whenever there are multiple merges to various stages of the master.

I tried various things that failed but then I got the following to work as I wanted. Want to give it a try?

```
# Bonus detail:
# this makes your commit be associated with your username on GitHub
# but you could also add the email regn.net email to your user account
git config --global user.email "nsshah1288@users.noreply.github.com"

git remote add upstream https://github.com/jupyterhub/kubespawner
git fetch upstream
git reset --hard upstream/master
git merge --squash 0836260 # this is the latest commit in this PR
git commit -m "add delete_forever hook to cleanup user's PVC on user deletion"
```

nsshah1288:
Amazing @consideRatio ! Thank you so much.
I followed your steps, and then I had to resolve a merge conflict so that created a second commit. Should I try to squash these two into one commit or is this fine? Thanks again

kubespawner/spawner.py
```diff
@@ -2779,3 +2779,28 @@ async def _ensure_namespace(self):
...
+            if e.status == 404:
+                self.log.warning(
+                    "Could not delete %s. This PVC does not exist.", self.pvc_name
+                )
+            else:
+                raise
+        else:
+            self.api.delete_namespaced_persistent_volume_claim(
+                self.pvc_name, self.namespace
+            )
```
consideRatio, line 2806: @minrk do you have any suggested changes to this section?

I'm a bit uncertain about the use of async without any `await` statement and that we don't have any timeouts defined for example.

For reference, this is the async stop() method:

```python
    async def stop(self, now=False):
        delete_options = client.V1DeleteOptions()

        if now:
            grace_seconds = 0
        else:
            grace_seconds = self.delete_grace_period

        delete_options.grace_period_seconds = grace_seconds

        ref_key = "{}/{}".format(self.namespace, self.pod_name)
        await exponential_backoff(
            partial(
                self._make_delete_pod_request,
                self.pod_name,
                delete_options,
                grace_seconds,
                self.k8s_api_request_timeout,
            ),
            f'Could not delete pod {ref_key}',
            timeout=self.k8s_api_request_retry_timeout,
        )

        try:
            await exponential_backoff(
                lambda: self.pod_reflector.pods.get(ref_key, None) is None,
                'pod %s did not disappear in %s seconds!'
                % (ref_key, self.start_timeout),
                timeout=self.start_timeout,
            )
        except TimeoutError:
            self.log.error(
                "Pod %s did not disappear, restarting pod reflector", ref_key
            )
            self._start_watching_pods(replace=True)
            raise
```

minrk: > I'm a bit uncertain about the use of async without any await

This in principle is fine, and correct if you are implementing a generally async API but don't have anything to wait for. In this case, however, all k8s API requests should be async via `self.asynchronize`, in which case there will be awaits. Following the timeout/conflict/error handling pattern in our other API requests makes sense.

Yes, I think we do want to make sure to always make kube API calls from the background with retries/error handling, as we do in e.g. `_make_*_request`. We should never call `self.api.anything()` from the main thread.

nsshah1288: Hi @consideRatio , thanks to you and @minrk for the suggestions! I tried to implement what you outlined above. I tested the changes again with sqlite-pvc and it worked well. Please let me know if any other changes are required.

nsshah1288: Hi @consideRatio , do the changes look OK?

consideRatio: @nsshah1288 thanks for your work on this! :tada:

If the tests pass and the following changes below LGTY I think we could be ready to merge this.

- I cleaned up the merge commit (reset everything to the current master branch and then manually cherry picked all but the merge commit and solving some trivial merge conflicts).
- By doing above I --force pushed to this PRs branch, backing up its previous state in https://github.com/consideRatio/kubespawner/tree/feature/shahn3_pvcDeletion-backup
- I added ea12ea79 because the function didn't have the correct signature matching JupyterHub's Spawner base class
- I opted to not default to delete_grace_period for the PVC deletion request because as it is explicitly described to be used for the pod deletion requests.

  ```python
      delete_grace_period = Integer(
          1,
          config=True,
          help="""
          Time in seconds for the pod to be in `terminating` state before is forcefully killed.
  
          Increase this if you need more time to execute a `preStop` lifecycle hook.
  
          See https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods for
          more information on how pod termination works.
  
          Defaults to `1`.
          """,
      )
  ```

kubespawner/spawner.py
```diff
@@ -2779,3 +2810,22 @@ async def _ensure_namespace(self):
...
                 raise
+
+    async def delete_forever(self):
+        """Called when a user is deleted.
+
+        This can do things like request removal of resources such as persistent storage.
+        Only called on stopped spawners, and is likely the last action ever taken for the user.
+
+        This will only be called once on the user's default Spawner.
+        Supported by JupyterHub 1.4.0+.
```
consideRatio, line 2821:
```suggestion
        Supported by JupyterHub 1.4.1+.
```

@minrk besides not being sure about if 1.4.1 or 1.4.0 is required, I consider this PR ready for merge.

consideRatio (approved):

kubespawner/spawner.py
nsshah1288: Thank you so much for your detailed help on this @consideRatio ! The changes look good to me.

kubespawner/spawner.py
minrk: Sure, let's say 1.4.1. It's added in 1.4, but has some bugs that mean it's probably good to be clear not to expect it to work before 1.4.1

minrk:
I added a condition and clarified a docstring:

- `KubeSpawner.delete_pvc = False` allows opt-out of deletion for anyone who wants to keep the older behavior
- check if a named server's PVC has `{servername}` in it, because we don't want deleting named servers to delete the default server's PVC
- The docstring wasn't accurate to say that only the default Spawner is deleted. All Spawners can be deleted, all of which trigger this `delete_forever` hook.

minrk (approved):

kubespawner/spawner.py
```diff
@@ -2577,6 +2593,37 @@ async def _make_delete_pod_request(
...
+            return True
+        except gen.TimeoutError:
+            return False
+        except ApiException as e:
+            if e.status == 404:
+                self.log.warning(
+                    "No pvc %s to delete. Assuming already deleted.",
+                    pvc_name,
+                )
+                # If there isn't a PVC to delete, that's ok too!
```
mriedem, line 2622: I've got a question about this and the warning a couple of lines above. In our deployment we're currently using kubespawner 0.15 and jupyterhub 1.3. I'm looking to upgrade to z2jh 1.2.0, jupyterhub 1.5.0 and kubespawner 1.1.2. We use a single PVC that is backed by cloud object storage so all of the singleuser-server pods share the same PVC. We leave the default of `storage_pvc_ensure=False` which from looking at the code means kubespawner won't actually create a PVC named `self.pvc_name` in `_start`. So shouldn't `delete_forever` also be checking `storage_pvc_ensure`? Maybe not if it's assumed that someone could pre-create the PVC using the same `pvc_name_template`, and in that case we just try to delete the PVC and if it doesn't exist we get a 404 and log the warning.

I guess in our case the point is I don't want to see the warnings when users get culled (which happens after 5 days of inactivity for us) so to avoid that we just need to set `delete_pvc=False`.

consideRatio: I extracted this to https://github.com/jupyterhub/kubespawner/issues/543 to avoid loosing track of this comment!

PR #476, opened 2021-01-21
Allow mounting of service account token to be configurable (automount_service_account_token)

dtaniwaki:
I don't think we always need a service account token even if we set a custom service account. To minimize a security risk by just setting a service account and make the behavior consistent with the default case, we should disable it by default and ask to modify it by `modify_pod_hook` manually although it's kind of breaking change.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -467,6 +467,8 @@ def make_pod(
 
     if automount_service_account_token is None:
         if service_account is None:
+            # This makes sure that we don't accidentally give access to the whole
+            # kubernetes API to the users in the spawned pods.
             pod.spec.automount_service_account_token = False
     else:
         pod.spec.automount_service_account_token = automount_service_account_token
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -417,11 +417,11 @@ def make_pod(
         security_context=container_security_context,
     )
 
-    if service_account is None:
-        # This makes sure that we don't accidentally give access to the whole
-        # kubernetes API to the users in the spawned pods.
-        pod.spec.automount_service_account_token = False
-    else:
+    # This makes sure that we don't accidentally give access to the whole
+    # kubernetes API to the users in the spawned pods.
+    pod.spec.automount_service_account_token = False
+
+    if service_account is not None:
         pod.spec.service_account_name = service_account
 
     notebook_container.resources.requests = {}
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -78,6 +78,7 @@ def make_pod(
     lifecycle_hooks=None,
     init_containers=None,
     service_account=None,
+    automount_service_account_token=False,
     extra_container_config=None,
     extra_pod_config=None,
     extra_containers=None,
@@ -417,10 +418,7 @@ def make_pod(
         security_context=container_security_context,
     )
 
-    # This makes sure that we don't accidentally give access to the whole
-    # kubernetes API to the users in the spawned pods.
-    pod.spec.automount_service_account_token = False
-
+    pod.spec.automount_service_account_token = automount_service_account_token
     if service_account is not None:
         pod.spec.service_account_name = service_account
 
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -80,7 +80,7 @@ def make_pod(
     lifecycle_hooks=None,
     init_containers=None,
     service_account=None,
-    automount_service_account_token=False,
+    automount_service_account_token=None,
     extra_container_config=None,
     extra_pod_config=None,
     extra_containers=None,
@@ -462,9 +462,14 @@ def make_pod(
         security_context=csc,
     )
 
-    pod.spec.automount_service_account_token = automount_service_account_token
     if service_account is not None:
         pod.spec.service_account_name = service_account
+
+    if automount_service_account_token is None:
+        if service_account is None:
+            pod.spec.automount_service_account_token = False
+    else:
+        pod.spec.automount_service_account_token = automount_service_account_token
 
     notebook_container.resources.requests = {}
     if cpu_guarantee:
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -404,12 +404,14 @@ class KubeSpawner(Spawner):
     )
 
     automount_service_account_token = Bool(
-        False,
+        None,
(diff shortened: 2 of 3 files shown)


dtaniwaki:
I exclude labextensions from the lint target as it's a git submodule.

dtaniwaki:
I sent a PR to the upstream for no automount by default.
https://github.pfidev.jp/chem/eneos-jupyterhub/issues/97

consideRatio:
I think I agree with this change, @minrk @manics @yuvipanda what do you think?

I figure that when a service account is specified, it may very well only be used only on a k8s level in many cases. So, for the same reasons we have it not mounted by default when the default service account is used, I think we should not mount it by default when it is specified also.

yuvipanda:
I'm trying to think of cases where a service account is specified but token isn't mounted. If there's no token, you can't authenticate as that service account, right? My understanding is that those give you the credentials to authenticate to the API. What are the use cases where you have a service account assigned but the credentials not mounted?

consideRatio:
To specify a k8s SA but not mount the SA can be relevant when a pod security policy is having a say about pods in the namespace. I have not used them myself yet but as i understand the pod can be denied scheduling or similar without a SA with permissions connected to it.

Another case is to attach imagePullSecrets to a k8s SA rather than the pod, but this isnt very favored practice if i recall k8s docs correctly.

@dtaniwaki what is your use case of specifying a SA not to be used from within the container?

manics:
Inspired by https://discourse.jupyter.org/t/request-for-implementation-mapping-notebook-user-identities-to-cloud-provider-identities/7637
in future I can envisage people wanting to allow their singleuser servers to access K8S functionality using a restricted service account, so it'd be useful to hear the current use-case for preventing it.

dtaniwaki:
@consideRatio We don't want to use the `default` service account because there're pods with other roles, so we use a service account of `jupyter` for notebook pods. Currently, we don't give any permission except PSP use, but we're not sure if we want to do something w/ the k8s API on a sidecar which will use the same service account.

consideRatio:
If we merge this, I think we should also expose configuration to mount it. Then our changelog could be saying "if you were specifying a pod service account, you now need to explicitly request it to be mounted on the container if you want that by `one line config here`."

yuvipanda:
> I think the main drawback as I see it, is that it is unexpected that the service account token isn't available after having specified a service account, and that it isn't obvious to me where to log it that makes sense. Should log a note about mounting not mounting a sa whenever a user starts to help mitigate this? Hmmm....

IMO, this is a big drawback + breaking change. PSP satisfying use cases aren't as common as folks wanting credentials to talk to the API. I think we should expose this as an option, but not set it as the default.

dtaniwaki:
I’m a little bit confused. Is PSP relevant to SA token in containers? I thought PSP permission check for the SA is done before mounting the service account on containers. Do I understand it incorrectly?

yuvipanda:
> Currently, we don't give any permission to the service account except PSP use, but we're not sure if we will want to do something w/ the k8s API on a sidecar which will use the same service account.

Oh, I was mostly responding to this ^. I don't actually know how PSPs and SAs interact :D

dtaniwaki:
@yuvipanda I updated the code so the automount flag can be configured.

yuvipanda:
This looks good to me, but I think it should default to `True` rather than `False`.

dtaniwaki:
I guess most of the admins don't set the `service_account` value, which is `None`, then the token is currently not mounted. Do we want to change the default for `service_account = None` case, or `service_account != None` case? Which has more risks?

dtaniwaki:
Or, maybe,

```
if service_account is None:
    pod.automount_service_account_token = False
else:
    pod.automount_service_account_token = config.automount_service_account_token
```

yuvipanda:
> Or, maybe, we can do something as below.
> 
> ```
> if service_account is None:
>     pod.automount_service_account_token = False
> else:
>     pod.automount_service_account_token = automount_service_account_token
> ```

This sounds good to me!

dtaniwaki:
@consideRatio How about you?

consideRatio:
I'm okay with the idea to retain the current kubespawner behavior to set `automount_service_account_token` to False if a no `service_account` was passed.

If `automount_service_account_token` is exposed, which I strongly support no matter what the default is, then I think we should make it default to `None` and only impose KubeSpawner's custom default if it isn't explicitly set to False or True by the user.

This means, that this logic below needs to be revised a bit.

```python
if service_account is None:
    pod.automount_service_account_token = False
else:
    pod.automount_service_account_token = automount_service_account_token
```

So, perhaps something like this instead?

```python
if automount_service_account_token is None:
    if service_account is None:
        pod.automount_service_account_token = False
else:
    pod.automount_service_account_token = automount_service_account_token
```

dtaniwaki:
Maybe,

```python
if service_account is not None:
        pod.spec.service_account_name = service_account

if automount_service_account_token is None:
    if service_account is None:
        pod.automount_service_account_token = False
else:
    pod.automount_service_account_token = automount_service_account_token
```

kubespawner/objects.py
```diff
@@ -80,6 +80,7 @@ def make_pod(
     lifecycle_hooks=None,
     init_containers=None,
     service_account=None,
+    automount_service_account_token=False,
```
consideRatio, line 83: Default to None here assuming our passed configuration can be None, False, and True.

kubespawner/spawner.py
```diff
@@ -394,10 +394,22 @@ def _namespace_default(self):
...
+        This makes sure that we don't accidentally give access to the whole
+        kubernetes API to the users in the spawned pods.
+        Set automount_service_account_token True to mount it.
 
         This `serviceaccount` must already exist in the namespace the user pod is being spawned in.
+        """,
+    )
+
+    automount_service_account_token = Bool(
+        False,
```
consideRatio, line 407:
```suggestion
        None,
        allow_none=True,
```

consideRatio (approved):
This LGTM! Thank you so much @dtaniwaki for your thorough work on this! :heart: :tada:

@yuvipanda smash merge?

meeseeksmachine:

https://discourse.jupyter.org/t/notebook-server-can-not-access-the-k8s-api-from-a-pod/8042/2

yuvipanda:
Thanks a lot, @dtaniwaki and @consideRatio!

PR #477, opened 2021-01-28
Add options_from_form as configurable

cbanek:
Without this part as being configurable, it doesn't make sense to
be able to configure the options_form part since you can't easily
get the values back to permute them into options the kubespawner
will understand. I think this meant to be decorated since it is mentioned
in the documentation in the options_form section.

Related to #452, jupyterhub/jupyterhub#3225

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2506,7 +2506,11 @@ class KubeSpawner(Spawner):
         else:
             return self._render_options_form(self.profile_list)
 
-    def options_from_form(self, formdata):
+    @default('options_from_form')
+    def _options_from_form_default(self):
+        return self._options_from_form
+
+    def _options_from_form(self, formdata):
         """get the option selected by the user on the form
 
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
This LGTM! Below is my review done aloud for transparency!

Okay jupyterhub/jupyterhub#3225 made the `options_from_form` a configurable Callable traitlet configuration. And hmmm KubeSpawner wants to have its own default for that it has previously also provided a `options_from_form` function.

Ah, and this PR aligns with JupyterHub's base Spawner class by respecting the configurable value if set, while just updating the default value for the configurable traitlet if the traitlet isn't set.

This LGTM!

consideRatio:
Thank you @cbanek! :tada: :heart:!

jabbera:
Can we get a release cut with this change please?

jabbera:
@consideRatio I don't mean to bug you but I realized I never tagged you in my ask:-)

Issue #478, opened 2021-02-02
Set fsGroupChangePolicy in Kubespawner


cyrilcros:
### Proposed change

One of the new beta feature in kubernetes 1.20 is the new `fsGroupChangePolicy` in a `securityContext`. See https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods That's a beta feature in 1.20, not alpha as the old text says.
It avoids the lengthy process of "chown'ing" a volume which can be very slow when images have lot of files (ie tensorflow etc...) - see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1252 This is especially true if the docker-stacks images people use all have the right permissions already set.

It would be nice to make this option a default if it works with version below 1.20...

That could also alternatively done if the securityContext configuration was exposed - ie https://github.com/jupyterhub/kubespawner/issues/454 was implemented.

It is present in the kubernetes Python client library v1.18 and up since it was in alpha then.

consideRatio:
`fsGroupChangePolicy` will be made configurable as part of the `pod_security_context` introduced in #480, so that part is a duplicate of #454.

__Let `fsGroupChangePolicy` default to `"OnRootMismatch"`?__

I don't think KubeSpawner should introduce a default value for such a significant change. This would influence all containers of the pod, and while I think it could be a nice default, it would be going against the k8s default which is something that I think is very problematic.

If we deviate from k8s defaults in KubeSpawner regarding this:

1. We would probably help a group of KubeSpawner users by providing a higher performance default
2. We would also introduce an opinionated choice that few will be aware of, which would lead to an unexpected outcome that can cause these users spend significant time debugging etc.

Due to this, I'm quite strongly opinionated against letting KubeSpawner become opinionated about this, less opinionated about the JupyterHub Helm chart becoming opinionated about it, but think the way to go is to provide documentation in z2jh.jupyter.org associated with the JupyterHub Helm chart regarding this as a choice.

elferherrera:
Is `fsGroupChangePolicy` configurable? I cannot find the setting modify this value

consideRatio:
This was not implemented i think, use pod_security_context

elferherrera:
is there an example of `pod_security_context` usage?

elferherrera:
Sorry. I think I found it here
[...]

Issue #454, opened 2020-10-25
Expose traitlets to configure the pod's and the container's securityContext


consideRatio:
Currently, the container's securityContext and the pod's securityContext are partially set if a uid/gid etc is set, but I think we should expose a way to set them directly which can merge with uid/gid etc.

An option to exposing this kind of configuration is to set the securityContext using `extra_pod_config` or `extra_container_config`.

PR #479, opened 2021-02-02
set fsGroupChangePolicy to always be OnRootMismatch

cyrilcros:
Fixes https://github.com/jupyterhub/kubespawner/issues/478 

I am not sure what would happen to versions below 1.18 / 1.19. In theory that would speed up mounting larger docker-stacks image. I have not checked what happens on older versions, I am hoping the CI can check it....

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -372,6 +372,7 @@ def make_pod(
         pod_security_context.supplemental_groups = [
             int(gid) for gid in supplemental_gids
         ]
+    pod_security_context.fs_group_change_policy = "OnRootMismatch"
     # Only clutter pod spec with actual content
     if not all([e is None for e in pod_security_context.to_dict().values()]):
         pod.spec.security_context = pod_security_context


consideRatio:
If this is non-breaking to use in k8s versions before 1.20, I'm open to having it as the default.

I suggest that we instead of hardcoding it though, that we make it a default value within a configurable dictionary closing https://github.com/jupyterhub/kubespawner/issues/454 as well.

By having solved https://github.com/jupyterhub/kubespawner/issues/454 first, we ensure this can always be mitigated if there are issues in some k8s clusters.

cyrilcros:
I can help with #454 , but I want to point out that in the relevant paragraph 
https://github.com/jupyterhub/kubespawner/blob/c215e24c3816e06b87409a50ffef2125465d1c44/kubespawner/objects.py#L379
The option names have diverged in Jupyterhub vs Kubernetes (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podsecuritycontext-v1-core). You can't go to the Jupyterhub Helm chart, paste a valid Kubernetes securityContext and hope it will be respected. You can align Jupyterhub to Kubernetes but that breaks users'profile changes...

Alternatively I can modify the pull request to explicitly define a new `fsGroupChangePolicy` with the default being `Always` like in Kubernetes. We then fix #454 with multiple separate pull requests, for the `seLinuxOptions` for example.

cyrilcros:
Do not merge for now as I am counting on #480 to set this.

consideRatio:
This can now be configured manually even though it won't be the default value. I discussed why I think we shouldn't make it the default value in https://github.com/jupyterhub/kubespawner/issues/478#issuecomment-775890018.

PR #480, opened 2021-02-03
Add pod_security_context and container_security_context config

### LINKED ISSUE #454: Expose traitlets to configure the pod's and the container's securityContext

cyrilcros:
This PR adds `container_security_context` and `pod_security_context` configuration options.

__Old vs new options__
The new options `container_security_context` and `pod_security_context` updates the partial security contexts built from the following older options:

- `supplemental_gids` (pod)
- `fs_gid` (pod)
- `privileged` (container)
- `allow_privilege_escalation` (container)
- `uid` (container)
- `gid` (container).

__Related__
Fixes #454 - pod/container security contexts can now be configured
Related to #478 - exposes fsGroupChangePolicy through pod security context, but doesn't provide a default value.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -397,14 +397,14 @@ def make_pod(
     if fs_gid is not None:
         psc["fs_group"] = int(fs_gid)
     if supplemental_gids is not None:
-        psc["supplemental_groups"] = [
-            int(gid) for gid in supplemental_gids
-        ]
+        psc["supplemental_groups"] = [int(gid) for gid in supplemental_gids]
     if pod_security_context is not None:
         psc.update(pod_security_context)
     try:
         # validate truthy config by casting it
-        pod.spec.security_context = get_k8s_model(V1PodSecurityContext, psc) if psc else None
+        pod.spec.security_context = (
+            get_k8s_model(V1PodSecurityContext, psc) if psc else None
+        )
     except TypeError as err:
         print("Not a valid security context: ", str(psc))
         print("You supplied: ", str(pod_security_context))
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -376,14 +376,18 @@ def make_pod(
     # ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podsecuritycontext-v1-core (pod)
     pod_context_dict = {}
     # populate with fs_gid  / supplemental_gids
-    pod_context_dict["fs_group"] = int(fs_gid) if fs_gid else None
-    pod_context_dict["supplemental_groups"] = (
-        [int(gid) for gid in supplemental_gids] if supplemental_gids else None
-    )
+    if fs_gid is not None:
+        pod_context_dict["fs_group"] = int(fs_gid)
+    if supplemental_gids is not None:
+        pod_context_dict["supplemental_groups"] = [
+            int(gid) for gid in supplemental_gids
+        ]
     if pod_security_context is not None:
         pod_context_dict.update(pod_security_context)
     try:
-        v1_pod_security_context = V1PodSecurityContext(**pod_context_dict)
+        # only clutter if actual contents
+        if pod_context_dict:
+            pod.spec.security_context = V1PodSecurityContext(**pod_context_dict)
     except TypeError as err:
         print("Not a valid security context: ", str(pod_context_dict))
         print("You supplied: ", str(pod_security_context))
@@ -391,23 +395,26 @@ def make_pod(
             """Do check if your Kubernetes versions supports the relevant options, as well as the kubernetes-client library"""
         )
         raise
-    # Only clutter pod spec with actual content
-    if not all([e is None for e in v1_pod_security_context.to_dict().values()]):
-        pod.spec.security_context = v1_pod_security_context
 
     container_context_dict = {}
     # populate with uid / gid  / privileged
-    container_context_dict["run_as_user"] = int(run_as_uid) if run_as_uid else None
-    container_context_dict["run_as_group"] = int(run_as_gid) if run_as_gid else None
-    # requires None for the 'silent' default behavior, not a boolean
-    container_context_dict["privileged"] = None if not run_privileged else True
-    container_context_dict["allow_privilege_escalation"] = (
-        None if allow_privilege_escalation else False
-    )
+    if run_as_uid is not None:
+        container_context_dict["run_as_user"] = int(run_as_uid)
+    if run_as_gid is not None:
+        container_context_dict["run_as_group"] = int(run_as_gid)
+    if run_privileged:  # false as default
+        container_context_dict["privileged"] = True
+    if not allow_privilege_escalation:  # true as default
+        container_context_dict["allow_privilege_escalation"] = False
     if container_security_context is not None:
         container_context_dict.update(container_security_context)
     try:
-        v1_container_security_context = V1SecurityContext(**container_context_dict)
+        # only clutter if actual contents
+        v1_container_security_context = (
+            V1SecurityContext(**container_context_dict)
+            if container_context_dict
+            else None
+        )
     except TypeError as err:
         print("Not a valid security context: ", str(container_context_dict))
         print("You supplied: ", str(container_security_context))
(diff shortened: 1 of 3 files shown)


cyrilcros:
@consideRatio OK this should be ready for review.

consideRatio:
@cyrilcros I just saw your edit of splitting into pod_security_context and container_security_context and I like it!

I suggest you let whatever is set there override for example fs_gid if there is a conflict as well. I'm not 100% my motivation for this, but I think it makes sense related to us not having defaults here yet, and if someone specifies this it is with intent.

__Pod specific__
- [x] supplementalGroups (`supplemental_gids`)
- [x] fsGroup (`fs_gid`)
- [ ] fsGroupChangePolicy
- [ ] sysctls

__Container specific__
- [x] privileged (`privileged`)
- [x] allowPrivilegeEscalation (`allow_privilege_escalation`)
- [ ] capabilities
- [ ] procMount
- [ ] readOnlyRootFilesystem

__Common__
- [x] runAsUser (`uid)
- [x] runAsGroup (`gid`)
- [ ] runAsNonRoot
- [ ] seLinuxOptions
- [ ] seccompProfile
- [ ] windowsOptions

---

I hope to find time to review this within a week, but my attention is too split at the moment. I should really be doing something else right now as well. Thanks for your work @cyrilcros! :heart:

kubespawner/objects.py
```diff
@@ -357,6 +366,9 @@ def make_pod(
     # The security settings that you specify for a Pod apply to all Containers
     # in the Pod, but settings on the container level can override them.
     #
+    # We also merge any valid options defined in security_context in
+    # each of those. Options that are not valid are ignored.
+    #
```
consideRatio, line 371: I don't understand the merging comment, but suggest that the logic is:

1. We bootstrap the pod and container security context:
  - Pod: `supplemental_gids`, `fs_gid`
  - Container: `privileged`, `allow_privilege_escalation`
  - (?) set `uid` and `gid`  wherever they were set before, was it on the pod or container level?
2. We let the explicit pod and container_security_context merge into the bootstrapped ones, giving priority to whatever is in the pod_/container_ security context.

Regarding ignoring, I'm not sure what makes most sense, but I'm afraid about plain ignore as I really dislike silent failures that show up during runtime. I'd prefer hard error if its reliable to not fail because a new API provides something more than before, but otherwise warning seems second best, and ignore as the least attractive option.

consideRatio: Does V1PodSecurityContext and V1SecurityContext contain all the entries as I listed before? If so then I suggest we fail hard if something not allowed is provided.

__Pod specific__
- [ ] supplementalGroups
- [ ] fsGroup
- [ ] fsGroupChangePolicy
- [ ] sysctls

__Container specific__
- [ ] privileged
- [ ] allowPrivilegeEscalation
- [ ] capabilities
- [ ] procMount
- [ ] readOnlyRootFilesystem

__Common__
- [ ] runAsUser
- [ ] runAsGroup
- [ ] runAsNonRoot
- [ ] seLinuxOptions
- [ ] seccompProfile
- [ ] windowsOptions

tests/test_objects.py
```diff
@@ -434,6 +434,143 @@ def test_allow_privilege_escalation_container():
...
+                }
+            ],
+            'restartPolicy': 'OnFailure',
+            'volumes': [],
+        },
+        "kind": "Pod",
+        "apiVersion": "v1",
+    }
+
+
```
consideRatio, line 573: :heart: thanks sooooo much for writing tests!!!!!!!!!!!!!!

kubespawner/objects.py
cyrilcros: - The merging comment is a leftover from my previous attempt at making a unifed `security_context`, disregard it.
- I will swap around my code to let the security context take precedence. I am not super sure about it because I believe docker-stacks use a 100 `gid` which you likely would have to add explicitly.
- For ignoring, the kubernetes-client library should raise a TypeError if you supply something unexpected. I can catch it and raise something else, I am not sure what works best.
- The tests are 99% copy-pasted (but correct hopefully).

consideRatio: > I will swap around my code to let the security context take precedence. I am not super sure about it because I believe docker-stacks use a 1000 uid / 100 gid which you likely would have to add explicitly.

Hmmm, I don't understand what you consider very well here. I think in this version of kubespawner the default is to not set uid/gid unless explicitly configured on KubeSpawner. And, if k8s gets a pod created without runAsUser or runAsGroup, I think the default depends on the k8s cluster on how it behaves, which makes it something you want to explicitly set.

With the `uid` configuration as we already have, and the new container_security_context.run_as_user, there will be two separate ways to configure it explicitly, where the latter takes precedence.

The JupyterHub helm chart sets KubeSpawners `uid` / `gid` to 1000 / 100 respectively by default btw.

consideRatio: > For ignoring, the kubernetes-client library should raise a TypeError if you supply something unexpected. I can catch it and raise something else, I am not sure what works best.

As long as we raise an error, and know that "something unexpected" won't include the API that was listed in k8s 1.20 specifications at least. I know that the python kubernetes client is quite far behind the release schedule, so it would be good to verify it can actually accept for example `seccompProfile` etc.

cyrilcros:
At this point you will get a TypeError if you supply an illegal argument, which could be also due to something like `fsGroupChangePolicy` running on an earlier Kubernetes version (<1.18 for alpha).

consideRatio:
@cyrilcros thank you for your work on this!!! :Tada:! 🥳

consideRatio:
@cyrilcros I'm actively working to get this PR to get merged but I've decided to wait for #483 to get merged before I finish some parts of this PR. Is it okay that I push some commits here?

cyrilcros:
@consideRatio Sure thing.

consideRatio:
@cyrilcros thanks for your work on this :tada: :heart:!

I consider this ready for review/merge @minrk @manics if you have time!

kubespawner/objects.py
```diff
@@ -353,41 +362,74 @@ def make_pod(
...
-    if not all([e is None for e in pod_security_context.to_dict().values()]):
-        pod.spec.security_context = pod_security_context
+        psc["fsGroup"] = int(fs_gid)
+    if supplemental_gids is not None:
+        psc["supplementalGroups"] = [int(gid) for gid in supplemental_gids]
+    if pod_security_context is not None:
+        if any("_" in key for key in pod_security_context.keys()):
+            raise ValueError(
+                "pod_security_context's keys should be directly k8s compliant camelCase, but snake_case formatting was detected."
+            )
```
minrk, line 409: When a value is wrong, it's a good idea to show that value so it's easier to find and fix. For example:

```suggestion
        for key in pod_security_context.keys():
            if "_" in key:
                raise ValueError(
                    f"pod_security_context's keys should have k8s camelCase names, got '{key}'"
                )
```

kubespawner/objects.py
```diff
@@ -353,41 +362,74 @@ def make_pod(
...
-    # Only clutter container spec with actual content
-    if all([e is None for e in container_security_context.to_dict().values()]):
-        container_security_context = None
+        csc["runAsGroup"] = int(gid)
+    if privileged:  # false as default
+        csc["privileged"] = True
+    if not allow_privilege_escalation:  # true as default
+        csc["allowPrivilegeEscalation"] = False
+    if container_security_context is not None:
+        if any("_" in key for key in container_security_context.keys()):
```
minrk, line 426: same here

minrk (commented):
Some minor feedback on improving the error messages, but otherwise looks great to me. Thanks!

consideRatio:
Thanks for the review @minrk! Resolved by 4661af8

minrk (approved):
Nice!

consideRatio:
@cyrilcros thank you so much for your work on this!!! :heart: :tada:

meeseeksmachine:

https://discourse.jupyter.org/t/singleuser-pod-fails-after-1-0-0-upgrade-on-supplementarygroups-with-psp/9660/1

PR #481, opened 2021-02-05
Fix failure to create a PVC being logged as failure to create a Pod

mriedem:
Looks like this was just a copy/paste error.


consideRatio:
Thanks @mriedem!

Issue #492, opened 2021-03-04
Expose Kubernetes client configuration options


kafonek:
### Proposed change
Expose a way to inject configuration options for the `kubespawner.KubeSpawner.api` object.  Right now it appears to be instantiated without passing in args or kwargs (`self.api = shared_client('CoreV1Api')`).  

### Alternative options
Some configuration options can be set via environment variables, but others cannot.  For instance, `kubernetes.client.Configuration.get_default_copy().ssl_ca_cert`

### Who would use this feature?
Kubernetes clusters that use custom CA certs may be misconfigured such that the ca-certs provided to service accounts (`/var/run/secrets/kubernetes.io/serviceaccounts/ca.crt` inside the Hub pod) do not work with the Kubernetes API server.  When Kubespawner attempts to initialize and list pods in the namespace, it raises a `CERTIFICATE_VERIFY_FAILED` error.  

See https://discourse.jupyter.org/t/enabling-jh-to-use-an-additional-ca-cert-when-calling-private-k8s-api/6662 and discussions in Jupyerhub Gitter between me and @consideRatio on Mar 1 / Mar 2 2021.

It is worth noting that there's probably a cluster administration solution to this problem.  I do not know what needs to change on the cluster configuration side to provision working `ca.crt` files to the serviceaccount / accompany secrets.  Either way, I think offering config options in Kubespawner and zero-to-jupyterhub `config.yaml` is worthwhile.

### (Optional): Suggest a solution
I defer to @yuvipanda @consideRatio on how to implement this.  From a user perspective, I would like to be able to mount my actual CA certs into the Pod and then tell KubeSpawner to use them either via ENV variables or as a config option.  Imagine I have a `ca-certs` Secret in my namespace with a `ca.crt` file that includes the k8s API ca's, and I'm deploying jupyterhub with the zero-to-jupyterhub-k8s helm chart.

```
# config.yaml
hub:
  extraVolumes:
    - name: ca-crt
      secret:
        secretName: ca-certs
  extraVolumeMounts:
    - name: ca-crt
      mountPath: /etc/pki/certs/ca-certificate.crt
      subPath: ca.crt
  # NEW SYNTAX (?)
  kubeSpawner:
    ssl_ca_cert: /etc/pki/certs/ca-certificate.crt
  # ALTERNATIVELY
  extraEnv:
    KUBESPAWNER_CA_CERT: /etc/pki/certs/ca-certificate.crt
```

```
# kubespanwer.clients.py
[...]

PR #494, opened 2021-03-09
Fix 492 -- config kubernetes client ssl_ca_cert

kafonek:
This is the first draft to fix https://github.com/jupyterhub/kubespawner/issues/492.  

- [x] Kubernetes client ssl_ca_cert configured with traitlets
- [x] Also included conf option for kubernetes client host
- [x] Documentation in the form of traitlet `help` strings
- [x] System tested, working in our cluster with z2jh (mounted root CA and `hub.config.KubeSpawner.k8s_api_ssl_ca_cert` in config.yaml)
- [x] `pytest`  60 passed, 12 skipped, 16 warnings on my dev environment
- [x] Adding this config logic to proxy.py -- @consideRatio you mentioned that was sort of a "parallel" effort to spawner.py but in the same project.  Does this need to be in there too?
- [x] Opinionated change -- calling `shared_client` from spawner.py before it is called in reflector.py.  I personally think that this is more intuitive, but totally understand if you want to keep it in reflector.py for historic/maintenance reasons I have missed.  
- [ ] Documentation in z2jh for mounting root CA's and setting `k8s_api_ssl_ca_cert` in config.yaml

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -254,7 +254,7 @@ class KubeSpawner(Spawner):
         In rare non-standard cases, such as using custom intermediate CA
         for your cluster, you may need to mount root CA's elsewhere in
         your Pod/Container and point this variable to that filepath
-        """
+        """,
     )
 
     k8s_api_host = Unicode(
@@ -265,7 +265,7 @@ class KubeSpawner(Spawner):
         
         Typically this is unnecessary, the hostname is picked up by 
         config.load_incluster_config() or config.load_kube_config.
-        """
+        """,
     )
 
     k8s_api_threadpool_workers = Integer(
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -23,6 +23,7 @@ from jupyterhub.traitlets import Command
 from jupyterhub.utils import exponential_backoff
 from kubernetes import client
 from kubernetes.client.rest import ApiException
+import kubernetes.config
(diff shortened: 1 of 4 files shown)


consideRatio (approved):
This LGTM, thank you for the thorough work @kafonek and thank you for bearing with the lengthy discussion.

consideRatio:
@manics what do you think?

kubespawner/spawner.py
```diff
@@ -198,14 +199,17 @@ def __init__(self, *args, **kwargs):
                     max_workers=self.k8s_api_threadpool_workers
                 )
 
+            # Set global kubernetes client configurations
+            # before reflector.py code runs
+            self._set_k8s_client_configuration()
+            self.api = shared_client('CoreV1Api')
```
manics, line 205: Since this is changing global state it might be worth adding a comment in https://github.com/jupyterhub/kubespawner/blob/c215e24c3816e06b87409a50ffef2125465d1c44/kubespawner/clients.py#L24 ?

manics (commented):
I've made a suggestion about improving commenting since this is changing global state which could be confusing to any newcomers to the code base. I don't feel able to review the change though.

consideRatio:
I've looked through this again after some time and it still LGTM so I think we should go for a merge at this point! Thank you for your thorough work on this @kafonek, I greatly appreciate your effort! :tada: :heart:

PR #499, opened 2021-04-06
Fix self.user_namespace setup bug and remove NS creation

galbarnissan:


consideRatio:
Hi @galbarnissan! Any suggested change must be quite well motivated to be accepted.

As this PR does not contain a description and suggests makes a breaking change, I'll go ahead and close it for now.

Perhaps a suitable path forward based on the topic, would be to open a bug report if you think it is a bug in the behavior of KubeSpawner you have observed. Separate that discussion from removing k8s namespace creation as an opt-in feature though.

Issue #493, opened 2021-03-09
Is there a bug?cmd, args, notebook_dir, default_url


consideRatio:
- `cmd`, `args`, `notebook_dir`, `default_url`, and `get_args(...)` are defined in in the Spawner base class
- `get_args(...)` combines `notebook_dir`, `default_url`, and `args` into a single appendix to `cmd`
- K8s containers have `command` and `args` which are the equivalent of Dockerfiles `entrypoint` and `cmd`
- KubeSpawner sets the containers k8s field `args` to the variable `real_cmd` decided below, which changed in 5ee6dc6f.

```diff
    async def get_pod_manifest(self):
        # ...
        if self.cmd:
            real_cmd = self.cmd + self.get_args()
        else:
            # change commit comment:
            # fix use of get_args() (calling get_args() alone would omit the command)
-           real_cmd = self.get_args()
+           real_cmd = None
```

behrica:
I am trying to launch a rstudio image with this, and I am very much struggling that I can not "completely control" the `command` and `args` passed to starting the container

manics:
Do you think we can resolve this issue before Z2JH 1.0.0 is released? Either by adding docs and warnings, or perhaps by adding a new `entrypoint` parameter, and making it clear that `entrypoint` and `cmd` have their Docker semantics.

Although this is a K8s project I think it's fine to use the Docker terminology, especially as it's also the [Open Container Initiative terminology](https://github.com/opencontainers/image-spec/blob/master/config.md). What's more important is that we make it clear and use consistent names throughout.

consideRatio:
I agree that we must be consistent, and in this case we must be consistent against the `Spawner` base class that is k8s unaware, so then adding `entrypoint` would make sense alongside `cmd`.

We would then document that `cmd` and `entrypoint` refers to the Dockerfile names, and that we will set the k8s Pod's container's `args` and `command` section when we set `cmd` and `entrypoint` in KubeSpawner.

To conclude: :+1: for adding `entrypoint` as a KubeSpawner traitlet to configure the containers `command`.

mapsacosta:
@consideRatio is there going to be a separate fix for `c.KubeSpawner.default_url `?

manics:
@mapsacosta if you set `cmd` the other args such as `default_url` should be passed. If it's not working can you show us your full configuration?

consideRatio:
## Update

In the initial post I wrote investigative conclusions, and I want to give an update about those two points.

[...]

PR #502, opened 2021-04-30
Rewrite help for Kubespawner.cmd

manics:
Whilst working on https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2138 I noticed the current description for `cmd` is mostly copied from the parent class and is a bit confusing. I've attempted to simplify it.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -408,9 +408,9 @@ class KubeSpawner(Spawner):
         Either
           - a string containing a single command or path to a startup script
           - a list of the command and arguments
-          - `None` (default) to use the Docker image's `Cmd`
+          - `None` (default) to use the Docker image's `CMD`
 
-        `cmd` will be passed to the image's `Entrypoint`.
+        `cmd` will be passed to the image's `ENTRYPOINT`.
         """,
     )
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -403,18 +403,14 @@ class KubeSpawner(Spawner):
         minlen=0,
         config=True,
         help="""
-        The command used for starting the single-user server.
+        The command used to start the single-user server.
 
-        Provide either a string or a list containing the path to the startup script command. Extra arguments,
-        other than this path, should be provided via `args`.
+        Either
+          - a string containing a single command or path to a startup script
+          - a list of the command and arguments
+          - `None` (default) to use the Docker image's `Cmd`
 
-        This is usually set if you want to start the single-user server in a different python
-        environment (with virtualenv/conda) than JupyterHub itself.
-
-        Some spawners allow shell-style expansion here, allowing you to use environment variables.
-        Most, including the default, do not. Consult the documentation for your spawner to verify!
-
-        If set to `None`, Kubernetes will start the `CMD` that is specified in the Docker image being started.
+        `cmd` will be passed to the image's `Entrypoint`.
         """,
     )
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -410,7 +410,7 @@ class KubeSpawner(Spawner):
           - a list of the command and arguments
           - `None` (default) to use the Docker image's `CMD`
 
-        `cmd` will be passed to the image's `ENTRYPOINT`.
+        If `cmd` is set, it will be augmented with `spawner.get_args(). This will override the `CMD` specified in the Docker image.
         """,
     )
 


kubespawner/spawner.py
```diff
@@ -403,18 +403,14 @@ def _namespace_default(self):
...
         help="""
-        The command used for starting the single-user server.
+        The command used to start the single-user server.
 
-        Provide either a string or a list containing the path to the startup script command. Extra arguments,
-        other than this path, should be provided via `args`.
+        Either
+          - a string containing a single command or path to a startup script
+          - a list of the command and arguments
+          - `None` (default) to use the Docker image's `Cmd`
```
consideRatio, line 411:
```suggestion
          - `None` (default) to use the Docker image's `CMD`
```

kubespawner/spawner.py
```diff
@@ -403,18 +403,14 @@ def _namespace_default(self):
...
 
-        This is usually set if you want to start the single-user server in a different python
-        environment (with virtualenv/conda) than JupyterHub itself.
-
-        Some spawners allow shell-style expansion here, allowing you to use environment variables.
-        Most, including the default, do not. Consult the documentation for your spawner to verify!
-
-        If set to `None`, Kubernetes will start the `CMD` that is specified in the Docker image being started.
+        `cmd` will be passed to the image's `Entrypoint`.
```
consideRatio, line 413:
```suggestion
        `cmd` will be passed to the image's `ENTRYPOINT`.
```

manics:
FYI I picked the original `Cmd`/`Entrypoint` casing from https://github.com/opencontainers/image-spec/blob/master/config.md

kubespawner/spawner.py
```diff
@@ -403,18 +403,14 @@ def _namespace_default(self):
...
+          - `None` (default) to use the Docker image's `CMD`
 
-        This is usually set if you want to start the single-user server in a different python
-        environment (with virtualenv/conda) than JupyterHub itself.
-
-        Some spawners allow shell-style expansion here, allowing you to use environment variables.
-        Most, including the default, do not. Consult the documentation for your spawner to verify!
-
-        If set to `None`, Kubernetes will start the `CMD` that is specified in the Docker image being started.
+        `cmd` will be passed to the image's `ENTRYPOINT`.
```
consideRatio, line 413: I think this is a problematic statement for two reasons.

1. It only influence the k8s Pod rather than the Dockerfile or similarly.
2. It actually configures the k8s container's `args` (like Dockerfile's CMD) rather than `command` (like Dockerfile's ENTRYPOINT).

A related confusion is described in https://github.com/jupyterhub/kubespawner/issues/493.

I think we can make small progressions and I find this change PR mergable if we just update this to reference that it implies setting the k8s Pod's container's `args` where the `command` won't be set. In other words, any Dockerfile's ENTRYPOINT will always be respected. Here is [related documentation about that](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes).

consideRatio: THANK YOU FOR YOUR WORK ON THIS @manics!! :heart:

I forgot to convey that part before!

manics, line 413: Is this what you mean?
```suggestion
        If `cmd` is set it will be passed to the image's `ENTRYPOINT` along with any additional arguments.
```

consideRatio:
```suggestion
        If `cmd` is set, it will be augmented with `spawner.get_args()` and passed to the spawned k8s Pod's container's `args` field that overrides the Dockerfile's `CMD` section. This will leave the Dockerfile's `ENTRYPOINT` section intact, and there is currently no way to specify the k8s Pod's Container's `command` which would override the Dockerfile's `ENTRYPOINT`.
```

I think something like this. This is a messy story but it is an accurate representation of the current situation.  `get_args()` is [defined in the JupyterHub Spawner base class](https://github.com/jupyterhub/jupyterhub/blob/3fec19d191119e4203c243f44c43e74c8fe62f04/jupyterhub/spawner.py#L995).

manics: Instead of bringing in the internal K8s details how about keep it simple? 
```suggestion
        If `cmd` is set, it will be augmented with `spawner.get_args(). This will override the `CMD` specified in the Docker image.
```
As you point out it's already complicated, and now I think it'd be better not to mention entrypoint, args or command. If we add entrypoint I think we could also keep it's help doc simple, and instead add the complicated explanation to the general docs instead, e.g. on the [overview page](https://jupyterhub-kubespawner.readthedocs.io/en/latest/overview.html) or a new customisation guide page? That way you can include examples without bulking out the API docs too much.

consideRatio:
I'll go for a merge on this as it is and work on an entrypoint traitlet PR.

PR #503, opened 2021-05-04
add method and test to delete namespaced PVC in spawner base class

### LINKED ISSUE #446: Deleting named servers should delete the corresponding PVC

nsshah1288:
**### PR Summary by Erik**

This PR implements a delete_forever hook for KubeSpawner. The delete_forever hook will be respected by JupyterHub 1.4.0 and later and was implemented in jupyterhub/jupyterhub#3337. The idea of this hook is to allow the Spawners to perform cleanup of a users associated resources at the time a JupyterHub user is deleted.

As KubeSpawner can create PVCs for users, this KubeSpawner implementation checks for a PVC that it could have created for a user and tries to delete it. This behavior should be documented properly in the CHANGELOG.md for the next KubeSpawner release, I'm not sure if I think it would be labelled as a "breaking" change, but it is a change in the user experience that is quite important to be aware about.

This PR closes #446.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2633,3 +2633,28 @@ class KubeSpawner(Spawner):
                 # It's fine if it already exists
                 self.log.exception("Failed to create namespace %s", self.namespace)
                 raise
+
+    async def delete_forever(self):
+        """Called when a user is deleted.
+
+        This can do things like request removal of resources such as persistent storage.
+        Only called on stopped spawners, and is likely the last action ever taken for the user.
+
+        This will only be called once on the user's default Spawner.
+        Supported by JupyterHub 1.4.0+.
+        """
+        try:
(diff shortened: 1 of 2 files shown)


nsshah1288:
Please disregard this PR. I apologize for making this PR which has the wrong content.

PR #507, opened 2021-05-13
Add changelog for 1.0.0

### LINKED ISSUE #505: Release 0.17.0, 17.0.0, or 1.0.0?
@jupyterhub/jupyterhubteam I'd like to cut a release for KubeSpawner for Z2JH to make use of soon.

I'd also in general like to start working with major.minor.patch versions and not just 0.minor.patch as there is far less flexibility to communicate what kind of changes to expect when one only has an ability to bump two version specifiers.

### Question
Due to this, I wonder if you think it's okay if we cut a major release, and, if we do, use 17.0.0 or 1.0.0.

### Current state
We are currently on 0.16.1 and the changes since then are the following: https://github.com/jupyterhub/kubespawner/pulls?q=is%3Apr+merged%3A%3E%3D2021-03-01. These include a breaking change in the sense that PVCs will now be deleted as part of deleting a user or removing a named server.

consideRatio:
I interpreted more voices suggesting we release 1.0.0 instead of 17.0.0 or 0.17.0, so here is a changelog for 1.0.0.

Closes #505.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -2,6 +2,50 @@
 
 <!-- PR link template: - [#](https://github.com/jupyterhub/kubespawner/pull/) ([@](https://github.com/)) -->
 
+### [1.0]
+
+### [1.0.0]
+
+This release is the continuation of version 0.16.1 and could have been 0.17.0 in
+practice. We opted to release 1.0.0 as it enables us to communicate changes
+according to [SemVer](https://semver.org/). Using SemVer versioning, a change in
+each of the three version numbers (major.minor.patch) represents a different
+kind of change.
+
+### Breaking changes
+
+- When using KubeSpawner 1.0.0 or later together with JupyterHub 1.4.1 or later,
+  deleting a JupyterHub user or deleting (not just stopping) a named server will
+  lead to removing the associated PVC resource. To opt out of this behavior set
+  the `delete_pvc` configuration to `False`.
+
+#### New features added
+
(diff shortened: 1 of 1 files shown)


docs/source/changelog.md
```diff
@@ -2,6 +2,49 @@
...
+This release is the continuation of version 0.16.1 and could have been 0.17.0 in
+practice. We opted to release 1.0.0 as it enables us to communicate changes
+according to [SemVer](https://semver.org/). Using SemVer versioning, a change in
+each of the three version numbers (major.minor.patch) represents a different
+kind of change.
+
+### Breaking changes
+
+- When using KubeSpawner 1.0.0 or later together with JupyterHub 1.4.1 or later,
+  deleting a user or removing a named server will imply removal of the
```
manics, line 18:
```suggestion
  deleting a user or removing a named server will remove the
```
unless I've misunderstood and you're trying to communicate some other behaviour?

manics (commented):
Just one query on language used

yuvipanda (approved):
YAY!

Happy to merge in a day if nobody objects :)

docs/source/changelog.md
consideRatio: Ah :) I just pushed a change updating that language, what do you think of it now?

> - When using KubeSpawner 1.0.0 or later together with JupyterHub 1.4.1 or later,
  deleting a JupyterHub user or deleting (not just stopping) a named server will
  lead to removing the associated PVC resource. To opt out of this behavior set
  the `delete_pvc` configuration to `False`.

manics: LGTM

manics (approved):

yuvipanda:
I said 24h but i think 16 is good enough.

\o/ ty, @manics and @consideRatio!

PR #508, opened 2021-05-14
Remove unneeded dep

dhirschfeld:
It seems [`jupyterhub-dummyauthenticator`](https://github.com/jupyterhub/dummyauthenticator) may no longer be required?

image


consideRatio:
:heart: thanks!

Issue #491, opened 2021-03-03
Kubernetes' native environment substitution doesn't work all the time


yuvipanda:
### Bug description

Kubernetes supports [environment variable substitution](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/) - you can derive one variable by string interpolation of others. However, you can only interpolate variables you have already declared in the *list* of environment variables. And since kubespawner takes a dict rather than a list, we can't control the order here - it's always alphabetically sorted (not sure if that is intentional).

### To reproduce

Here's some sample YAML config for z2jh

```yaml
singleuser:
  env:
    MESSAGE_FIRST: "Hello"
    MESSAGE_SECOND: "World"
    MESSAGE: "$(MESSAGE_FIRST) $(MESSAGE_SECOND)
```

In the user container, you would expect `echo $MESSAGE` to give you `Hello World`, but it'll actually give you `$(MESSAGE_FIRST) $(MESSAGE_SECOND)` - no expansion. This is because, `MESSAGE` sorts *above* `MESSAGE_FIRST` and `MESSAGE_SECOND`, so when Kubernetes tries to process the expansions, it fails since these aren't defined yet.

If you instead change it to this, to force ordering:
```yaml
singleuser:
  env:
    MESSAGE_FIRST: "Hello"
    MESSAGE_SECOND: "World"
    ZESSAGE: "$(MESSAGE_FIRST) $(MESSAGE_SECOND)
```

The `ZESSAGE` is set to `Hello World` correctly.

consideRatio:
We could introduce an algorithm to stop using alphabetical order, but instead try to ensure we avoid this issue?

1. Expand all values using KubeSpawner expansions
2. Put anything not matching `"$(...)"
3. Loop until the remaining envs are 0:
   1. If all remaining envs reference other remaining envs, a circular dependency can be assumed and we emit a warning and put the envs in any order.
   2. Else put the first env remaining not referencing other remaining envs

yuvipanda:
Hmm, this requires we keep track of what features kubernetes supports and keep it up to date. Instead, I'd prefer some way of either letting environment be a list instead of string, or a 'weight's key that can be used. All keys added by KubeSpawner will be weight 0, z2jh weight 1, and then every other key is user specified weight. This lets users specify things as they wish and control ordering for whatever reason they might want. This already happens elsewhere (`hub.extraConfig` for example), so would probably work ok for most cases?

[...]

PR #510, opened 2021-06-04
Sort env to reliably expand nested env references

consideRatio:
KubeSpawner exposes a dictionary as configuration of environment variables, but a k8s container specification contain an ordered list where the order actually matters.

### How the order matters

If an environment variable is set to be a string that contains `$(SOME_ENV)`, it will be expanded if `SOME_ENV` has been defined earlier in the ordered list in the container specification.

### What this PR does

This PR makes KubeSpawner emit the container's environment list in a way that first and foremost ensures as many environment variables are expanded as possible.

### Is this a breaking change?

In a way yes, but I think most users will think of it as a bugfix - that the changed behavior will always be appreciated.

### Related

- https://github.com/jupyterhub/kubespawner/issues/491
- https://github.com/2i2c-org/pilot-hubs/issues/442

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -450,7 +450,8 @@ def make_pod(
             return set()
         # $(MY_ENV) pattern: $( followed by non-)-characters to be captured, followed by )
         re_k8s_env_reference_pattern = r"\$\(([^\)]+)\)"
-        return set(re.findall(re_k8s_env_reference_pattern, env_var_obj.value))
+        deps = set(re.findall(re_k8s_env_reference_pattern, env_var_obj.value))
+        return deps - {env_var_obj.name}
 
     unsorted_env = {}
     for env_var_key, env_var_obj in (env or {}).items():
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -435,24 +435,82 @@ def make_pod(
     if not csc:
         csc = None
 
-    # Transform a dict into valid Kubernetes EnvVar Python representations. This
-    # representation shall always have a "name" field as well as either a
+    # Transform KubeSpawners env input to valid Kubernetes EnvVar Python
+    # representations. They should have a "name" field as well as either a
     # "value" field or "value_from" field. For examples see the
     # test_make_pod_with_env function.
-    prepared_env = []
-    for k, v in (env or {}).items():
-        if type(v) == dict:
-            if not "name" in v:
-                v["name"] = k
-            prepared_env.append(get_k8s_model(V1EnvVar, v))
+    #
+    # While doing this, also extract information about references to other envs
+    # as we want to use those to make an intelligent sorting before we render
+    # this into a list with an order that matters.
+    #
+    def _get_env_var_deps(env_var_obj):
+        # only env var objects with explicit string values set will be evaluated
+        if not env_var_obj.value:
+            return set()
+        # $(MY_ENV) pattern: $( followed by non-)-characters to be captured, followed by )
+        re_k8s_env_reference_pattern = r"\$\(([^\)]+)\)"
+        return set(re.findall(re_k8s_env_reference_pattern, env_var_obj.value))
+
+    unsorted_env = {}
+    for env_var_key, env_var_obj in (env or {}).items():
+        if type(env_var_obj) == dict:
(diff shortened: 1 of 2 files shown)


tests/test_objects.py
```diff
@@ -2136,19 +2164,19 @@ def test_make_pod_with_ssl():
...
-                        {'name': 'JUPYTERHUB_USER', 'value': 'TEST'},
                         {
                             'name': 'JUPYTERHUB_SSL_CLIENT_CA',
                             'value': '/etc/jupyterhub/ssl/notebooks-ca_trust.crt',
                         },
+                        {
+                            'name': 'JUPYTERHUB_SSL_KEYFILE',
+                            'value': '/etc/jupyterhub/ssl/ssl.key',
+                        },
+                        {'name': 'JUPYTERHUB_USER', 'value': 'TEST'},
```
consideRatio, line 2179: Just re-ordered.

tests/test_objects.py
```diff
@@ -2176,23 +2204,3 @@ def test_make_pod_with_ssl():
...
-    assert namespace == {
-        'metadata': {
-            'annotations': {},
-            'labels': {
-                'component': 'singleuser-server',
-                'heritage': 'jupyterhub',
-            },
-            'name': 'test-namespace',
-        },
-    }
```
consideRatio, line 2198: This test function was defined twice. Removed.

tests/test_objects.py
```diff
@@ -2012,79 +2113,6 @@ def test_make_ingress_external_name():
...
-                    'name': 'jupyterhub-internal-certs',
-                    'secret': {'defaultMode': 511, 'secretName': 'ssl'},
-                }
-            ],
-        },
-        "kind": "Pod",
-        "apiVersion": "v1",
-    }
-
-
```
consideRatio, line 2087: This test function was added twice. Removed this entry.

consideRatio (commented):
@yuvipanda @minrk could you have a look at this? I'd like to cut z2jh 1.1.0 with this included in a kubespawner release.

kubespawner/objects.py
```diff
@@ -435,24 +435,82 @@ def make_pod(
...
+    # as we want to use those to make an intelligent sorting before we render
+    # this into a list with an order that matters.
+    #
+    def _get_env_var_deps(env_var_obj):
+        # only env var objects with explicit string values set will be evaluated
+        if not env_var_obj.value:
+            return set()
+        # $(MY_ENV) pattern: $( followed by non-)-characters to be captured, followed by )
+        re_k8s_env_reference_pattern = r"\$\(([^\)]+)\)"
+        return set(re.findall(re_k8s_env_reference_pattern, env_var_obj.value))
```
minrk, line 453: Remove self-references for e.g. `PATH=/prefix/bin:$(PATH)`

kubespawner/objects.py
```diff
@@ -435,24 +435,82 @@ def make_pod(
...
+            env_var_obj = get_k8s_model(V1EnvVar, env_var_obj)
+        else:
+            env_var_obj = V1EnvVar(name=env_var_key, value=env_var_obj)
+
+        unsorted_env[env_var_obj.name] = {
+            "deps": _get_env_var_deps(env_var_obj),
+            "env_var_key": env_var_key,
+            "env_var_obj": env_var_obj,
+        }
+
```
minrk, line 469: I think this can be simplified with a single key function passed to `sorted`. To do that, you'll need *all* transitive dependencies in the `deps` field, which can be computed after this direct-dependency dict exists. Need to avoid infinite loops with cycles.

consideRatio: @minrk I gave this a shot but I failed to arrive at something that I felt were about was simpler. I believe that even if I end up going for a pairwise comparison strategy which can have a clean `sorted()` statement, I perceived myself ending up shifting the complexity to first create a list that contained enough information to be pairwise comparable.

I did refactor my existing code to be simpler to read though. If you think it is acceptable as it is now, I suggest we go for it. I'd also be happy to replace this logic with an alternative implementation if you'd like to provide one, but I'd like that implementation to pass the test cases I've implemented.

minrk (approved):

PR #514, opened 2021-07-06
Generalize omit_namespace functionality

droctothorpe:
This PR generalizes the `omit_namespace` functionality leveraged by `pods` and `events` and leverages it for `services`, `ingresses`, and `endpoints`. 

We're altering our deployment to leverage user-scoped namespaces and found that this was required in our ingress proxy subclasses, but I think it may be useful enough to implement in the base. 

Totally open to feedback.

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -179,22 +179,23 @@ class ResourceReflector(LoggingConfigurable):
         #  their __init__() methods to derive list_method_name, but you
         #  could just set it directly in the subclass.
         if not self.list_method_name:
-            # This logic can be extended if we add other reflector types or
-            #  it can be directly supplied or overridden in a subclass.
-            if self.kind in ["pods", "events", "services", "ingresses", "endpoints"]:
-                if self.kind == "ingresses":
-                    singular_kind = "ingress"
-                if self.kind == "endpoints":
-                    # This handles inconsistent nomenclature in the
-                    # Python Kubernetes client.
-                    singular_kind = self.kind
-                else:
-                    singular_kind = self.kind[:-1]
+            plural_to_singular = {
+                "endpoints": "endpoints",
+                "events": "event",
+                "ingresses": "ingress",
+                "pods": "pod",
+                "services": "service",
+            }
 
+            if self.kind in plural_to_singular:
                 if self.omit_namespace:
-                    self.list_method_name = f"list_{singular_kind}_for_all_namespaces"
+                    self.list_method_name = (
+                        f"list_{plural_to_singular[self.kind]}_for_all_namespaces"
+                    )
                 else:
-                    self.list_method_name = f"list_namespaced_{singular_kind}"
+                    self.list_method_name = (
+                        f"list_namespaced_{plural_to_singular[self.kind]}"
+                    )
 
         # Make sure we have the required values.
         if not self.kind:
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -184,6 +184,10 @@ class ResourceReflector(LoggingConfigurable):
             if self.kind in ["pods", "events", "services", "ingresses", "endpoints"]:
                 if self.kind == "ingresses":
                     singular_kind = "ingress"
+                if self.kind == "endpoints":
+                    # This handles inconsistent nomenclature in the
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
This seems to look good to me, but I doubt we have good test coverage.

Is this something you have already tested or could test pre-merge?

droctothorpe:
Thanks for responding so quickly, @consideRatio! I was having issues running the tests and was hoping CI could execute them, but I've since fixed my local dev env and see that 6 tests are failing. I'll update the tests and push the changes. As far as documentation is concerned, there are no references to `omit_namespace` outside of the doc strings, which as far as I can tell do not require updates (i.e. they don't specify any particular objects).

droctothorpe:
Pushed one tiny change. Tests confirmed passing. 

image

kubespawner/reflector.py
```diff
@@ -172,26 +172,24 @@ def __init__(self, *args, **kwargs):
...
-            if self.kind == "pods":
-                if self.omit_namespace:
-                    self.list_method_name = "list_pod_for_all_namespaces"
-                else:
-                    self.list_method_name = "list_namespaced_pod"
-            elif self.kind == "events":
+            if self.kind in ["pods", "events", "services", "ingresses", "endpoints"]:
+                singular_kind = self.kind[:-1]
+                if self.kind == "ingresses":
+                    singular_kind = self.kind[:-2]
```
consideRatio, line 187: Feel free to ignore this, just ended up considering how I'd write this.

```suggestion
                if self.kind == "ingresses":
                    singular_kind = "ingress"
                else:
                    singular_kind = self.kind[:-1]
```

droctothorpe: Committed. That's much more straightforward. Thanks!

droctothorpe:
Not sure what the typical timeline is for merges, but I want to amend the `Generalize omit_namespace functionality` commit to include some co-authors (the need for this emerged in a mob programming session so I want to make sure that credit is fairly distributed, even if the change is minor in scope) by noon EST. I hope the amendment is not an inconvenience! 🙏

droctothorpe:
Change pushed. Nothing modified other than the commit message.

kubespawner/reflector.py
```diff
@@ -172,26 +172,29 @@ def __init__(self, *args, **kwargs):
...
+                if self.kind == "ingresses":
+                    singular_kind = "ingress"
+                if self.kind == "endpoints":
+                    # This handles inconsistent nomenclature in the
+                    # Python Kubernetes client.
+                    singular_kind = self.kind
                 else:
-                    self.list_method_name = "list_namespaced_pod"
-            elif self.kind == "events":
+                    singular_kind = self.kind[:-1]
```
yuvipanda, line 192: Can we instead just list singular and plurals in a dictionary instead? Automatically trying to figure out singulars from plurals usually gets filled with exceptions anyway, and I think that's clearer.

yuvipanda (changes requested):
Thanks for the PR, @droctothorpe! I've left a comment about style, but am happy to merge this after

kubespawner/reflector.py
droctothorpe: You're 1000x right. 

Pushed up the pivot to a dictionary. 

Validated tests are passing and execute pre-commit.

droctothorpe:
If you're interested, we would be happy to contribute some documentation around using kubespawner with nginx. It may even be worthwhile to integrate the implementation with z2jh to provide a turnkey solution that doesn't depend on CHP, which eliminates at least one single point of failure (HA for hub is a much bigger lift ofc, but one that would probably be really fun to work on).

yuvipanda:
> f you're interested, we would be happy to contribute some documentation around using kubespawner with nginx.

That would be absolutely amazing! Please do.

> It may even be worthwhile to integrate the implementation with z2jh to provide a turnkey solution that doesn't depend on CHP, which eliminates at least one single point of failure

This would be great to do if we can minimize the added maintenance complexity in z2jh itself. We can't make this the default since that'll require users install and manage nginx-ingress and cert-manager, but making it an option without requiring a lot of changes would be rad. I'd definitely use it.

yuvipanda:
Thanks, @droctothorpe!

PR #515, opened 2021-07-08
Set configuration before instantiating reflectors

droctothorpe:
This PR resolves a problem with KubeIngressProxy introduced after [this](https://github.com/jupyterhub/kubespawner/commit/08bf48796d5061f2586df83f21855c95df038be9#diff-23f3b56e5e9b6c8e67401eab22ce6a30ccfdfe9a2979bea23d33505d74515efd) commit moved the client configuration out of `reflector.py`.

The config was relocated to `spawner.py` and `proxy.py`. `spawner.py` works fine but `proxy.py` fails because the reflectors are instantiated **before** the configuration is set.

Relocating the configuration before the reflectors are instantiated in the `__init__` function resolved this problem for us. We're using KubeIngressProxy with nginx instead of CHP to enhance HA. 

The funny thing is that the `# Global configuration before reflector.py code runs` comment actually anticipated this problem.

### DIFF
--- a/kubespawner/proxy.py
+++ b/kubespawner/proxy.py
@@ -114,6 +114,12 @@ class KubeIngressProxy(Proxy):
         # an accidental bottleneck. Since we serialize our create operations, we only
         # need 1x concurrent_spawn_limit, not 3x.
         self.executor = ThreadPoolExecutor(max_workers=self.app.concurrent_spawn_limit)
+
+        # Global configuration before reflector.py code runs
+        self._set_k8s_client_configuration()
+        self.core_api = shared_client('CoreV1Api')
+        self.extension_api = shared_client('ExtensionsV1beta1Api')
+
         labels = {
             'component': self.component_label,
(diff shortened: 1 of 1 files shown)


droctothorpe:
Following up on this. Seems like an obvious merge since it resolves a regression.

yuvipanda:
@droctothorpe if you can fix the failing pre-commit test, am happy to merge.

droctothorpe:
Thanks for reviewing, @yuvipanda. I removed the offending blank line. There's no mention of pre-commit in https://github.com/jupyterhub/kubespawner/blob/master/CONTRIBUTING.md. Would you like me to submit a PR documenting installation / usage?

yuvipanda:
@droctothorpe thank you for updating it! And yes, that would be deeply appreciated :) We can steal from / point to https://github.com/jupyterhub/jupyterhub/blob/main/CONTRIBUTING.md perhaps?

PR #518, opened 2021-07-16
Expand username etc. in configured service_account

### LINKED ISSUE #512: {username} does not expand for service_account
Bug description
I am attempting to dynamically assign services accounts to singleuser pods based on the username.

Expected behaviour
c.KubeSpawner.service_account('user-{username}') results in c.KubeSpawner.service_account('user-myusername')

Actual behaviour
serviceaccount "user-{username}" not found"

consideRatio:
Closes #512 by implementing the wanted expansion of the service_account configuration. This change is actually a single line of logic change, but I updated some documentation systematically regarding the expansion of these variables for various configuration options.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -348,10 +348,11 @@ class KubeSpawner(Spawner):
         Template to use to form the namespace of user's pods (only if
         enable_user_namespaces is True).
 
-        `{hubnamespace}` is expanded to the namespace name in which the pod
-        is running, unless that is `default`, in which case it is expanded
-        to `user`.
-        `{username}` is expanded to the escaped, dns-label-safe username.
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
+        found within strings of this configuration. The username and servername
+        come escaped to follow the [DNS label
+        standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).
         """,
     )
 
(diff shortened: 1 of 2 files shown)


kubespawner/spawner.py
```diff
@@ -1844,7 +1890,7 @@ async def get_pod_manifest(self):
             extra_resource_guarantees=self.extra_resource_guarantees,
             lifecycle_hooks=self.lifecycle_hooks,
             init_containers=self._expand_all(self.init_containers),
-            service_account=self.service_account,
+            service_account=self._expand_all(self.service_account),
```
consideRatio, line 1893: This is the actual change.

kubespawner/spawner.py
```diff
@@ -1844,7 +1894,7 @@ async def get_pod_manifest(self):
...
```
consideRatio, line 1897: This is the actual change, the rest is documentation and tests.

yuvipanda:
Thanks a lot, @consideRatio! We should systematically expand it in everything - how do you think we should approach that?

consideRatio:
@yuvipanda oh I'd say the strategy would be to consider if we could do a single `_expand_all()` call on a high level object rather than individual pieces we add to the big pod object for example. But, if we do that, I see it as a quite big risk that it may need to be reverted because we didn't consider something unknown so I'd be a bit hesitant merging such change without thinking it through a lot.

consideRatio:
Thanks for review and merge @yuvipanda!!! :heart:

yuvipanda:
> But, if we do that, I see it as a quite big risk that it may need to be reverted because we didn't consider something unknown so I'd be a bit hesitant merging such change without thinking it through a lot.

Yeah, I agree - especially since shell scripts might have things inside `{}` all the time.

Maybe we can do this in two steps:

- [ ] Evaluate places that aren't going to contain shell scripts (like serviceaccount) and expand them automatically. I don't think this needs a major version bump.
- [ ] During next major version bump, let's make it expand everything. Folks will then have to explicitly escape any `{}` in shell scripts.

How does that sound?

PR #521, opened 2021-07-22
changed hostname format for supporting IPv6 single stack deployment

kaz-tk:
when case of type of IPv6 single stack deployment.
KubeSpawner does not work.

due to URL format for registering to ConfigurableProxy is wrong.

so this commit aim to change hostname format for supporting IPv6 single deployment.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -5,6 +5,7 @@ This module exports `KubeSpawner` class, which is the actual spawner
 implementation that should be used by JupyterHub.
 """
 import asyncio
+import ipaddress
 import multiprocessing
 import os
 import string
@@ -1802,15 +1803,22 @@ class KubeSpawner(Spawner):
             hostname = self.dns_name
         else:
             proto = "http"
-            hostname = pod["status"]["podIP"]
-
+            _address_str = pod["status"]["podIP"]
+            _address = ipaddress.ip_address(_address_str)
+            if type(_address) == ipaddress.IPv6Address:
+                hostname = f"[{_address_str}]"
+            elif type(_address) == ipaddress.IPv4Address:
+                hostname = _address_str
+            else:
+                hostname = _address_str
         if self.pod_connect_ip:
+            # ToDo: need to change for supporting IPv6 single stack
             hostname = ".".join(
                 [
                     s.rstrip("-")
                     for s in self._expand_user_properties(self.pod_connect_ip).split(
-                        "."
-                    )
+                    "."
+                )
                 ]
             )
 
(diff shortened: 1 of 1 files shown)


gracefullife:
sorry, forgot to run pre-commit scripts.
now run pre-commit and re-push again

gracefullife:
need further discussion, when case of using connect_ip.
https://github.com/jupyterhub/kubespawner/pull/521/commits/13e6ce83654e928752d03dacc7b4052661cb4b0b#diff-6fdb3a928956e8ab2b810679d51fae72870d45ed4a088e85e16814339aff8901L1807-L1815

at this scope of commit, only support case of using pod_ip.

consideRatio:
@gracefullife thanks for contributing to this project! :tada: I currently lack knowledge about IPv4 and IPv6 difference, so my review may not so straightforward - I'm hoping you can help me learn what's relevant!

- Can you combine your two commit into a single commit to avoid cluttering the git history with format?
   (`rebase -i HEAD~2` followed by marking the second commit with `fixup`)
- Can you provide logic that never risks failing to assign the hostname variable? These are two conditionals that may or may not be exclusive, I'm not sure, but if they are they can just as well be `if -> else` rather than `if -> elif`. I think though that sometimes the address could be something like username1.service.my-namespace.cluster.local as well, so perhaps one should assign the hostname variable directly and then adjust it in case its a IPv6 address?
  ```python
              if type(_address) == ipaddress.IPv6Address:
                  hostname = f"[{_address_str}]"
              elif type(_address) == ipaddress.IPv4Address:
                  hostname = _address_str
  ```
- Can you help me understand what happens if we wrap the address in `[]`? I lack IPv6 experience and the foundation to understand if this change makes sense by itself without learning some more first.

gracefullife:
thanks for your reply.

>Can you combine your two commit into a single commit to avoid cluttering the git history with formatting?
(rebase -i HEAD~2 followed by marking the second commit with fixup)
done.
appreciate for your guide

>Can you provide logic that never risks failing to assign the hostname variable? 
hostname variable comes from podIP.
so suppose there is no risk/possibility that hostname is not IP address.
https://github.com/jupyterhub/kubespawner/commit/773e31193ebc48ecd375d75f11d60d8da64ba2ad#diff-6fdb3a928956e8ab2b810679d51fae72870d45ed4a088e85e16814339aff8901R1806

>Can you help me understand what happens if we wrap the address in []? I lack IPv6 experience and the foundation to understand if this change makes sense by itself without learning some more first.
that is correct format of IPv6 when we specify in browser, with port number.

in example,  case of IPv4, for specifying localhost, 8080 port, we usually specify, 127.0.0.1:8080
IPv6 case, we need to specify [::1]:8080.

if there is no bracket [], this will be like this.
::1:8080

but in this case, application can not recognize which is IP address and which is port.
so we need to use [].

consideRatio:
Thanks for helping me understand. I'm quickly coming back to various PRs before cutting a release.

Note that I could not push a commit to this PR that fixes the introduced formatting changes that remain, so you have to do it.

> > Can you provide logic that never risks failing to assign the hostname variable?
>
>hostname variable comes from podIP.
>so suppose there is no risk/possibility that hostname is not IP address.

The podIP field can actually be a non-ip value =/ I recall it can be domain names etc as well in certain k8s cluster setups.

consideRatio (changes requested):
Feedback comments needs to be addressed.

consideRatio:
Closing as stale, but I also think #619 may have resolved this.

PR #525, opened 2021-09-14
Terminate process correctly from reflector thread

yuvipanda:
Our reflectors run on their own thread, and sys.exit() from
a thread does not actually exit the process - just the thread.
When the reflector has to shut down because it can not talk
to the k8s master anymore for a while, we had just been stopping
the reflector thread and doing nothing else.

This will instead actually cause the hub to stop, and in z2jh
restart - which seems the right thing to do.

Ref https://github.com/2i2c-org/pilot-hubs/issues/680,
where we encountered this during a period of time when the
k8s master was unreachable for about 1 minute.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -11,6 +11,7 @@ import multiprocessing
 import os
 import string
 import sys
+import signal
 import warnings
 from asyncio import sleep
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks for the bump @yuvipanda! I hesitated with this during an initial glance a while back and forgot about it.

My initial hesitations were
1. Is this really a desired behavior in the logic?
   Answer: yes - the logic has logging statements indicating that this is really meant to shut down the hub itself, but failed to do so.
2. Will this happen when there is a brief failure of the k8s api-server?
   Answer: no - this happens after several retries and failures.

Conclusion: this is a straight forward bugfix.

PR #539, opened 2021-11-03
ci: misc fixes, don't run tests on markdown changes, etc

consideRatio:


.github/workflows/publish.yaml
```diff
@@ -8,17 +8,6 @@ on:
...
-  run:
-    # Declare bash be used by default in this workflow's "run" steps.
-    #
-    # NOTE: bash will by default run with:
-    #   --noprofile: Ignore ~/.profile etc.
-    #   --norc:      Ignore ~/.bashrc etc.
-    #   -e:          Exit directly on errors
-    #   -o pipefail: Don't mask errors from a command piped into another command
-    shell: bash
-
```
consideRatio, line 21: I added this in the past, thinking it was relevant - but it wasn't unless we used windows runners.

minrk:
Nice!

consideRatio (commented):
Merged before i could comment clarifications :D Thanks Min!

Issue #544, opened 2021-11-12
The default of allow_privilege_escalation should be False


consideRatio:
# Suggested changes

- [ ] We should support having `allow_privilege_escalation` be `None`, `True`, or `False`.
  - `None` should mean it won't be set explicitly
  - `True`/`False` should mean it will be set explicitly.
- [ ] We should make the default value of `allow_privilege_escalation` be `False`
  Currently the default value is True but behaves as None.

# Current situation

Currently, we the default value of `allow_privilege_escalation` is `True`, and we don't allow the value to be `None`.

https://github.com/jupyterhub/kubespawner/blob/fdfa2ad8fb756d52175a0f1515c7c66addc50a55/kubespawner/spawner.py#L915-L927

Currently, our logic is to treat `True` as a `None` value should and would be treated. I consider this a bug. The current behaviour is as if the default value was None, because our logic here makes us not explicitly set it.

https://github.com/jupyterhub/kubespawner/blob/fdfa2ad8fb756d52175a0f1515c7c66addc50a55/kubespawner/objects.py#L427-L428

I consider this a bug because there is a difference between `True` and `None` in case there is a PodSecurityPolicy resource in the k8s cluster that has a `DefaultAllowPrivilegeEscalation` set to False. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation for some details about this.

# Background

Root escalation is one required step towards doing bad things, to default to preventing it by default is to put another barrier up against something problematic like https://www.wiz.io/blog/chaosdb-explained-azures-cosmos-db-vulnerability-walkthrough.

PR #545, opened 2021-11-15
Default allow_privilege_escalation to False

yuvipanda:
Allows it to be set to None as well, to not set the property.

This is a breaking change for hubs where admins were granting
sudo rights to users. That already required some extra work,
so this would be an additional propety to set for that. The
added security benefit from this much more secure default is
well worth the breakage IMO.

Fixes https://github.com/jupyterhub/kubespawner/issues/544

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -64,7 +64,7 @@ def make_pod(
     fs_gid=None,
     supplemental_gids=None,
     privileged=False,
-    allow_privilege_escalation=True,
+    allow_privilege_escalation=False,
     container_security_context=None,
     pod_security_context=None,
     env=None,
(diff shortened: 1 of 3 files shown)


consideRatio (approved):
This LGTM.

It would be good to have another reviewer agree we get this right as its a breaking change. @minrk @manics, perhaps you could have a look?

Note that this is both a bugfix, and a change of default.

1. Bugfix: because our "True" value in the past acted as "None" rather than an explicit True.
2. A change of default: because now we instead of having "None" as a default, we have "False" as an explicit default.

minrk:
Makes sense!

yuvipanda:
Ty, @minrk and @consideRatio

PR #547, opened 2021-11-19
Add changelog for KubeSpawner 2.0.0

consideRatio:
I requested a lot of people to review this change, as its relevant that both we are in agreement of making a major version release, and that the communication about this breaking change is clear within the changelog.

This would be a release that would go with the Z2JH 2.0.0 release, and that the breaking change in #545 would align well with all other containers we run as part of the Helm chart - we default to a secure configuration and ask users to escalate explicitly rather than the other way.


docs/source/changelog.md
```diff
@@ -4,6 +4,31 @@
 
 ## [Unreleased]
 
+## 2.0
+
+### 2.0.0 - 2021-11-22
+
+#### Breaking changes
+
+A breaking change was introduced in [#545](https://github.com/jupyterhub/kubespawner/pull/545), making the default value of `allow_privilege_escalation` be `False`. Practically a consequence of this is that a user can't use `sudo` unless this or the `privileged` configuration is explicitly set to `true`. Technically it means that the JupyterHub user Pod that KubeSpawner creates will have a container with a `securityContext` that has `allowPrivilegeEscalation` set to `false` by default now.
```
yuvipanda, line 13: I don't think `privileged` needs to be true - that has a special meaning quite different from root. I think the only difference with this change is that `allow_privilege_escalation` needs to be set to True, not `privileged`

yuvipanda (approved):
one comment about `privileged` but me likey

docs/source/changelog.md
consideRatio:
```suggestion
A breaking change was introduced in [#545](https://github.com/jupyterhub/kubespawner/pull/545), making the default value of `allow_privilege_escalation` be `False`. Practically a consequence of this is that a user can't use `sudo` unless `allow_privilege_escalation` is explicitly set to `True`. Technically it means that the JupyterHub user Pod that KubeSpawner creates will have a container with a `securityContext` that has `allowPrivilegeEscalation` set to `false` by default now.
```

Ah, I meant to say something relating to...

> AllowPrivilegeEscalation is true always when the container is: 1) run as Privileged OR 2) has CAP_SYS_ADMIN.

But, this is written out below, so why make the text more complicated by including it in this paragraph as well.

yuvipanda (approved):

docs/source/changelog.md
```diff
@@ -4,6 +4,31 @@
 
 ## [Unreleased]
 
+## 2.0
+
+### 2.0.0 - 2021-11-22
+
+#### Breaking changes
+
+A breaking change was introduced in [#545](https://github.com/jupyterhub/kubespawner/pull/545), making the default value of `allow_privilege_escalation` be `False`. Practically a consequence of this is that a user can't use `sudo` unless `allow_privilege_escalation` is explicitly set to `True`. Technically it means that the JupyterHub user Pod that KubeSpawner creates will have a container with a `securityContext` that has `allowPrivilegeEscalation` set to `false` by default now.
```
manics, line 13:
```suggestion
A breaking change was introduced in [#545](https://github.com/jupyterhub/kubespawner/pull/545), making the default value of `allow_privilege_escalation` be `False`. This means a user can't use `sudo` unless `allow_privilege_escalation` is explicitly set to `True`. The JupyterHub user Pod that KubeSpawner creates will have a container with a `securityContext` that has `allowPrivilegeEscalation` set to `false` by default.
```
I've simplified the language a bit.

docs/source/changelog.md
```diff
@@ -4,6 +4,31 @@
...
+
+For reference, the following can be read about `allowPrivilegeEscalation` in [Kubernetes official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/):
+
+> AllowPrivilegeEscalation: Controls whether a process can gain more privileges than its parent process. This bool directly controls whether the `no_new_privs` flag gets set on the container process. AllowPrivilegeEscalation is true always when the container is: 1) run as `Privileged` OR 2) has `CAP_SYS_ADMIN`.
+
+To revert to the previous behavior, set `allow_privilege_escalation` explicitly to `None`.
```
manics, line 19: What's the difference between `None` and `False`?

manics (commented):
Are you planning to release this immediately? Or are you going to hold off/release an RC in case any other issues appear between now and the release of Z2JH 2?

docs/source/changelog.md
consideRatio: Then it is not explicitly set, and that means thre k8s clusters default will be used, which is true by default, but could be false via a podsecuritypolicy

consideRatio:
> Are you planning to release this immediately? Or are you going to hold off/release an RC in case any other issues appear between now and the release of Z2JH 2?

I'm just nudging us towards a KubeSpawner 2.0.0 release, which i consider a blocker for Z2JH 2.0.0, which is also blocked by JupyterHub 2.0.0.

Do you know of things you want to introduce as breaking changes in KubeSpawner 2.0.0?

docs/source/changelog.md
```diff
@@ -4,6 +4,31 @@
...
+
+#### Breaking changes
+
+A breaking change was introduced in [#545](https://github.com/jupyterhub/kubespawner/pull/545), making the default value of `allow_privilege_escalation` be `False`. This means a user can't use `sudo` unless `allow_privilege_escalation` is explicitly set to `True`. The JupyterHub user Pod that KubeSpawner creates will have a container with a `securityContext` that has `allowPrivilegeEscalation` set to `false` by default.
+
+For reference, the following can be read about `allowPrivilegeEscalation` in [Kubernetes official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/):
+
+> AllowPrivilegeEscalation: Controls whether a process can gain more privileges than its parent process. This bool directly controls whether the `no_new_privs` flag gets set on the container process. AllowPrivilegeEscalation is true always when the container is: 1) run as `Privileged` OR 2) has `CAP_SYS_ADMIN`.
+
+To revert to the previous behavior, set `allow_privilege_escalation` explicitly to `None`.
```
manics, line 19:
```suggestion
To revert to the previous behavior of using the cluster's default, set `allow_privilege_escalation` explicitly to `None`.
```

manics (commented):
> Do you know of things you want to introduce as breaking changes in KubeSpawner 2.0.0?

Nope! Just wanted to check, in case anyone else has something lurking on their todo list.

minrk (approved):

docs/source/changelog.md
```diff
@@ -4,6 +4,31 @@
 
 ## [Unreleased]
 
+## 2.0
+
+### 2.0.0 - 2021-11-22
```
consideRatio, line 9:
```suggestion
### 2.0.0 - 2021-11-28
```

PR #558, opened 2022-01-05
Accomodate breaking k8s client change in 21.7.0

athornton:


consideRatio:
Thank you @athornton!

PR #559, opened 2022-01-19
Wait 0.1s between events, not 1s

yuvipanda:
Given that each pod spawn seems to take at least 5 events,
this meant each user start took *at least* 5 seconds, even if
the pod actually started up much much faster! This affects UX
a bit, as the hub seems 'slower' to users than it really is.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2214,7 +2214,8 @@ class KubeSpawner(Spawner):
 
             if break_while_loop:
                 break
-            await asyncio.sleep(1)
+            # Check for new events every 0.1s
+            await asyncio.sleep(0.1)
 
     def _start_reflector(
         self,


yuvipanda:
Ideally there'd be some sort of notification mechanism here and not a sleep, but this should give us what we want already without (hopefully) too much performance overhead.

Alternatively, we could exponentially backoff here, but only if there is a real performance problem.

consideRatio:
> Given that each pod spawn seems to take at least 5 events,
this meant each user start took at least 5 seconds, even if
the pod actually started up much much faster!

Hmmm, I don't read the code to reflect that. I read it as each while loop iteration where the sleep happen, an inner loop will be able to process all events that has showed up so far. This matches my UX where I often see a lot of events pop up quickly in succession rather than waiting 1s between each.

I think 1s is reasonable as it is, and merging this would make me a bit worried that the performance of those super-massive hubs are degraded - which would certainly also degrade the end UX more detrimentally than waiting something between 0-1 seconds more than needed which is what i think will happen.

yuvipanda:
ah, hmm - I see this is a bit more complicated than I had thought. So it looks like:

1. check if we have any events since the last time we checked,
2. if we do, then emit them
3. wait 1s
4. go to step 1

Does that match your understanding, @consideRatio?

I agree re: performance. If that is the only concern, I'll just deploy this on one of my super massive hubs next week and run it for a few days to see how it performs. Will that alleviate your concern?

consideRatio:
> Does that match your understanding, @consideRatio?

Yepp that matches!

- My understanding is that you with this PR would reduce the average wait time from this logic from 0.5 seconds (average of 0-1 seconds) to 0.05 seconds (average of 0 to 0.1 seconds).
- My concern is that by polling ten times more often to gain such a small performance gain, all kinds of response times and availability of the hub will be reduced under relatively heavy load. I believe it will be very tricky for you to demonstrate there is no performance penalty that leads to problems in situations when very many users are spawning at the same time.

Overall, I don't think it's justified to save on average ~0.45 seconds by risking a performance penalty that could lower the hubs capacity to handle larger amount of traffic. I quite strongly opinionated against this change to 0.1 sleep given my assumptions on the average time saved are correct. If you want to save time like that, maybe we could settle for something like 0.5 second sleep instead of 0.1? You get less and less time saved and take a higher and higher performance hit the more you try to optimize this.

minrk:
I think you're both right - each iteration it processes _all_ available events, and the sleep only occurs when there are no events available to process. So this sleep should add _at most_ 1 second total (even if it occurs 10 times), and on average 0.5s to each progress completion as it waits for the final event. In practice, this means that every 1 second, all events from Kubernetes during that second will be emitted by JupyterHub.

Further, I also do not believe it should affect spawn performance (if all my async generator code is right, which is a big if!), as we [stop consuming progress](https://github.com/jupyterhub/jupyterhub/blob/2.0.2/jupyterhub/apihandlers/users.py#L725) when the original spawn finishes.

However, I agree that a hardcoded sleep isn't great, as it makes it hard to test if improved responsiveness will be too costly performance-wise.

Aside: it would be good to have a sense of the overhead of a single coroutine iteration, because then we'd have a good baseline order of magnitude for how much it would cost per-spawner to have this iterator going with a given sleep (for instance, if it took 1ms, 100 concurrent spawns with a sleep of 0.1 would use 100% cpu just idly checking for an empty list of events).

We could switch the wait for be event-based instead of sleep-based, but that can be tricky to get right and possibly even more costly performance-wise, since they events come from another thread, and threadsafe event signaling is expensive.

consideRatio:
@yuvipanda and I spoke on slack about this after I made the comment above!

I've learned about the experience @yuvipanda has and would be excited to:
- understand the performance implications of this change
- learn more about the process of testing something like this

I'm very aware that I'm not knowledgeable enough to make a useful estimate the performance implications though. Given that you both (@yuvipanda and @minrk) have a lot more experience with these kinds of questions, I don't want to be a blocker of this PR.

@yuvipanda if you do performance testing I'd love to learn a bit on how you go about it, perhaps via a video call where I can watch while you investigate things.

yuvipanda:
I don't have a good way to measure kubernetes pod startup latency itself properly yet - am surprised there isn't an easy kubernetes metric...

yuvipanda:
Turning off events completely for the hub had no effect on server start times:

image

It was turned off around the 01/25 8:00 mark, and no discernable change in server start times is found. This leads me to conclude that you're both right and events don't seem to be the cause of the slowdown. So I'm going to close this one. 

I think the way forward here is to:

1. Figure out a real way to measure pod startup latency,
2. Take a profiler to the hub to see what we can eeke out there, if that's what is causing it.

Thanks :)

consideRatio:
Thanks for following this up @yuvipanda!!

I'm excited about potential performance improvements to make z2jh installations more snappy so you are most welcome to loop me into anything about you work on.

Hmmm... I wonder if there is a way to conclude if its: jupyterhub being slow, k8s api-server being slow, k8s nodes being slow, k8s storage mounting or similar being slow. At this point, I don't have an idea on how to separate these kinds of issues from each other very well.

yuvipanda:
> At this point, I don't have an idea on how to separate these kinds of issues from each other very well.

Yeah, more data is needed to figure this out. Right now, all we can say is that this `sleep` isn't affecting anything

minrk:
> Figure out a real way to measure pod startup latency,

I think it could be useful to instrument the events themselves, which are enough to get info like image pull time, scheduling time, etc. Events wouldn't include the time between starting the container and Spawner.start completing, though

PR #560, opened 2022-01-25
tickets/DM-33044D

athornton:
Convert the existing KubeSpawner framework to asyncio with kubernetes_asyncio

Empirically, it appears to perform the same or slightly better on average than the status quo implementation, and reduce the reported error rate by a factor of 3 or 4.

Work still required:

The testing framework is a mess.

In an ideal world we'd replace all the *real* Kubernetes operations it's doing with the K8s mock from Safir, and then be able to convince upstream that it's just as good (and much faster, of course).

We don't have a multichannel ws client in kubernetes_asyncio, which means that the tests that rely on capturing return codes from our commands will not work; currently they are skipped.

The namespace fixture is not being correctly torn down, and I do not know where to put the finalizers to cleanly close the watches that give us pod stdout output in the test framework.  Frankly I find asyncio testing and fixtures quite confusing, and I am going to need some help with this.


consideRatio:
Hi @athornton, it looks like you have put in a lot of work into this and I want to cheer you on for contributing to the project. Looking at this at the moment, I'm quite lost though.

- Could you update your top level description to introduce the change, for example mention if/where it has been discussed before?
- Could you update the topic to reflect the change to be comprehensible for maintainers/users?
- Could you ensure you provide a motivation and background etc to this suggested change in a way that is comprehensible without assuming too much pre-requisite knowledge, allowing a broad range of people understand the change and motivation?

I converted the PR into a draft for now as I consider it a draft at this point.

athornton:
Whoops!  This is my internal PR; I'll be submitting it once we've done some more work on it at Rubin.  I meant to PR against lsst-sqre/kubespawner instead.  Closing.

consideRatio:
@athornton ah, okay thanks! Seems like very relevant exploratory work!! I hope it turns out well!

Issue #496, opened 2021-03-15
We need tests for the KubeIngressProxy for its maintainability


consideRatio:
This project includes KubeSpawner and KubeIngressProxy, one is a custom Spawner class, and one is a custom Proxy class - both are configurable classes JupyterHub can be configured to rely on in order to spawn pods and control an external proxy's routing of traffic.

KubeIngressProxy currently resides in this repo but there are no tests setup for it or similar, which makes it quite complicated to develop KubeSpawner as a change to the classes to track the k8s resources (these are the Reflector classes) can break KubeIngressProxy that also use them.

No matter if KubeIngressProxy becomes a standalone project or similar, I think it would be great to have some tests developed for it. Just making sure it remains functional on a very primitive level would be great.

Issue #163, opened 2018-04-22
KubeIngressProxy docs


consideRatio:
While working on a PR (#161) I came across [`KubeIngressProxy`](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/proxy.py). I'm hoping to learn where and how it is used in order to be able to make a thorough PR that doesn't break things or leave the repo in an inconsistent state.

### TODO
- [ ] Comprehend where `KubeIngressProxy` is used
- [ ] Comprehend how `KubeIngressProxy` is used
- [ ] Create a docstrings for the KubeIngressProxy class definition
- [ ] Consider exporting `KubeIngressProxy` as well from [`__init__.py`](https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/__init__.py) along with KubeSpawner.

/cc: @minrk @yuvipanda as you have contributed to `proxy.py`

clkao:
This seems to be an alternative proxy class implemented as k8s operator that manages the ingress' route table directly for the hub and spawned user notebooks, replacing the default `configurable-http-proxy`.  It can be configured with `c.JupyterHub.proxy_class`

There are some explanation and diagram in [the slides](https://indico.cern.ch/event/663264/contributions/2863771/attachments/1591005/2517845/erokhin_spbu_swan_final.pdf).  I am also curious about the status of the implementation and if anyone uses it in production.

If this is working well it should be (optionally) enabled for z2jh as a more native ingress-based proxy.

gweis:
Hi, ... I played around with it, and I can say that it works, but it needed a bit of tweaking for my environment to play nice with nginx ingress controller.

1. I patched the kubespawner.objects.make_ingress method to include ingress.spec.rule.host property (which is only populated if you use per user sub domains), and I have added some TLS setup as well there. After that it was all good.
2. You have to make sure that notebooks and jupyterhub run the same version, otherwise jupyterhub will bail out due to a redirect loop too early. (It takes a little bit for the ingress controller to configure the route)

Otherwise it seems to be working well and hub and notebooks feel a bit faster as well (one less proxy and bottleneck in the request chain).

[...]

PR #563, opened 2022-02-04
Rely on the event loop: use `kubernetes_asyncio` instead of `kubernetes` and dedicated threads

athornton:
## Change description

__No dedicated threads -> async/await with the event loop__
KubeSpawner has relied on [kubernetes-client/python (aka. `kubernetes` on PyPI)](https://github.com/kubernetes-client/python#readme) to speak with the k8s api-server to watch, create, delete pods etc. For KubeSpawner to function, it has what we call _reflectors_  watching for what is going on in the k8s cluster. These reflectors has previously been running in dedicated threads, but are via this PR scheduled to run via [the _event loop_](https://docs.python.org/3/library/asyncio-eventloop.html) with `async`/`await`.

__Replaced dependency: `kubernetes` -> `kubernetes_asyncio`__
For this to be possible, we have switched from kubernetes-client/python (aka. `kubernetes` on PyPI) to [tomplus/kubernetes_asyncio (aka. `kubernetes_asyncio` on PyPI)](https://github.com/tomplus/kubernetes_asyncio#readme).

## Benefits of the change

__Reduction of transient k8s errors__
@athornton had observed transient errors when spawning user pods. The errors were showing up as k8s Event's emitted that could be seen via `kubectl describe pod <name>` and said for example "failed to sync configmap cache" or described failures to mount a volume.

With this change, @athornton observed those errors reduced significantly.

> [...] I think that's because coroutines do a much better job of getting fast-to-create K8s objects (like secrets and configmaps) actually built before something tries to operate on them than a bunch of threads hammering willy-nilly on the control plane did. That, however, is closer to a guess than a well-instrumented theory.

__Reducing codebase complexity__
While KubeSpawner's codebase will remain quite complicated due to its quite complicated task, its complexity can be reduced notably by not using multiple threads.

__Indications of slightly improved performance__
Load tests seem to provide a similar or slightly improved performance.

## Technical notes

__A strategy to await required async object initialization__
An object constructor (`__init__`) can't be `async`, but we needed to `await` some initialization logic before JupyterHub that created a KubeSpawner object would start using its public methods. To do this we schedule the async init logic (`_async_init()`) in the constructor and then ensure the public functions JupyterHub may call, that requires the async init to finish, would ensure to await it by using a decorator (`_await_async_init`).

## Breaking changes

- Dropped support for Python 3.6
- The configuration `k8s_api_threadpool_workers` is deprecated as we don't create any threads any more, but instead relies on scheduling everything to run in the event loop.
- Changed dependencies: `kubernetes` dropped, `kubernetes_asyncio` introduced.
- This project includes KubeSpawner and KubeIngressProxy, both influenced by this PR, but only KubeSpawner has been properly tested, so maybe `KubeIngressProxy` got broken unintentionally.

---

_This PR summary is written by @consideRatio_


athornton:
*This* time, I meant to PR it upstream.

consideRatio:
@athornton can you put in work on a thorough description for this very significant change? You overview this change suggestion far better than any one else can, and it saves so much effort on the reviewer part if some of that can be conveyed by text ahead of technical review.

- What motivated this change and what related discussions are there?
- What are the major things done
- What do you expect to be breaking
- Do you expect issues about something for end users?
- What to focus extra review attention to
- etc.

These kinds of things would be relevant to write about for example. If you want to focus on technical aspects for a while longer, mark this as a draft PR!

athornton:
Absolutely!  I'd be delighted to talk about what, conceptually, this patch is doing.

The major motivation for us, the Kubespawner users at Rubin Observatory, using it in our Science Platform's Notebook Aspect, is that we are seeing a lot of (usually transient) errors when spawning with Kubespawner.  Usually they are of the form "failed to sync configmap cache", or failed mountpoints (usually secrets or configmaps, sometimes real NFS volumes).  These mostly seem to resolve (so users do eventually get their pods).  Nevertheless, we've been trying to eliminate spurious errors in our logs so that the remaining logs are meaningful.  While attempting that, we realized that understanding what was happening in a threaded Kubespawner was really, really difficult.

So I undertook to make Kubespawner asyncio/coroutine-based, using tomplus/kubernetes_asyncio, because it would be a lot easier for me to reason about.  Coroutines with fairly explicit that control is yielded are something I find a lot easier to keep in my head than threads, where I have no idea of what's executing where at any given time.

The major change is conceptually quite simple: replace the official (and synchronous) Kubernetes Python client with kubernetes_asyncio.  That in turn would ripple throughout the codebase removing the need for the asynchronize method, and since I was using asyncio anyway, I replaced the Tornado gen.timeouts with asyncio.wait_for.

That, by itself, is the vast majority of the technical changes.  My first cut at this was a lot more radical in terms of reflector redesign (even at one point abandoning the reflector entirely).  This one aims to be much more minimal.

The major structural changes are that reflectors and proxy objects are created now with classmethods, simply because you cannot call async methods from `__init__`.  Unfortunately, I cannot impose that change on how the Spawner object itself is instantiated, so each of its start/stop/poll entrypoints (and indeed the hook when a pre_spawn_hook was called) needed to do reflector and client config initialization.

I do not expect anything to break.  After some work, the test suite passes all the cases that pass for me in the standard Kubespawner.  I'm using it on my development cluster now with no issues.  Load tests seem to indicate that performance is equivalent to, and maybe slightly better than, the threaded version, but I don't have firm data on that yet.  My reported error rate is down significantly, and I think that's because coroutines do a much better job of getting fast-to-create K8s objects (like secrets and configmaps) actually built before something tries to operate on them than a bunch of threads hammering willy-nilly on the control plane did.  That, however, is closer to a guess than a well-instrumented theory.

The one change that may be controversial is that it requires a much, much newer Kubernetes client: specifically 19.15.1 (rather than 10.1).  The reason for this is simple: https://github.com/jupyterhub/kubespawner/issues/423 / https://github.com/jupyterhub/kubespawner/pull/424 use a little partial-application hack to get the K8s client to NOT deserialize the JSON it receives, and this is necessary to keep CPU consumption reasonable in busy clusters.  The asyncio_kubespawner didn't let us do this until I alerted @tomplus to it (https://github.com/tomplus/kubernetes_asyncio/pull/177) ; he committed support for this very quickly, but I'm not going to ask him to backport it to the distant past.

The major remaining issue is related to https://github.com/tomplus/kubernetes_asyncio/issues/12 , which remains open.  In order to retain the ability to execute Python in a pod, because there is no multichannel ws client in kubernetes_asyncio presently, I had to make the synchronous K8s client a test dependency, and patch it into the test suite, while using the async client everywhere I could.

The Kubespawner, and particularly its reflectors, is still a complex and fairly difficult-to-understand piece of code, but I think this change, to make it coroutine-based rather than thread-based, does manage to make it easier (still far from easy) to think about what it's doing.

athornton:
Black and isort are fighting over changes; if someone will point me to the correct settings for both I'm happy to apply them.

yuvipanda:
This is super awesome! Threading is hard, async is hard, mixing those two is super hard. Very glad to remove that from the codebase!

kubespawner/proxy.py
```diff
@@ -189,18 +181,15 @@ async def add_route(self, routespec, target, data):
 
         async def ensure_object(create_func, patch_func, body, kind):
             try:
-                resp = await self.asynchronize(
-                    create_func, namespace=self.namespace, body=body
-                )
+                resp = (await create_func(namespace=self.namespace, body=body),)
```
minrk, line 184: This is creating a tuple, which I assume is not intended. I don't think we use the response, though.

```suggestion
                resp = await create_func(namespace=self.namespace, body=body)
```

minrk:
> Black and isort are fighting over changes;

We use pre-commit to do the autoformatting (runs black, reorder-python-imports):

```shell
pre-commit install # registers hooks so it runs automatically on future commits
pre-commit run --all-files # run once on already committed files to satisfy everything, should be automatic after that
```

kubespawner/reflector.py
```diff
@@ -203,12 +205,27 @@ def __init__(self, *args, **kwargs):
...
+    async def reflector(cls, *args, **kwargs):
+        """
+        This is how you should instantiate a reflector in order to bring
+        it up in a usable state, as this classmethod does load
+        Kubernetes config, acquires an API client, and starts the watch
+        task.
+        """
+        inst = cls(*args, **kwargs)
+        await load_config()
+        inst.api = shared_client(inst.api_group_name)
```
minrk, line 224: Do we still need `shared_client`? The reason we had that for sync kubernetes is that each client object a ThreadPool, so the process exploded with threads if we kept creating clients (especially when it was an unconfigurable nthreads=n_cpus per client!). Do kubernetes_asyncio clients also have state we need to avoid creating too many times?

kubespawner/reflector.py
```diff
@@ -257,6 +275,11 @@ def _watch_and_update(self):
         and as long as we don't try to mutate them (do a 'fetch / modify /
         update' cycle on them), we should be ok!
         """
+        #
+        # I guess?  It is the case that the resources are read-only in the
+        #  spawner, so the worst that will happen will be that they're out-
+        #  of-date.  I think.
```
minrk, line 281: Not sure we need this additional comment in addition to the docstring.

It is threadsafe to write to dicts from one thread and read from another, as long as you don't make assumptions across multiple statements that the dict isn't changing:

```
# safe:
pod = reflector.pods.get(pod_name)
if pod:
    ... # pod exists

# unsafe
if pod_name in reflector.pods:
    # could change between check and access!
    pod = reflector.pods[pod_name]
```


_however_ since this is now async, there are no threads to worry about, so it's much easier to be confident that the reflector is perfectly safe, and the docstring should probably be updated to reflect that there are no longer any concerns here. It takes finite time to be updated, but there's no practical difference between waiting for the k8s API to notify a change vs waiting for it to propagate to `reflector.resources` via coroutines.

kubespawner/reflector.py
```diff
@@ -372,17 +402,29 @@ def start(self):
         start of program initialization (when the singleton is being created),
         and not afterwards!
         """
-        if hasattr(self, 'watch_thread'):
-            raise ValueError('Thread watching for resources is already running')
+        if self.watch_task and not self.watch_task.done():
+            raise ValueError('Task watching for resources is already running')
```
minrk, line 406: While we're here, `RuntimeError` might be more appropriate for this error.

kubespawner/reflector.py
```diff
@@ -372,17 +402,29 @@ def start(self):
...
 
-        self._list_and_update()
-        self.watch_thread = threading.Thread(target=self._watch_and_update)
-        # If the watch_thread is only thread left alive, exit app
-        self.watch_thread.daemon = True
-        self.watch_thread.start()
+        await self._list_and_update()
+        self.watch_task = asyncio.create_task(self._watch_and_update())
```
minrk, line 409: `create_task` [requires Python 3.7](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task), but setup currently requires 3.6. No objection to bumping required Python in this PR since it's about time anyway, but the same result is accomplished with the older `asyncio.ensure_future` without changing the required Python.

setup.cfg
```diff
@@ -4,3 +4,5 @@ test=pytest
 [tool:pytest]
 # Ignore thousands of tests in dependencies installed in a virtual environment
 norecursedirs = lib lib64
+# At some point probably want to make this "strict"
+asyncio_mode = auto
```
minrk, line 8: This is new to me! I'll be happy to remove our [hacks to do this](https://github.com/jupyterhub/jupyterhub/blob/39da98f13345fa1bee1cfae34bf9bf7d748ab438/jupyterhub/tests/conftest.py#L60-L70).

I think auto mode is what we want, and is [recommended by pytest-asyncio](https://pypi.org/project/pytest-asyncio/#auto-mode). We don't have a reason to switch to strict, which is only recommended for testing with multiple async frameworks, which isn't relevant here.

setup.py
```diff
@@ -2,25 +2,24 @@
...
 if v[:2] < (3, 6):
-    error = "ERROR: jupyterhub-kubespawner requires Python version 3.6 or above."
+    error = "ERROR: kubespawner requires Python version 3.6 or above."
     print(error, file=sys.stderr)
     sys.exit(1)
 
 setup(
-    name='jupyterhub-kubespawner',
-    version='2.0.1.dev',
+    name='kubespawner',
```
minrk, line 14: Looks like the setup.py changes other than the version bumps are for your fork? Should be removed before it lands in the main repo.

tests/conftest.py
```diff
@@ -479,8 +517,14 @@ def _exec_python_in_pod(kube_client, kube_ns, pod_name, code, kwargs=None, _retr
...
     # see https://github.com/kubernetes-client/python/issues/812
-    client = stream(
-        kube_client.connect_get_namespaced_pod_exec,
+    #
+    # That's why we are using the synchronous Kubernetes client here
+    # and why we imported them in the first place: kubernetes_asyncio
+    # does not yet support multichannel ws clients, which are needed
+    # to get the return code.
+    # cf https://github.com/tomplus/kubernetes_asyncio/issues/12
+    client = sync_stream(
```
minrk, line 526: Great solution! I never thought of the fact that switching to kubernetes_asyncio didn't mean we couldn't still use the sync client for the missing feature. Makes perfect sense.

minrk (commented):
Wonderful to see this!

athornton:
Thank you all for the suggestions!  I don't know if we still *need* the shared client, but not opening a separate connection for every single time we talk to the control plane is probably a big win.  Which is a little sad because the asyncio client really wants to be used as a context manager, and if we used a shared client and singleton reflectors we lose having a place to cleanly close the aiohttp objects (which you can see at test shutdown, for instance).

My inclination is to leave it as is for now and then benchmark shared-client-vs-context-manager in a future PR (although I suspect that is something we will only be able to determine at scale, and my load testing framework needs a lot more work before I feel like it's solid).

kubespawner/proxy.py
athornton: Unintended.

kubespawner/reflector.py
athornton: I don't know if they have state, but we at least would be doing TLS connection establishment once per client, right?  I suspect at scale a shared client is a win, but I think we should keep the shared client for this PR anyway.

kubespawner/reflector.py
athornton: Let's bump requirements, since I've used create_task in a bunch of places.

kubespawner/reflector.py
athornton: Done and done.

setup.py
athornton: Thank you, yes.

tests/conftest.py
athornton: Wish I could claim credit, but that idea on our end came from Russ Allbery (@rra)

athornton:
Also, my huge thanks to @rra for the internal review process at Rubin, which made the reception of this work upstream in the JupyterHub community much, much easier than it otherwise would have been.

athornton:
Is there anything else I should do on this?  Tests are now clean, and I think I hit the review comments.

setup.py
```diff
@@ -6,21 +6,21 @@
 from setuptools import setup
 
 v = sys.version_info
-if v[:2] < (3, 6):
-    error = "ERROR: jupyterhub-kubespawner requires Python version 3.6 or above."
+if v[:2] < (3, 7):
+    error = "ERROR: kubespawner requires Python version 3.7 or above."
```
minrk, line 10: The `python_requires` on line 27 also needs to be updated to `>=3.7`

minrk (commented):
One minor thing left in setup.py, but I think this is all set! Anything more you want to test @consideRatio or @yuvipanda?

kubespawner/spawner.py
```diff
@@ -2067,7 +2038,12 @@ async def poll(self):
         Note that a clean exit will have an exit code of zero, so it is
         necessary to check that the returned value is None, rather than
         just Falsy, to determine that the pod is still running.
+
+        We cannot be sure the Hub will call start() before poll(), so
+        we need to load client configuration and start our reflectors
+        at the top of each of those methods.
         """
+        await self.initialize_reflectors_and_clients()
```
consideRatio, line 2046:
```suggestion
        """
        # We cannot be sure JupyterHub will call start() before poll(), so
        # we need to load client configuration and start our reflectors
        # here.
        await self.initialize_reflectors_and_clients()
```

athornton:
Pinging again.

consideRatio (changes requested):
# Thank you!

Thank you for your work on this @athornton! :heart: :tada: :sunflower:

I fully agree that this reduces the complexity of the logic. I can also see that a having multiple threads running k8s api-client calls to get things done could reasonably create things out of order and lead to transient errors.

I absolutely support getting this PR merged and will do my best to help it get merged quickly. I have a few open questions for us to consider still though!

# Request for changes/additional consideration

- [x] `k8s_api_threadpool_workers` has been removed.
  - Is it correct that this doesn't make sense any more?
  - Should it removed like this, or retained with a deprecation notice?
  - [x] REVIEW: It has been deprecated like this, which doesn't print any warning etc - is there a pattern to re-use for this?
    ```python
    k8s_api_threadpool_workers = Integer(
        config=True,
        help="""
        DEPRECATED.
        No longer has any effect, as there is no threadpool anymore.
        """,
    )
    ```
    - Looking at jupyterhub's deprecated values, it has often been deprecated without something besides a notice in the help string.
- [x] `k8s_api_ssl_ca_cert` and `k8s_api_host` are stranded configuration no longer respected, but probably still relevant.
  - Is there equivalent configuration for the new k8s client so we can avoid making this breaking change?
  - [x] This is resolved.
  - [x] REVIEW: the implementation of solving this.
    - I'm at ease with the solution as it is now.
- [x] Naming of `initialize` and `reflector` - use `create` instead?
- [x] Revise the GitHub Workflow [test.yaml](https://github.com/jupyterhub/kubespawner/blob/main/.github/workflows/test.yaml#L36-L75) to not run against different versions of the old kubernetes client
  - [x] I'll update it to stop including references to the python k8s client we don't care about any more.
- [x] Misc small details about comments etc
- [x] @athornton can you provide an example of a transient error or two in more detail?
  - I'm for example not sure if they are errors only related to spawning user pods, or pods in general, or errors related to more than spawning pods for example.
  - [ ] I'll summarize this for the PR description

consideRatio:
# Draft PR description

Providing a PR description for a large change like this is crucial for future maintenance work, such as tracking if an issue relates to this change or doing something unrelated but simply being able to understand that this change is unrelated.

I'll go ahead and use the draft below before merge to update the PR description.

## Problems addressed

- **Reducing transient k8s pod creation errors**

  > [...] we are seeing a lot of (usually transient) errors when spawning with Kubespawner. Usually they are of the form "failed to sync configmap cache", or failed mountpoints (usually secrets or configmaps, sometimes real NFS volumes). These mostly seem to resolve (so users do eventually get their pods).

  > My reported error rate is down significantly, and I think that's because coroutines do a much better job of getting fast-to-create K8s objects (like secrets and configmaps) actually built before something tries to operate on them than a bunch of threads hammering willy-nilly on the control plane did. That, however, is closer to a guess than a well-instrumented theory.

- **Reducing codebase complexity**

  KubeSpawner's codebase is complicated, but this complexity could be lowered by relying on a Python kubernetes API client library that works natively with `async/await`. KubeSpawner has relied on [kubernetes-client/python (kubernetes @ PyPI)](https://github.com/kubernetes-client/python#readme) but are with this PR replacing it with [tomplus/kubernetes_asyncio (kubernetes_asyncio @ PyPI)](https://github.com/tomplus/kubernetes_asyncio#readme).

  Before this PR, we've had an [`asyncronize` method](https://github.com/jupyterhub/kubespawner/blob/d59c14cc3bcf4fead2d339ba6f7b549262f2deff/kubespawner/spawner.py#L2128-L2130) in the KubeSpawner class that wrapped calls for us which added complexity for example. This PR removes it from the KubeSpawner class.

## Technical notes

- **Reflectors / Proxy class objects are to be created via factory functions**

  An object constructor (`__init__` function) can't be async, but a factory function that returns an initialized object can be. Due to that, there is a refactoring switch from relying on `__init__` to create the Reflector class objects and Proxy class objects to instead rely on class functions.

  > Unfortunately, I cannot impose that change on how the Spawner object itself is instantiated, so each of its start/stop/poll entrypoints (and indeed the hook when a pre_spawn_hook was called) needed to do reflector and \[Kubernetes\] client config initialization.

## Breaking changes

This PR is not meant to cause breaking changes besides dropping support for Python 3.6. It has been tested in production where each user get an individual namespace.

This project includes KubeSpawner and KubeProxy, but only KubeSpawner is properly tested. This PR hasn't verified that it didn't break the KubeProxy project.

## Performance changes

Load tests seem to provide a similar or slightly improved performance.

kubespawner/proxy.py
```diff
@@ -97,37 +96,38 @@ def _namespace_default(self):
...
-        self.executor = ThreadPoolExecutor(max_workers=self.app.concurrent_spawn_limit)
-
-        # Global configuration before reflector.py code runs
-        self._set_k8s_client_configuration()
-        self.core_api = shared_client('CoreV1Api')
-        self.extension_api = shared_client('ExtensionsV1beta1Api')
+    @classmethod
+    async def initialize(cls, *args, **kwargs):
+        """
+        This is how you should get a proxy object.
```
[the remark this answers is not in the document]
athornton: Sounds good.

kubespawner/reflector.py
```diff
@@ -203,12 +211,27 @@ def __init__(self, *args, **kwargs):
...
 
-    def __del__(self):
-        self.stop()
+        # Note that simply instantiating the class does not load
+        # Kubernetes config, acquire an API client, or start the
+        # watch task.
 
-    def _list_and_update(self):
+    @classmethod
+    async def reflector(cls, *args, **kwargs):
```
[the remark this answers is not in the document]
athornton: I am certainly not wedding to the specific name.  create() is fine by me.

setup.py
```diff
@@ -6,29 +6,30 @@
...
     print(error, file=sys.stderr)
     sys.exit(1)
 
 setup(
     name='jupyterhub-kubespawner',
-    version='2.0.1.dev',
+    version='3.0.0',
```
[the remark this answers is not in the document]
athornton: That's fine.  I wanted to bump the major version not because the API is incompatible, but because it's a very, very major change.

kubespawner/reflector.py
```diff
@@ -224,22 +247,21 @@ def _list_and_update(self):
         if not self.omit_namespace:
             kwargs["namespace"] = self.namespace
 
-        initial_resources = getattr(self.api, self.list_method_name)(**kwargs)
+        method = getattr(self.api, self.list_method_name)
+        initial_resources_raw = await ((method)(**kwargs))
```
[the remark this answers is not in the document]
athornton: I'll test that.

tests/conftest.py
```diff
@@ -443,19 +485,19 @@ def __str__(self):
...
     kwargs are passed to the function, if it is given.
     """
-    if V(kubernetes.__version__) < V("11"):
-        pytest.skip(
-            f"exec tests require kubernetes >= 11, got {kubernetes.__version__}"
-        )
-    pod = wait_for_pod(kube_client, kube_ns, pod_name)
+    if V(sync_version) < V("11"):
+        pytest.skip(f"exec tests require kubernetes >= 11, got {sync_version}")
+    pod = await wait_for_pod(kube_client, kube_ns, pod_name)
```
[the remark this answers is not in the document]
athornton: Works for me.

kubespawner/spawner.py
```diff
@@ -227,25 +203,34 @@ def __init__(self, *args, **kwargs):
...
+    async def initialize(cls, *args, **kwargs):
+        """In an ideal world, you'd get a new KubeSpawner with this.
+        That's not how we are called from JupyterHub, though; it just
+        instantiates whatever class you tell it to use as the spawner class.
+
+        Thus we need to do the initialization in poll(), start(), and stop()
+        since any of them could be the first thing called (imagine that the
+        hub restarts and the first thing someone does is try to stop their
+        running server).  Further (and this is actually true in the Rubin
+        case) a pre-spawn-start hook that depends on internal knowledge of
```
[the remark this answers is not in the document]
athornton: This particular way of using a pre-spawn hook is Rubin-specific, and perhaps not generalizable.

kubespawner/spawner.py
```diff
@@ -227,25 +203,34 @@ def __init__(self, *args, **kwargs):
...
-        except kubernetes.config.ConfigException:
-            kubernetes.config.load_kube_config()
-        if self.k8s_api_ssl_ca_cert:
-            global_conf = client.Configuration.get_default_copy()
-            global_conf.ssl_ca_cert = self.k8s_api_ssl_ca_cert
-            client.Configuration.set_default(global_conf)
-        if self.k8s_api_host:
-            global_conf = client.Configuration.get_default_copy()
-            global_conf.host = self.k8s_api_host
-            client.Configuration.set_default(global_conf)
```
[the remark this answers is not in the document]
athornton: I didn't *mean* to not respect them.  So, if not, that's a bug.

kubespawner/spawner.py
```diff
@@ -273,20 +258,6 @@ def _set_k8s_client_configuration(self):
         """,
     )
 
-    k8s_api_threadpool_workers = Integer(
```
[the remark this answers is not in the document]
athornton: Marking as ignored and tacking a deprecation warning onto it is probably a good idea.

kubespawner/spawner.py
athornton: Fine by me.

athornton:
The transient errors we see are of two main classes, although I think they represent the same underlying control-plane races.  "Failed Volume Mount" is very common, although it almost always eventually resolves and we get a running Pod with the correct mounts.

Sometimes the failed mount is an actual NFS volume (we are using the magic "nfs" type in Volume specifications for the Pod, so the PV/PVC creation is happening under the hood and hidden from us, but what K8s is doing is at least the moral equivalent of defining a PV that maps to the NFS volume, and then attaching a PVC to bind that to the Pod), but more often it is a ConfigMap or Secret that we generate at spawn time and mount into the pod.

The other common error is "failed to sync configmap cache".  It too pops up quite often, and also almost always resolves and we do eventually get a running Pod with everything mounted.  I *think* these happen when the control plane receives and acknowledges a request to create the resource, but then the Pod-being-created tries to mount that resource before it actually exists.

kubespawner/proxy.py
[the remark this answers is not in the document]
athornton: Slightly reworded, but, basically, yes.

kubespawner/reflector.py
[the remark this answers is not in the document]
athornton: Yup.  That phrasing baffles me too.  I wish I could tell you how I arrived there in the first place.

setup.py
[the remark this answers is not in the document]
athornton: Reverted (to 2.0.2.dev)

tests/conftest.py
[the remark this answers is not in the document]
athornton: Done.

kubespawner/spawner.py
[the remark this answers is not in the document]
athornton: Revised.  I hope it's clearer now.

kubespawner/spawner.py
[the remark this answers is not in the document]
athornton: I've now duplicated the Proxy code in Spawner, basically.  Ugly but effective.

kubespawner/spawner.py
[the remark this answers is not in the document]
athornton: Done.

kubespawner/spawner.py
athornton: Done.

athornton:
@consideRatio Please have a look at the current state.

athornton:
In a perfect world, we'd have KubeIngressProxy tests and then we'd know if it worked.  It feels silly to me to *not* convert it to asyncio at the same time as we do the rest of the work.  Specifically...if we don't convert it, it means we have to carry around the old sync reflector for a while longer, and that's where a huge part of the complexity is.

I'm willing to bet that if it's currently broken, fixing it is not hard, just because there's not all that much to the KubeIngressProxy class.

kubespawner/spawner.py
```diff
@@ -227,25 +203,46 @@ def __init__(self, *args, **kwargs):
...
+        # in kubernetes.py syntax.  It is being set here (but after
+        # load_config()) for readability / coupling with traitlets values
         if self.k8s_api_ssl_ca_cert:
             global_conf = client.Configuration.get_default_copy()
             global_conf.ssl_ca_cert = self.k8s_api_ssl_ca_cert
             client.Configuration.set_default(global_conf)
         if self.k8s_api_host:
             global_conf = client.Configuration.get_default_copy()
-            global_conf.host = self.k8s_api_host
+            global_conf.ssl_ca_cert = self.k8s_api_ssl_ca_cert
```
consideRatio, line 240:
```suggestion
            global_conf.host = self.k8s_api_host
```

kubespawner/clients.py
```diff
@@ -41,8 +43,16 @@ def shared_client(ClientType, *args, **kwargs):
...
         _client_cache[cache_key] = weakref.ref(client)
+
     return client
+
+
+async def load_config():
+    try:
+        kubernetes_asyncio.config.load_incluster_config()
+    except kubernetes_asyncio.config.ConfigException:
+        await kubernetes_asyncio.config.load_kube_config()
```
consideRatio, line 58: You extracted parts of the load_config logic into its dedicated function, I figure we should extract all parts there. I understand that would mean we need to pass some configuration but thinks that the right call.

It makes sense for such logic to be here, but not for it to be directly under initialize resources or a resource named like that.

consideRatio (commented):
Could you coalesce the logic setting the kubernetes client config into the introduced load_config?

I believe proxy.py had one implementation with a standalone function being called after load_config, and then spawner.py inlined the code etc.

kubespawner/clients.py
athornton: Will do.

athornton: Done.  I'm not a huge fan of how I did it but short of refactoring it into a mixin class that Spawner, Proxy, and Reflector all used, I don't immediately see a better way than having the load_config requestor pass a reference to itself into load_config so load_config can check for the attributes and use them if it finds them.

kubespawner/spawner.py
athornton: I implemented this by adding an optional "caller" parameter which users of load_config() can fill out with a reference to themselves; in that case load_config checks whether the caller has the right attributes, and whether those attributes have values.

I'm not wild about that, but I am even less wild about implementing a mixin class with only one method that Reflector, Proxy, and Spawner all inherit from, which was my other idea.

kubespawner/clients.py
```diff
@@ -41,8 +44,29 @@ def shared_client(ClientType, *args, **kwargs):
...
+
+    if caller:
+        if hasattr(caller, "k8s_api_ssl_ca_cert") and caller.k8s_api_ssl_ca_cert:
+            global_conf = Configuration.get_default_copy()
+            global_conf.ssl_ca_cert = caller.k8s_api_ssl_ca_cert
+            Configuration.set_default(global_conf)
+        if hasattr(caller, "k8s_api_host") and caller.k8s_api_host:
+            global_conf = Configuration.get_default_copy()
+            global_conf.host = caller.k8s_api_host
+            Configuration.set_default(global_conf)
```
consideRatio, line 72: I figure `caller` is better made not optional. Having it optional made the code become more complicated, and made me think that perhaps this function was actively used in a way were it was optional - which it wasn't.

```suggestion
async def load_config(caller):
    """
    Loads configuration for the Python client we use to communicate with a k8s
    api-server, and tweaks that configuration based on specific settings on the
    passed caller object.
    """
    try:
        kubernetes_asyncio.config.load_incluster_config()
    except kubernetes_asyncio.config.ConfigException:
        await kubernetes_asyncio.config.load_kube_config()

    if caller.k8s_api_ssl_ca_cert:
        global_conf = Configuration.get_default_copy()
        global_conf.ssl_ca_cert = caller.k8s_api_ssl_ca_cert
        Configuration.set_default(global_conf)
    if caller.k8s_api_host:
        global_conf = Configuration.get_default_copy()
        global_conf.host = caller.k8s_api_host
        Configuration.set_default(global_conf)
```

consideRatio:
Hmmm... I understand now that the `create` function isn't used for proxy/spawner class, only for the reflector class. Can we remove them - but retain the one used to provide us with the Reflectors?

The documentation we would delete could migrate.
- The `__init__` docstring could mention there that one should await the initialize function following having created the object with `__init__`.
- The initialize functions' docstrings (there are two, one named `initialize_reflectors_and_clients` for KubeSpawner, and one named`_initialize_resources` for KubeIngressProxy), could mention that they represent logic that have been extracted from `__init__` as we want to make use of `async/await` for it - which `__init__` can't.

athornton:
We can do that...although I think it would also make sense (but, sure, not for this PR) to provide a create() classmethod, and use that upstream in JupyterHub (maybe with some fancy introspection to see whether there is a create() classmethod on the spawner/proxy class, using it preferentially).  As we move to a more asyncio-based implementation across the board their utility will only increase.

athornton:
Well, we can sort of do that.  KubeIngressProxy doesn't actually have its own `__init__` so I'll just put the usage note in the class docstring instead.

athornton:
If the Proxy reflectors are supposed to be singletons too, we could do the same trick we do with the spawner, and call initialize_resources() at the start of each of the defined methods.  I don't know enough about the Proxy and its usage to know whether that's right, so for now I'm going to just add the "you better call initialize_resources() after __init__()" to the docstring.

kubespawner/clients.py
athornton: I don't see how we avoid calling it from the Reflector, unless we make the Reflector require a parent object.

That might be the right thing to do...or we could just add those attributes to the Reflector too.

athornton: And parent is always specified, so all we need to do is check that it's not None.

athornton:
I think this will do the trick...as long as anyone creating a KubeIngressProxy reads the documentation, anyway.

athornton:
(if all Reflectors *are* supposed to be singletons, maybe we enforce that in the class itself?  I think that would buy us a fair amount of simplicity)

kubespawner/clients.py
```diff
@@ -28,6 +29,8 @@ def shared_client(ClientType, *args, **kwargs):
     so that concurrent calls to shared_client
     will all return the same instance until
     all references to the client are cleared.
+
+    You should usually await load_config before calling this.
```
consideRatio, line 33: I look for more clarity. I'm working under the idea that "you must" rather than "you should". Do you agree @athornton?

I'm looking through the code, aiming to enforce that - which we need to do - right?

athornton: I agree.

consideRatio:
@athornton I've submitted https://github.com/lsst-sqre/kubespawner/pull/8 which is a PR to this PR!

kubespawner/reflector.py
```diff
@@ -300,41 +326,48 @@ def _watch_and_update(self):
                 if self.timeout_seconds:
                     # set watch timeout
                     watch_args['timeout_seconds'] = self.timeout_seconds
+                # This is a little hack to defeat the Kubernetes client
+                # deserializing the objects it gets, which can use a lot
+                # of CPU in busy clusters.  cf
+                # https://github.com/jupyterhub/kubespawner/pull/424
```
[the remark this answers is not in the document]
athornton: Agree.  I will update.  "compare" is fine.  It is short for "confer", and is usually used as short-hand for "cross-reference"...but it's pretty archaic and also almost certainly from the liberal arts side of the world.

kubespawner/spawner.py
```diff
@@ -123,12 +118,10 @@ class KubeSpawner(Spawner):
...
 
-    # We want to have one single threadpool executor that is shared across all
-    # KubeSpawner instances, so we apply a Singleton pattern. We initialize this
-    # class variable from the first KubeSpawner instance that is created and
-    # then reference it from all instances. The same goes for the PodReflector
-    # and EventReflector.
-    executor = None
+    # The PodReflector and EventReflector are singletons.  Where to initialize
+    # them becomes a sort of thorny question in an asyncio world.  See the
+    # commentary on the initialize() method.
```
consideRatio, line 123:
```suggestion
    # commentary on the initialize_reflectors_and_clients() method.
```

kubespawner/spawner.py
```diff
@@ -274,16 +252,11 @@ def _set_k8s_client_configuration(self):
     )
 
     k8s_api_threadpool_workers = Integer(
-        # Set this explicitly, since this is the default in Python 3.5+
-        # but not in 3.4
-        5 * multiprocessing.cpu_count(),
         config=True,
         help="""
-        Number of threads in thread pool used to talk to the k8s API.
+        DEPRECATED.
```
consideRatio, line 254:
```suggestion
        DEPRECATED in KubeSpawner 3.0.0.
```

consideRatio:
## Status update and request for help

I think I'm OK with the changes from a KubeSpawner perspective, but from a KubeIngressProxy perspective I'm very concerned. I think we have broken it because we haven't made its reflectors become initialized or provided a way for them to be that in any location like KubeSpawner does via the `start()` `poll()` and `stop()` functions.

Overall, I think we should make a time investment with `KubeIngressProxy` to make it retain life in this repo, or discard it. To retain life, at least we should have a primitive test (see #496), but preferably also some documentation improvements (see #163).

- Q1 - KubeIngressProxy maintenance:
  @yuvipanda @minrk what do you think about the maintenance path onwards for `KubeIngressProxy`, and what do you think we should do associated with getting this merged - which breaks KubeIngressProxy as it currently is?
- Q2 - Short term fixes to KubeIngressProxy:
  @athornton do you have ideas on how to potentially make KubeIngressProxy keep functioning by making sure its reflectors gets initialized?
- Q3 - Longer term fixes to KubeIngressProxy:
  @minrk being very experienced with Python and understanding async things very well - how do you think we should solve the situation with `__init__` not being able to be async and how JupyterHub always initializes the spawners syncronous with `__init__`?
  - @athornton proposed defining a async `create()` function that does instance creation + the async initialization, but the problem is that JupyterHub will initialize the proxy_class / spawner_class with a typical `ClassName()` rather than using a async factory function.
  - Currently we rely on initializing the async parts from all the async functions that could be called first: start, poll, stop - and have no equivalent strategy for KubeIngressProxy.

athornton:
I do have an idea for Q2: if KubeIngressProxy's Reflectors should each *also* be singletons, then we do what we do with KubeSpawner, and just call initialize_resources() each time we enter any of the publicly-exposed methods.  Even if they're not singletons, we could do something lame like an internal bool, `_configured` or something, which we'd set the first time we ran the configuration, and if it was set, skip reconfiguration, so that the configuration and initialization of reflectors is idempotent.  I could probably implement that in a couple hours--there aren't all that many methods the proxy provides, after all.

If they are supposed to be singletons, then I'd propose we bring the enforcement-of-singleton-ness into the Reflector class itself, and then call the initialization with each publicly exposed method (and a warning that if, somehow, you create a KubeIngressProxy and then call an internal method on it before any public method has been called, you'd better call its initialization method yourself).

For Q3: how much influence do we have over upstream JupyterHub?  Where it initializes its spawner and proxy classes, we could look into the class, and if it has a method named create(), and if that method is a classmethod, then call that and use the resulting object rather than just using init().  I have no idea how controversial that change would be upstream.

minrk:
> what do you think we should do associated with getting this merged - which breaks KubeIngressProxy as it currently is?

I'm not sure on this one. On the one hand, this is really our one repo for jupyterhub+kubernetes interactions. On the other, the IngressProxy is a lot less mature and widely used than the Spawner, so it probably makes the most sense for it to be its own package. I don't like the proxy blocking this PR, but I don't want to break things if folks are using them.

> if KubeIngressProxy's Reflectors should each also be singletons

Unlike Spawners, there is only one Proxy so it's fine for things to be instance attributes. The singletons are a solution to sharing resources across many Spawner instances. It's not a principle we need to adopt everywhere. If it's simpler for consistency among reflectors for them to be singletons, that's fine. Or if it's simpler for them to be instance attributes, that's fine, too.


> how do you think we should solve the situation with __init__ not being able to be async and how JupyterHub always initializes the spawners syncronous with __init__?

`__init__` must be sync, but it is called from within the event loop. That means the easiest way to kick-off async initialization is to schedule it with `asyncio.ensure_future()`, e.g.:

```python
def __init__(self, **kwargs):
    super().__init__(**kwargs)
    # kick-off async init, if there is any
    self._init_future = asyncio.ensure_future(self.async_init())

async def async_init(self):
    ...
```

Then if you need to wait for the async init to have finished (may not be necessary), you can do that in later methods (check_routes, etc.):


```python
async def check_routes(self):
    if self._init_future is not None:
        await self._init_future
        self._init_future = None
    ...
```

If folks want to get DRY, this can be added to methods with a decorator without changing the existing method bodies at all:

```python
def init_first(method):
    @functools.wraps(method)
    async def async_method(self, *args, **kwargs):
         if self._init_future is not None:
            await self._init_future
            self._init_future = None
        return await method(self, *args, **kwargs)
    return async_method

...

@init_first
def check_routes(self): ...
```

consideRatio:
Oh I really like that strategy Min! I've made an attempt on implementing it in https://github.com/lsst-sqre/kubespawner/pull/10 - what do you think @athornton and @minrk ?

athornton:
I love it!  The decorator is non-intrusive and makes everything much clearer.

athornton:
Test_spawn_progress is hanging forever.  Hmmm.

kubespawner/clients.py
```diff
@@ -1,33 +1,43 @@
...
+    """Return a shared kubernetes client instance
+    based on the provided arguments.
 
     A weak reference to the instance is cached,
     so that concurrent calls to shared_client
     will all return the same instance until
     all references to the client are cleared.
+
+    Note that we must await load_config before calling this function as it
+    relies on global client configuration being loaded.
```
consideRatio, line 40: We have an inline comment about this, this is redundant.

```suggestion
```

kubespawner/proxy.py
```diff
@@ -97,63 +96,74 @@ def _namespace_default(self):
...
+        # Schedules async initialization logic that is to be awaited by async
+        # functions by decorating them with @_await_async_init.
+        self._async_init_future = asyncio.ensure_future(self._async_init())
 
-        # Global configuration before reflector.py code runs
-        self._set_k8s_client_configuration()
+    async def _async_init(self):
+        """
+        This method is scheduled to run from `__init__`, but not awaited there
+        as it can't me marked as async.
```
consideRatio, line 115:
```suggestion
        as it can't be marked as async.
```

kubespawner/proxy.py
```diff
@@ -97,63 +96,74 @@ def _namespace_default(self):
...
+
+        Since JupyterHub won't await this method, we ensure the async methods
+        JupyterHub may call on this object will await this method until
```
consideRatio, line 118:
```suggestion
        JupyterHub may call on this object will await this method before
```

kubespawner/reflector.py
```diff
@@ -27,6 +28,15 @@ class ResourceReflector(LoggingConfigurable):
     kubernetes resources.
 
     Must be subclassed once per kind of resource that needs watching.
+
+    Creating a reflector should be done with the create() classmethod,
+    since that, in addition to creating the instance, initializes the
+    Kubernetes configuration, acquires a K8s API client, and starts the
+    watch task.
```
consideRatio, line 35: No longer relevant, right?

```suggestion
```

kubespawner/reflector.py
```diff
@@ -27,6 +28,15 @@ class ResourceReflector(LoggingConfigurable):
...
+
+    Shutting down a reflector should be done by awaiting its stop() method.
+    JupyterHub does not currently do this, so the watch task runs until the
+    Hub exits.
```
consideRatio, line 39: Is this relevant or correct still?

kubespawner/reflector.py
consideRatio, line 35: Hmmm...  I wonder if we may want to make this a `create()` be `async_init_and_start()` or similar?

```suggestion
    Creating a reflector should be done with the create() classmethod,
    since that, in addition to creating the instance starts the watch task.
```

kubespawner/reflector.py
consideRatio, line 39: I figure JupyterHub shouldn't do this, its not aware of k8s or reflectors etc, only KubeSpawner is, so its a KubeSpawner topic only. The question is, should we make KubeSpawner do this? Maybe. If so, how? Via an async destructor/finalizer?

```suggestion
    Shutting down a shared reflector should be done by awaiting its stop()
    method. KubeSpawner doesn't gracefully do this currently, so the watch
    task runs until the Hub exits.
```

kubespawner/reflector.py
```diff
@@ -203,12 +216,28 @@ def __init__(self, *args, **kwargs):
         if not self.list_method_name:
             raise RuntimeError("Reflector list_method_name must be set!")
 
-        self.start()
+        self.watch_task = None  # Test this rather than absence of the attr
+
+        # Note that simply instantiating the class does not load
+        # Kubernetes config, acquire an API client, or start the
+        # watch task.
```
consideRatio, line 223: There is a comment about this in the docstring, let's settle for that.

```suggestion
```

kubespawner/spawner.py
```diff
@@ -227,25 +197,55 @@ def __init__(self, *args, **kwargs):
...
-            global_conf = client.Configuration.get_default_copy()
-            global_conf.host = self.k8s_api_host
-            client.Configuration.set_default(global_conf)
+    async def _async_init(self):
+        """
+        This method is scheduled to run from `__init__`, but not awaited there
+        as it can't me marked as async.
+
+        Since JupyterHub won't await this method, we ensure the async methods
+        JupyterHub may call on this object will await this method until
```
consideRatio, line 206:
```suggestion
        as it can't be marked as async.

        Since JupyterHub won't await this method, we ensure the async methods
        JupyterHub may call on this object will await this method before
```

kubespawner/proxy.py
```diff
@@ -97,63 +96,74 @@ def _namespace_default(self):
...
             parent=self, namespace=self.namespace, labels=labels
         )
-        self.service_reflector = ServiceReflector(
+        self.service_reflector = await ServiceReflector.create(
             parent=self, namespace=self.namespace, labels=labels
         )
-        self.endpoint_reflector = EndpointsReflector(
+        self.endpoint_reflector = await EndpointsReflector.create(
             parent=self, namespace=self.namespace, labels=labels
         )
```
consideRatio, line 153: We could make this less serial / possibly more performant by using `gather`? What do you think @athornton?

athornton:
OK, I will go do the batch suggestion approval.  I want to change the KubeSpawner-doesn't-call-stop() doc a bit.  stop() should be called to gracefully stop a reflector.  KubeSpawner doesn't do so because its reflectors are singletons and therefore it doesn't know whether or not another spawner instance is using them.  I guess we could add a reference count to the reflector or something, but that seems like a lot of hassle.  Leaving the task running while the Spawner exists is, I think, fine.  Anyway...I'll go approve these and then reword that.

athornton:
There (as far as I know) isn't a way to do async destructors.  This actually is the source of some of the log ugliness when you shut down a JupyterHub with the asyncio KubeSpawner: aiohttp has open connections that are rudely severed.  If we were using the async API object as a context manager, that problem would go away, but that would prevent us from sharing the client, and then we'd have to pay the TLS setup overhead on each method.  I don't want to commit to doing that until I can do some real scale testing (as with K8s object deserialization, I have a suspicion that it's something that's going to look fine in testing and then become problematic at a few hundred users).

consideRatio:
@athornton created https://github.com/lsst-sqre/kubespawner/pull/11 as a PR to this PR about using gather when creating three reflectors.

consideRatio:
@athornton https://github.com/jupyterhub/kubespawner/pull/563#pullrequestreview-887586052 didn't make it into the batch commit!

athornton:
Arrrgh.  Dunno how I missed that one.

athornton:
It always passes here for me.  It must be something I've done wrong in my local environment, and is not relevant.

consideRatio (approved):
With the last fixes in the PR to this PR https://github.com/lsst-sqre/kubespawner/pull/12, this looks great to me and is ready for a final review.

consideRatio:
## To squash or not?

@athornton I'm not sure about my understanding. Will the squashing the back and forth commit changes reduce complexity for a person rebasing another PR on this merged PR? Thats the key value I think there is, but I'm not sure it helps. Do you know?

@minrk/@yuvipanda do you have input about squashing this PR or not?

## Final review request

@minrk/@yuvipanda could you do a final review pass? I would like another person to hit approve besides me. Note that the PR description is updated to reflect the current state.

athornton:
@consideRatio as you may have noticed, I'm not very good at git, and my background is...not that of a traditional developer.  So please don't take my words as representative of what real developers from the modern world want.

I generally would prefer the whole messy commit history.  Almost all of the time I don't use it: I just work with the tip of the branch.  But when there *is* something where I want to see how it was changed, then having that history can be helpful.  A case in point would be whether we squash around the introduction/renaming/removal of the instantiation classmethod.

In most cases, I don't think it's going to matter.  But I can imagine some scenarios in which a developer is confused because things aren't getting initialized.  Probably they'll do fine by reading the comments and realizing they need to use the decorator, but I can also see there being value in watching the back-and-forth from sync K8s client to classmethod to renamed classmethod to return-future-and-await-it-on-method-use.  Knowing the blind alleys someone else went down sometimes helps me avoid having to explore them myself.  But then, as someone who failed to get his Ph.D. in the History of Computation, I can tell you that if I had to summarize the field in one sentence, it's that we've made all the same mistakes in each of the mainframe, mini, micro, and mobile eras because everyone would (understandably!) rather start coding than read the literature.

So I guess my tl;dr is, I don't think squashing would help me, but it may very well help someone who has internalized better development practice.  I am happy to do whatever the three of you think is right.

consideRatio:
I think what is most important is that you feel good about what is done, not that its merge with or without squash - so I'm  💯  for not squashing!

manics:
@droctothorpe @remche You've both contributed to KubeIngressProxy in the past year. Would you mind checking this PR hasn't broken anything as we don't have any tests for it?

minrk (approved):
This looks great. Thanks for all the effort and review, folks!

kubespawner/proxy.py
```diff
@@ -97,23 +96,45 @@ def _namespace_default(self):
...
+        But, how do we figure out the methods to decorate? Likely only those
+        exposed by the base class that JupyterHub would know about. The base
+        class is Proxy, as declared in proxy.py:
+        https://github.com/jupyterhub/jupyterhub/blob/HEAD/jupyterhub/proxy.py.
+
+        From the Proxy class docstring we can conclude that the following
+        methods, if implemented, could be what we need to decorate with
+        _await_async_init:
+
+          - get_all_routes (implemented and decorated)
```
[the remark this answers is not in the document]
consideRatio: Thanks Yuvi for reviewing this!

This is a sound concern but I'm failing to address it with the suggestion as a blob of complexity makes it hard for me to even describe why: but I'll try below.

`shared_client` is now used from `_async_init` in KubeSpawner (where `self.api` is set) and KubeIngressProxy (were `self.core_api` and `self.extension_api` is set). `load_config` is awaited just before calling `shared_client`, so from that perspective it's not an issue of making `shared_client` a async getter, but doing so would not resolve the situation. We still need `_async_init` to finish for various other methods, because they reference `self.api` / `self.core_api` - so we need to ensure those won't run before we have initialized `self.api`.

# Ideas considered

- __Idea 1__
  To ensure that when we need to use an api client, we don't rely on something set on the instance like `self.api`, but request it fresh from `shared_client`.
  - Ugly detail: `k8s_api_host` traitlet's value for example is needed by `load_config`, which now is to be part of `shared_client`'s logic. This traitlet's value is not available to the reflector classes even though they are part of KubeSpawner/KubeIngressProxy. This is likely addressable, but I find it to be a bit tricky on passing the relevant information onwards to the reflector and I'm not sure how to do it in a relatively clean way.
  - Ugly detail: requesting the api client on demand, also now `async`, means we need to await the client first. Following that we also await some some function on it. This is not a technical problem, just bloats the code a bit.
- __Idea 2__
  We ensure that whenever we want to use `self.api` etc from a function we have defined, we do it from a function decorated with `_await_async_init`. That way, we have control within this project when it may be needed. This idea benefits from the overview below about where things are used.
- __Idea 3__
  We double down on the current solution, where we have relied on understanding JupyterHub's Spawner/Proxy base classes which is the interface JupyterHub use to work against our code in this project. While we rely on JupyterHub's internal working, we rely on the public interface we expose to JupyterHub by being a Spawner class (and Proxy class).
  
  I think this could lead to some documentation improvements in JupyterHub. I believe an overview of all interactions JupyterHub has with the Spawner / Proxy class can be improved to for example cover all hooks and functions in a single page for situations like this.

# Where are things used?

## `shared_client` is used from...

spawner.py:
- `_async_init`

proxy.py:
- `_async_init`

reflector.py:
- `__init__` (`load_config` is assumed to have run here)

## The api clients are used from...

`self.api` in spawner.py is used from these functions:
- `_async_init` (initialized)
- `_make_create_pod_request`
- `_make_create_pvc_request`
- `_ensure_not_exists`
- `_make_create_resource_request`
- `_make_delete_pod_request`
- `_make_delete_pvc_request`
- `_ensure_namespace`

`self.core_api` and `self.extension_api` in proxy.py is used from these functions:
- `_async_init` (initialized)
- `add_route`
- `delete_route`

`self.api` in reflector.py is used from these functions:
- `__init__` (initialized)
- `_list_and_update`
- `_watch_and_update`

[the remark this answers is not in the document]
athornton: Yes, as @consideRatio ratio said, there are two loosely-coupled problems here.

1) `shared_client` needs to know the traitlet values in order to mess with the TLS settings
2) The client has to be configured before use, but we don't know what method will be first used and we don't know that it even will be one of the documented methods rather than something happening in a hook.

It feels like there's something we can do with making a SharedClient mixin class that the Spawner and Proxy both inherit...

...but I think maybe the thing to do is go with what we have now and put that into the whole bundle of investigation about how much TLS setup really costs.  Conceptually it would be a lot cleaner to NOT share the client and treat it as an async context manager, which is how the underlying library wants you to use it.  I did not go with this approach because while I'm sure it's going to look fine in the test cases, I really want to try that out at n=500 or so in order to see whether it scales.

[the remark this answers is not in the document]
athornton: ...and even if it's a mixin there's still the problem that sooner or later you're going to have to `await()` the configuration one way or another, and thus that can't happen from `__init__()`.  I think I am in favor of changing the calling conventions from JupyterHub to do something like call a `create()` classmethod, if such exists, instead of just instantiating a new instance of the class, but I don't know how feasible asking for that upstream would be.

kubespawner/proxy.py
```diff
@@ -128,32 +149,26 @@ def __init__(self, *args, **kwargs):
...
-        except kubernetes.config.ConfigException:
-            kubernetes.config.load_kube_config()
-        if self.k8s_api_ssl_ca_cert:
-            global_conf = client.Configuration.get_default_copy()
-            global_conf.ssl_ca_cert = self.k8s_api_ssl_ca_cert
-            client.Configuration.set_default(global_conf)
-        if self.k8s_api_host:
-            global_conf = client.Configuration.get_default_copy()
-            global_conf.host = self.k8s_api_host
-            client.Configuration.set_default(global_conf)
```
consideRatio, line 152: I'm a bit confused about this section in general, and I'm looking to learn more in case it opens up a possibility to refactor this smarter.

1. What is the difference between `load_incluster_config` and `load_kube_config`?
2. Do we need to have the config loading happen before we do the latter part, or can we switch the ordering?

athornton: `load_incluster_config` is what JupyterHub will use.  It relies on the pod's serviceaccount credentials, automatically mounted.  `load_kube_config` is what is used if you're using a .kubeconfig file, and is how running the test via pytest works.

I think we have to set the TLS stuff *after* loading the configuration, because I believe that loading the config overwrites all configuration settings, but I have not verified that.

athornton: This isn't encouraging from upstream (granted, the sync client, but even so...):

https://github.com/kubernetes-client/python/issues/1005

like, everyone agrees it'd be nice to have a method that does the fallback, but no one has actually put it into the base client yet.

athornton: I think I'm going to try the mixin class approach and move _await_async_init into there.

kubespawner/proxy.py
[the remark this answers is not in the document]
consideRatio: @minrk would it be reasonable to make JupyterHub look for if the Spawner / Proxy class provides a async constructor, and if so, make JupyterHub await it?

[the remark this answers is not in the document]
athornton: Specifically, one with a well-known name (I like `create()`).  The logic would look like:

for a thing with a specified class (e.g. Spawner, Proxy, dunno if there's anything else):

Does that class have an async  classmethod named `create()`?  If so, the instance of the class is created with `await kls.create()`; if not it's just `kls()` (the current behavior)

athornton:
https://github.com/jupyterhub/jupyterhub/pull/3801 is a draft of what I mean (untested as of yet, fresh off the neurons).

kubespawner/proxy.py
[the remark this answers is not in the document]
consideRatio: I inspected how JupyterHub interacted with an initialized Spawner object, and it seems its quite hard to overview. I think the right call is to go with Idea 2 - to decorate all the methods in KubeSpawner we know will end up needing to await the `_async_init` call that initialized the `shared_client` ahead of time before it is used.

## JupyterHub interactions with Spawner class instances

The [`_new_spawner` function](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L379) is used as a (sync) factory function.

It is used only once, [in the (sync) `__init__ function](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L242) of the User class, which has a [`_SpawnerDict` class instance](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L210-L217) create a Spawner class whenever the `_SpawnerDict` instance, now available in the user instance `spawners` property is referenced. For example, via `my_user.spawners["any_key_here"]` - then a Spawner instance is returned in a sync manner.

Since JupyterHub inits Spawner instances from sync functions, and sync properties indirectly, it seems like a relatively major change to rely on a async factory function.

## Observed functions used by User object

- [`_delete_spawner` function](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L344) calls:
  - `delete_forever()`
- [`running` property](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L457-L462) could init a Spawner object
- [`server` property](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L487-L489) could init a Spawner object
- [`spawn` function](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L605) calls:
  - `clear_state()`
  - `pre_spawn_start()`
  - `run_auth_state_hook()`
  - `run_pre_spawn_hook()`
  - `create_certs()`
  - `move_certs()`
  - `start()`
  - `start_polling()`
- [`stop` function](https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L885) calls:
  - `stop_polling()`
  - `poll()`
  - `stop()`
  - `run_post_stop_hook()`
  - `clear_state()`
  - `get_state()`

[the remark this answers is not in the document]
consideRatio: @yuvipanda @athornton I've implemented `Idea 2` in https://github.com/lsst-sqre/kubespawner/pull/13, a PR to this PR.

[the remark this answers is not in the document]
consideRatio: @minrk @yuvipanda @athornton I'm really struggling to get somewhere. I'm running low on steam =/ I've run into test failures in https://github.com/lsst-sqre/kubespawner/pull/13.

The clean way of ensuring we await the async init is simply to do it from JupyterHub with a custom async constructor I think... I looked into the JupyterHub code base about that recently, and it seems quite tricky as well. See https://github.com/jupyterhub/jupyterhub/pull/3801 PR for comments about it.

athornton:
Given that this is turning out to be really gross, should we just go with what we have and address a cleaner solution in 3.1 or something?

yuvipanda (approved):
Oh wow, this ended up being far more complicated than I had imagined. THANK YOU for explaining and digging into it, @athornton and @consideRatio. Given the complexity I agree we don't need to make this PR wait on that to be resolved - the current solution is a pretty cool hack I deeply respect too (although we should remove it).

minrk:
🎉 🎆 👏 yay!

If it's really just the async-config-loading that's causing the biggest problem, we can make it sync by running it in a thread. It's a bit silly for that to be async, anyway.

athornton:
I mean that does have the effect of condensing much of the complexity into a single terrifying line....

athornton:
Huge thanks to all of you for your encouragement, persistence, and patience!

yuvipanda:
@minrk I 100% agree it's silly for it to be async! I like your solution :D

consideRatio:
> @minrk I 100% agree it's silly for it to be async! I like your solution :D

I was thinking it was silly as well, which made me track down this which is the reason `load_kube_config` has to be async - because this `_load_authentication` is async - which it seems to need to be because of credential helpers may need to do slow work. Now I'm just troubled that we have a load of work related to it - and it won't even matter for anyone using kubespawner from a k8s Pod with a mounted ServiceAccount as we do in the JupyterHub Helm chart.

From https://github.com/jupyterhub/kubespawner/pull/563#discussion_r811299522.

> * Q: Why is `load_kube_config` async in `kubernetes_asyncio`?
>   A: Because it may reference user credentials to be used, but not declared explicitly there. See the [`_load_authentication` function](https://github.com/tomplus/kubernetes_asyncio/blob/c88b3b9c78a9388ef3af00858b8c516b84b0bf30/kubernetes_asyncio/config/kube_config.py#L183-L195)

consideRatio:
:heart: :tada: ahhhhhhh!

So what about the solution of...

1. Make `load_kube_config` for out-of-cluster kubespawner's blocking with the trick described above
2. Ensure we only have to do that once `load_kube_config` once with some boolean

If we do that, I think we are good to go and remove all _async_init logic across the board!

PR #565, opened 2022-02-04
Support {server_base_url} variable substitution

yuvipanda:
Very helpful when you need to tell some sidecars containers,
or even set some specific environment variable, where the full
URL of the server itself must be known.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -351,7 +351,7 @@ class KubeSpawner(Spawner):
         Template to use to form the namespace of user's pods (only if
         enable_user_namespaces is True).
 
-        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`, `{server_base_url}`,
         `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
         found within strings of this configuration. The username and servername
         come escaped to follow the [DNS label
@@ -427,7 +427,7 @@ class KubeSpawner(Spawner):
         The working directory where the Notebook server will be started inside the container.
         Defaults to `None` so the working directory will be the one defined in the Dockerfile.
 
-        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`, `{server_base_url}`,
         `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
         found within strings of this configuration. The username and servername
         come escaped to follow the [DNS label
@@ -482,7 +482,7 @@ class KubeSpawner(Spawner):
         help="""
         Template to use to form the name of user's pods.
 
-        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`, `{server_base_url}`,
         `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
         found within strings of this configuration. The username and servername
         come escaped to follow the [DNS label
@@ -510,7 +510,7 @@ class KubeSpawner(Spawner):
 
         e.g. 'jupyter-{username}--{servername}.notebooks.jupyterhub.svc.cluster.local',
 
-        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`, `{server_base_url}`,
         `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
         found within strings of this configuration. The username and servername
         come escaped to follow the [DNS label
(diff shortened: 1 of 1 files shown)


yuvipanda:
Well this doesn't work of course, because you might not know the server_url before starting... I think me and @minrk chatted about a related issue as well

consideRatio:
Hmmmm... Sometimes you can mount k8s metadata as an env, for example, maybe about local IP address. Maybe that would be relevant?

minrk:
Do you mean the public URL, or the internal URL? `self.server.url` is where the _internal_ URL is stored (what proxy, hub connect to). You're right that that db record won't be populated until some time after `start` returns.

The public URL is (partly) knowable, and is accessible as `user.server_url(server_name)`. If you aren't using subdomains, this will only be the path (`/user/name[/server_name]`), you'll have to get the `proto://host` from another source (not usually part of JupyterHub config, since it doesn't know or need to know what proxies, etc. might be involved in the actual public endpoint).

The internal connect URL _may_ also be knowable, though it is not by default (e.g. using the pod name via DNS instead of the not-yet-assigned pod ip). If you specify `get_pod_url` as a function that doesn't take the pod into account, e.g.

```python

# this should work by default, but is not the default behavior
def get_pod_url(spawner, pod):
    return f"http://{spawner.dns_name}:{spawner.port}{spawner.user.server_url(spawner.name)}"

c.KubeSpawner.get_pod_url = get_pod_url
```

But because we _can_ use the pod itself (and do by default by using pod.status.podIP) to derive this URL in general, we can't _assume_ `_get_pod_url()` can be called before the pod exists.

minrk:
Ah, If you just mean the server URL prefix (i.e. `/user/username[/servername]/`) that's accessible as `self.server.base_url` and is known ahead of time. I think it makes sense to have this as `{base_url}` in the template namespace.

yuvipanda:
@minrk aha, that's what I actually want! I pushed in a change using that.

yuvipanda:
@minrk do you know if these test failures are mock failures or real failures?

kubespawner/spawner.py
```diff
@@ -1724,6 +1724,7 @@ def _expand_user_properties(self, template):
             servername=safe_servername,
             unescaped_servername=raw_servername,
             hubnamespace=hub_namespace,
+            server_base_url=self.server.base_url,
```
minrk, line 1727: FWIW, I'd call this `base_url` since `server_base_url` for the server feels redundant to me.

minrk (commented):
> do you know if these test failures are mock failures or real failures?

_Mostly_ Mock, but it depends when this might be called. Spawner.server is set prior to calling `Spawner.start()`, and cleared when it's finished shutting down. So if you are calling it somehow in init (it's not typical for Spawners to _have_ non-empty init, but kubespawner is an exception due to the reflectors), it would be a bug here, but if you're calling it only in methods that are called as part of a start, it's because the test context is not realistic.

You can pass a server to the mocks explicitly:

```python
from jupyterhub.objects import Server
server = Server.from_url("http://127.0.0.1:8888/user/whatever")
```

or make the use of base_url conditional with a default value `if self.server is None`, [like we do in base Spawner.get_env](https://github.com/jupyterhub/jupyterhub/blob/39da98f13345fa1bee1cfae34bf9bf7d748ab438/jupyterhub/spawner.py#L837-L845):

```python
        if self.server:
            base_url = self.server.base_url
        else:
            # this should only occur in mock/testing scenarios
            base_url = '/'
```

consideRatio (commented):
@yuvipanda I'm looking into the various PRs of kubespawner as part of planning a release (3.0.0). This change as it now is, where we expose the `base_url` from the associated server, seems fine to me.

I'm vague on what I think with regards to @minrk who considered if it should be named `server_base_url` or `base_url` - I'm open to both.

If this PR is pushed towards having successful tests etc, I consider this to be ready for merge! It would be great to not need to make a new release with it later as it is a bit of a chore to both update z2jh + kubespawner with changelogs etc.

yuvipanda:
@minrk I want to possibly support `hub_base_url` as well in the future, which is why I specifically said `server_base_url` here.

@consideRatio what kinda test would you like for this? I see we already have https://github.com/jupyterhub/kubespawner/blob/88fd98125052612f177ee24d003270ce8429dd5e/tests/test_spawner.py#L584 which tests the process of variable expansion, but doesn't test individual expansion variables. Should I add to that?

consideRatio:
Thanks for following up on this @yuvipanda!

Regarding tests, I mainly looked to see that the existing tests went green. I haven't looked into the failures more than conclude that it seemed related to the changes made in the PR. If you want to add a dedicated test for this or add some aspect to a test go for it but I didn't think of it as required for a merge.

yuvipanda:
oh lol I didn't actually notice these tests were failing! Sorry I lost context. I'm just going to convert this to draft, as I don't have time to look into this right now.

manics:
The tests are failing on main too https://github.com/jupyterhub/kubespawner/actions?query=branch%3Amain (as well as other PRs such as https://github.com/jupyterhub/kubespawner/pull/572) so something's broken somewhere.

consideRatio:
@manics the main branch failures are related to https://github.com/jupyterhub/kubespawner/issues/582 and the new release of JupyterHub 2.2.0 that introduced a change that breaks tests in KubeSpawner.

There were also a few failures that has been resolved by a quickly fixed regression in `kubernetes_asyncio`.

This PR though has more tests failing than those on the main branch, directly related to accessing `self.server.base_url` when `self.server` was returning `None`.

yuvipanda:
I'm going to let this one go! Thanks for the comments everyone.

PR #568, opened 2022-02-18
Add a long docstring to KubeIngressProxy class

### LINKED ISSUE #163: KubeIngressProxy docs

consideRatio:
For a long time I've wanted to document the `KubeIngressProxy` class that ships with this project even though its not needed for KubeSpawner. In the past I've often been to confused to do so, but at this point I felt ready with a better comprehension of the class, its background, and role.

This isn't perfect, and relies on my memory of conversations. @yuvipanda can probably improve this documentation further if time permits.

I think it's an important improvement not only for KubeIngressProxy curious people, but especially for anyone contributing with maintenance of KubeSpawner itself that finds this code - they will need to understand the entire code base to know they are not making breaking changes.

Closes #163 for now I think.

### DIFF
--- a/kubespawner/proxy.py
+++ b/kubespawner/proxy.py
@@ -45,15 +45,19 @@ class KubeIngressProxy(Proxy):
     """
     DISCLAIMER:
 
-        This class is not maintained thoroughly with tests and documentation, or
-        actively used in any official distribution of JupyterHub.
+        This JupyterHub Proxy class is not maintained thoroughly with tests and
+        documentation, or actively used in any official distribution of
+        JupyterHub.
 
-        When it was originally developed and piloted by Yuvi (@yuvipanda), it is
-        my (@consideRatio) unverified understanding that it was found to not be
-        reliable, responsive, or performant enough compared to having a
-        dedicated configurable proxy managed by JupyterHub that routed traffic
-        to users and services - something that ships by default in the
-        JupyterHub Helm chart.
+        KubeIngressProxy was originally developed by @yuvipanda in preparation
+        to support a JupyterHub that could end up with ~20k simultaneous users.
+        The single ConfigurableHTTPProxy server that is controlled by the
+        default JupyterHub Proxy class could bottleneck for such heavy load,
+        even though it has been found to be good enough for ~2k simultaneous
+        users. By instead using this JupyterHub Proxy class that would create
+        k8s resources (Ingress, Service, Endpoint) that in turn controlled how a
+        flexible set of proxy servers would proxy traffic, the bottleneck could
+        be removed.
 
         KubeIngressProxy's efficiency relates greatly to the performance of the
         k8s api-server and the k8s controller that routes traffic based on
@@ -98,17 +102,14 @@ class KubeIngressProxy(Proxy):
     FIXME: Verify what k8s RBAC permissions are required for KubeIngressProxy
            to function.
 
-           Preliminary one can note that there is an IngressReflector,
-           ServiceReflector, and EndpointsReflector. So at least permission to
-           read/list/watch those resources would be needed.
+           - The IngressReflector, ServiceReflector, and EndpointsReflector
+             require permission to read/list/watch those resources.
+           - `add_route` and `delete_route` requires permission to
+             create/update/patch/delete Ingress, Service, and Endpoints
+             resources.
 
-           For Ingress resources, one would also need the ability to create,
-           patch, and delete them, as concluded by inspection of `add_route` and
-           `delete_route`.
-
-           Without having verified these permissions are sufficient, it looks
-           like these permissions are needed on a k8s Role resource bound to the
-           k8s ServiceAccount (via a k8s RoleBinding) used on the k8s Pod where
+           These permissions are needed on a k8s Role resource bound to the k8s
+           ServiceAccount (via a k8s RoleBinding) used on the k8s Pod where
            JupyterHub runs:
 
            ```yaml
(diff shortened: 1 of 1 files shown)


consideRatio:
Pinged @GeorgianaElena for review for your experience with JupyteHub Proxy classes and @yuvipanda for being the author of this Proxy class.

kubespawner/proxy.py
```diff
@@ -44,6 +44,90 @@ def endpoints(self):
...
+        reasons related to sharing code with KubeSpawner in reflectors.py, it
+        has been made and remained part of the jupyterhub/kubespawner project.
+
+        Related issues:
+        - Need for tests: https://github.com/jupyterhub/kubespawner/issues/496
+        - Need for docs:  https://github.com/jupyterhub/kubespawner/issues/163
+
+    ---
+
+    KubeIngressProxy is a implementation of a JupyterHub Proxy class that
```
GeorgianaElena, line 77:
```suggestion
    KubeIngressProxy is an implementation of the JupyterHub Proxy class that
```

kubespawner/proxy.py
```diff
@@ -44,6 +44,90 @@ def endpoints(self):
...
+    JupyterHub can be configured to rely on:
+
+        c.JupyterHub.proxy_class = "kubespawner:KubeIngressProxy"
+
+    The idea of KubeIngressProxy is that, like all JupyterHub Proxy
+    implementations, will react to requests like `get_all_routes`, `add_route`,
+    and `delete_route` in a way that ensures traffic gets routed to the user
+    pods or JupyterHub registered external services. For reference, official
+    documentation on writing a custom Proxy class like this is documented here:
+    https://jupyterhub.readthedocs.io/en/stable/reference/proxy.html.
```
GeorgianaElena, line 87:
```suggestion
    Like all JupyterHub Proxy implementations, KubeIngressProxy will know 
    how to respond to hub requests like `get_all_routes`, `add_route`, and
    `delete_route` in a way that will ensure traffic will get routed to the user
    pods or JupyterHub registered external services. For reference, the official
    documentation on writing a custom Proxy class like this is documented here:
    https://jupyterhub.readthedocs.io/en/stable/reference/proxy.html.
```

kubespawner/proxy.py
```diff
@@ -44,6 +44,90 @@ def endpoints(self):
...
+
+    KubeIngressProxy doesn't route traffic by itself, but instead relies on a
+    k8s cluster's ability to route traffic according to Ingress resources. The
+    only thing KubeIngressProxy does is to speak with a k8s api-server to
+    create/delete such resources.
```
GeorgianaElena, line 92:
```suggestion
    KubeIngressProxy communicates with the k8s api-server in order to 
    create/delete Ingress resources according to the hub's `add_route`/`delete_route`
    requests. It doesn't route traffic by itself, but instead relies on the k8s cluster's
    ability to route traffic according to these Ingress resources.
```

kubespawner/proxy.py
```diff
@@ -44,6 +44,90 @@ def endpoints(self):
...
+
+    Because KubeIngressProxy interacts with a k8s api-server and working with
```
GeorgianaElena, line 94:
```suggestion
    Because KubeIngressProxy interacts with a k8s api-server that manages
```

GeorgianaElena (approved):
This is an awesome thing to do @consideRatio :heart: You're great!

From what I understand, and a proxy perspective, the docstring looks very good and accurate! I've left some comments with suggestions that either add a few details, or just reword the phrases a bit, but this should be good to go!

Thank you!

consideRatio:
@GeorgianaElena thank you so much for your swift, uplifting, and thorough review!!! Excellent suggestions you made!

kubespawner/proxy.py
```diff
@@ -44,6 +44,90 @@ def endpoints(self):
 
 
 class KubeIngressProxy(Proxy):
+    """
+    DISCLAIMER:
+
+        This class is not maintained thoroughly with tests and documentation, or
+        actively used in any official distribution of JupyterHub.
+
+        When it was originally developed and piloted by Yuvi (@yuvipanda), it is
```
yuvipanda, line 53: Ah, I think the understanding instead should be that the default chp was good enough for the eventual traffic circumstances I ended up having (~2k users, not 20k users). It was not less reliable or responsive - just that the chp level of reliability was 'good enough'

kubespawner/proxy.py
```diff
@@ -44,6 +44,90 @@ def endpoints(self):
...
+           JupyterHub runs:
+
+           ```yaml
+           kind: Role
+           apiVersion: rbac.authorization.k8s.io/v1
+           metadata:
+           name: kube-ingress-proxy
+           rules:
+             - apiGroups: [""]
+               resources: ["endpoints", "services"]
```
yuvipanda, line 123: I think it would need rights to create services and endpoints as well.

yuvipanda (changes requested):
Thanks a lot for working on clarifying this, @consideRatio! I've left a couple of clarifications.

https://github.com/berkeley-dsep-infra/data8xhub was the entire config for the infrastructure I'd set up for supporting >20k active users under peak loads, and this was built for that - if you're interested in looking :)

kubespawner/proxy.py
consideRatio: Ah! I rewrote this section!

consideRatio:
@yuvipanda thanks for your input! I've rebased and added 04457e6 with the changes made since your review.

yuvipanda:
Awesome, thanks a lot @consideRatio and @GeorgianaElena!

PR #572, opened 2022-02-22
Add underscore prefix to private functions: `safe_name_for_routespec` and `delete_if_exists`

### LINKED ISSUE #570: maint: add `_` prefix to `load_user_options` and `delete_if_exists`
I think these functions are private, lets mark them with a leading `_` to indicate that as part of going for a KubeSpawner v3 release with the changes in #563.

consideRatio:
Closes #570.

These functions are not overriding a function in a base class right? Also, they are actually supposed to be private, right?

I was considering what functions JupyterHub would call on KubeSpawner and concluded that I wasn't sure what was overriding Spawner functions and what was new private helper functions in KubeSpawner etc. This kind of change would benefit such understanding. It also risks breaking someone that has developed some advanced tweaks to KubeSpawner.

### DIFF
--- a/kubespawner/proxy.py
+++ b/kubespawner/proxy.py
@@ -179,7 +179,7 @@ class KubeIngressProxy(Proxy):
         )
         return safe_name
 
-    async def delete_if_exists(self, kind, safe_name, future):
+    async def _delete_if_exists(self, kind, safe_name, future):
         try:
             await future
             self.log.info('Deleted %s/%s', kind, safe_name)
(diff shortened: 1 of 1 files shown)


manics (commented):
https://github.com/jupyterhub/jupyterhub/blob/2.1.1/jupyterhub/spawner.py#L67 is the abstract class for Spawners, so JupyterHub should never directly call any methods that aren't in that class.

That doesn't mean all other methods should be private though, since as you point out people may want to subclass KubeSpawner- which is why it'd be good in future to include this decision in PR reviews.

I think making these `safe_name_for_routespec` and `delete_if_exists` methods private now is fine and can be documented in the Changelog as a breaking change.

Not sure about `load_user_options` though, especially since the function doc states it's meant to be overridable
https://github.com/jupyterhub/kubespawner/blob/09f4cc58a1a30bbdce29dece3fc48a30272b644e/kubespawner/spawner.py#L2857-L2864

consideRatio:
Excellent points @manics, I'll remove `load_user_options` from being renamed!

@minrk what do you think?

consideRatio:
I removed the commit where prefixed `load_user_options` with an underscore.

minrk (approved):

manics (approved):
I'll let you decide whether to merge, or wait for https://github.com/jupyterhub/kubespawner/issues/582 to be fixed

PR #573, opened 2022-02-22
Followup to kubernetes_asyncio transition change

consideRatio:
I wanted to ensure that JupyterHub wouldn't end up erroring by using a KubeSpawner object that wasn't fully initialized after a trick was introduced in #563 to ensure we await initialization logic that couldn't be awaited in `__init__`.

This PR is inspired by a more thorough inspection of what JupyterHub actually does against a Spawner object, as documented in https://github.com/jupyterhub/kubespawner/pull/563#discussion_r811498699.

I'm mainly concerned about this PR reduce performance by decorating `progress()` which may be called quite often.

### DIFF
--- a/kubespawner/clients.py
+++ b/kubespawner/clients.py
@@ -68,8 +68,18 @@ async def load_config(caller):
     KubeIngressProxy both have these attributes.
     """
     try:
+        # This attempts to load configuration available within k8s Pod's
+        # containers with a mounted service account.
         kubernetes_asyncio.config.load_incluster_config()
     except kubernetes_asyncio.config.ConfigException:
+        # This attempts to load kubernetes client configuration and credentials
+        # to an api-server in a similar way as `kubectl` on your computer would
+        # do which is relevant if KubeSpawner runs from outside a k8s cluster.
+        #
+        # There seems to be a need for this to be async that relates to
+        # acquiring credentials via credential helpers as can be inspected here:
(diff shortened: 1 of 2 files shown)


minrk:
It may be that `_async_init` is not actually the best way to accomplish this. That's the bluntest, simplest way to say "some things are initialized asynchronously, wait for all that to be done". But we can also be more specific and enumerate what we really need to wait for, and what might not be done yet in each method. For example, if they need to access `self.pod_reflector`, that attribute needs to be defined. Right now, that's behind `_async_init`, but it doesn't need to be. If the reflector object exists by the end of `__init__`, we can use the wait we [already have](https://github.com/jupyterhub/kubespawner/blob/4743a226faae96b4a217d2562f1056d4be08a74d/kubespawner/spawner.py#L2068-L2069):

```python
        if not self.pod_reflector.first_load_future.done():
            await asyncio.wrap_future(self.pod_reflector.first_load_future)
```

to ensure the pod reflector has loaded its state at least once, which is the real condition that's important in `.poll()` for example.

Critically, this is a no-op (nothing is awaited) in the 99% case that it's already done (the same approach [is taken in _await_async_init](_await_async_init) where it results in an actual await at most one time).

I don't think referencing the JupyterHub API should be part of what's we need to think about here, because it's all purely internal to KubeSpawner what might not be done yet for each method. Only KubeSpawner knows, and only KubeSpawner should need to know.

consideRatio:
@minrk could you go for a PR about this? I think your ideas are excellent!

consideRatio:
Wieee closing as #576 is the way to go!

PR #576, opened 2022-02-23
Simplify async init

minrk:
By making config-loading sync, almost all async-init is gone.

The only async init is now the first-load of reflector state (clients and reflectors are once again defined in `__init__`), which can be awaited individually as required (rarely needed). We already had a `first_load_future` for this, and had awaits where we needed them, I think.

closes #573

### DIFF
--- a/kubespawner/proxy.py
+++ b/kubespawner/proxy.py
@@ -104,37 +104,7 @@ class KubeIngressProxy(Proxy):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-        # Schedules async initialization logic that is to be awaited by async
-        # functions by decorating them with @_await_async_init.
-        self._async_init_future = asyncio.ensure_future(self._async_init())
-
-    async def _async_init(self):
-        """
-        This method is scheduled to run from `__init__`, but not awaited there
-        as `__init__` can't be marked as async.
-
-        Since JupyterHub won't await this method, we ensure the async methods
-        JupyterHub may call on this object will await this method before
-        continuing. To do this, we decorate them with `_await_async_init`.
-
-        But, how do we figure out the methods to decorate? Likely only those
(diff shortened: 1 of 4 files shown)


kubespawner/proxy.py
```diff
@@ -151,6 +121,28 @@ class is Proxy, as declared in proxy.py:
...
+        Since JupyterHub won't await this method, we ensure the async methods
+        that may be called on this object will await this method before
+        continuing.
+        To do this, we can decorate them with `@_await_async_init`.
+
+        But, how do we figure out the methods to decorate?
+        Any method that accesses the state created in this method.
+
+        Currently, that is only the first-load of our reflectors,
+        which can also be awaited individually with `await reflector._first_load_future`.
```
consideRatio, line 144: The comments here seems out of date with the rest of the changes

kubespawner/spawner.py
```diff
@@ -2677,8 +2656,9 @@ async def _make_delete_pvc_request(self, pvc_name, request_timeout):
             else:
                 raise
 
-    @_await_async_init
+    @_await_pod_reflector
     async def stop(self, now=False):
+
         delete_options = client.V1DeleteOptions()
```
consideRatio, line 2662: Off topic to the PR, but noted right now - this isn't a shared client but an entirely new.

consideRatio (commented):
:heart: omg so much better! So far this looks good - ping me for another review when you have stopped working on it @minrk!

kubespawner/spawner.py
minrk: This isn't a client at all - it's the kubernetes.client package from which we import models.

minrk:
@consideRatio ready for another look, I think.

kubespawner/proxy.py
minrk: Updated to be shorter and hopefully be more precise.

PR #579, opened 2022-02-24
close shared clients when loop closes

### LINKED ISSUE #578: Avoid emitting warnings/errors during normal operation
## Background

@athornton reports there are warnings/errors shown when using KubeSpawner that clutters logs. This issue is meant to help us dig into the details about that. My goal is to resolve it in a way within KubeSpawner that doesn't make us: risk suppressing a relevant warning/error, confuse an end user with irrelevant warnings/errors, maintain a fragile patch making changes to `aiohttp`.

## Technical details

We have switched the library we use to communicate with the k8s api-server to using [`kubernetes_asyncio`](https://github.com/tomplus/kubernetes_asyncio#readme) from [`kubernetes`](https://github.com/kubernetes-client/python#readme). When `kubernetes_asyncio` makes web requests, it uses [aiohttp](https://github.com/aio-libs/aiohttp#client).

- [`kubespawner` use `CoreV1Api` etc](https://github.com/jupyterhub/kubespawner/blob/main/kubespawner/spawner.py#L197) from `kubernetes_asycio`, via [the getter function `shared_client`](https://github.com/jupyterhub/kubespawner/blob/717e7ea65ef8780f0f5b693c35f5f03ada0ba8f8/kubespawner/clients.py#L54).
- [CoreV1Api use `RESTClientObject`](https://github.com/tomplus/kubernetes_asyncio/blob/0f7cb1658c97775fed0bb101b2a6ed225d35470a/kubernetes_asyncio/client/api_client.py#L75) from `aiohttp`
- [RESTClientObject use a `ClientSession` instance](https://github.com/tomplus/kubernetes_asyncio/blob/f028cc793e3a2c519be6a52a49fb77ff0b014c9b/kubernetes_asyncio/client/rest.py#L80-L82) from `aiohttp`.
- [ClientSession has a `__del__` destructor](https://github.com/aio-libs/aiohttp/blob/d01e257da9b37c35c68b3931026a2d918c271446/aiohttp/client.py#L295-L306) that emits warnings and that [raises errors (unsure about this)](https://github.com/aio-libs/aiohttp/blob/d01e257da9b37c35c68b3931026a2d918c271446/aiohttp/client.py#L306).

## Open questions about the warnings/errors

- _When_ do we see these errors when using KubeSpawner?
  - Is it only when we shut down a JupyterHub process?
  - Is it when we a user server?
  - Is it at all seemingly random times?
- _How_ do we see these errors when using KubeSpawner?
  - Is it via logs from the JupyterHub process?
  - How exactly does the logs look as reported from there?

## Open technical questions

- What options do we have?
  1. Patching `aiohttp`
  2. ???

## Related

- A strategy to supress warnings by patching `aiohttp` provided by @athornton in https://github.com/jupyterhub/kubespawner/issues/578.
- #563
- #576
- > The `__del__()` method is a known as a destructor method in Python. It is called when all references to the object have been deleted i.e when an object is garbage collected.
- > The [atexit](https://docs.python.org/3/library/atexit.html#module-atexit) module defines functions to register and unregister cleanup functions. Functions thus registered are automatically executed upon normal interpreter termination. [atexit](https://docs.python.org/3/library/atexit.html#module-atexit) runs these functions in the reverse order in which they were registered; if you register A, B, and C, at interpreter termination time they will be run in the order C, B, A.
  >
  >__Note__: The functions registered via this module are not called when the program is killed by a signal not handled by Python, when a Python fatal internal error is detected, or when [os._exit()](https://docs.python.org/3/library/os.html#os._exit) is called.

minrk:
closes #578 

Closes `api_client` by running a coroutine that will be cancelled when the event loop finishes holding `with client.api_client`. JupyterHub cancels all tasks prior to closing the loop, so this should be okay, but it isn't rigorous for all uses of asyncio.


minrk:
If we duplicate JupyterHub's task cleanup into our test fixtures, I think I can make this work without patching.

tests/test_utils.py
```diff
@@ -120,3 +122,25 @@ def test_get_k8s_model():
         'post_start': None,
         'pre_stop': {'exec': {'command': ['/bin/sh', 'test']}},
     }
+
+
+def test_asyncio_atexit():
```
consideRatio, line 127: Maybe add this to a new test file associated with testing content part of the new file you created? `test_utils.py` with a file named `utils.py` makes me think it only tests things in utils.py.

kubespawner/_asyncio_atexit.py
```diff
@@ -0,0 +1,64 @@
...
+            print(
+                f"Unhandled exception in asyncio atexit callback {callback}: {e}",
+                file=sys.stderr,
+            )
+
+
+def _asyncio_atexit_close(loop):
+    """Patched EventLoop.close method to run atexit callbacks
+
+    prior to the unpatched close method.
```
consideRatio, line 48:
```suggestion
    """Patched EventLoop.close method to run atexit callbacks
    prior to the unpatched close method.
```

consideRatio (approved):
Beautiful! I think this is relatively little intrusive as it just appends some logic while otherwise preserving all logic.

And I understand it as this only becomes relevant when the event loop shuts down - and that should only happen as part of the entire application starts to shut down - right?

Non-critical comments made - this LGTM as a solution to #578.

minrk:
I pushed another commit that avoids any monkeypatching, but instead relies on cleanup of pending tasks, which JupyterHub does in https://github.com/jupyterhub/jupyterhub/pull/2519.

kubespawner/_asyncio_atexit.py
minrk: this black line was added by black, I believe

kubespawner/clients.py
```diff
@@ -34,27 +33,36 @@ def shared_client(ClientType, *args, **kwargs):
...
+        # relies on JupyterHub's task cleanup at shutdown
+        async def close_client_task():
+            try:
+                async with client.api_client:
+                    while True:
+                        await asyncio.sleep(300)
+            except asyncio.CancelledError:
+                pass
+            finally:
+                _client_cache.pop(cache_key, None)
```
consideRatio, line 63: Hmm... You are using `with` as a way to ensure that its `__exit__()` is called (which should be `close()`), but you will also end up calling `__enter__`.

I considered something like this, I'm not at all confident about this suggestion though - probably wrong but I don't get the details here yet.

Perhaps is relate to what CancelledError you expect to be thrown and passed etc as well - perhaps you can be explicit about what you mean to catch->pass there?

```suggestion
            try:
                while True:
                    await asyncio.sleep(300)
            except asyncio.CancelledError:
                pass
            finally:
                await client.api_client.close():
                _client_cache.pop(cache_key, None)
```

minrk: Yes, this is fine, too! `__enter__` only is `return self` so there's no reason to avoid it.

I think the only difference is `with` is the standard Python way to say "clean up after yourself as you see fit", whereas yours explicitly calls `close`, which we know happens to be the cleanup method.

I initially did the `with` because I didn't have the outer try.

consideRatio: Ah, the key aspect I appreciate about my method is that we try-catch the CancelledError specifically for asyncio.sleep() now.

Perhaps we can have `async with client.api_client` in the finally statement?

I want to adopt common practices etc, so let's go with what you think is common practice!

consideRatio: I guess one could call `__exit__()` as well directly? But that may depend on something being done in `__enter__()` so probably a bad idea also.

minrk: I think `with` where supported is the most appropriate way to let things clean up after themselves. It should handle errors, etc. appropriately. Then in our own `try/finally` we do our _own_ cleanup (the cache pop).

The reason we catch and suppress the CancelledError is that our task is actually complete at this point, so we don't need to propagate our CancelledError up to anyone checking if we were interrupted.

consideRatio: With "our task", do you mean `await asyncio.sleep(300)`? I think I understand things a bit better.

Another idea given that we should stick with `with` is this.

```suggestion
            try:
                while True:
                    await asyncio.sleep(300)
            except asyncio.CancelledError:
                pass
            finally:
                async with client.api_client:
                    pass
                _client_cache.pop(cache_key, None)
```

I'm focused on my own and others comprehension of this code only at this point, not a belief that what we have already is problematic etc.

minrk:
The one test failure is a real one and a race condition - the reflector tasks themselves may not be canceled before the close-client is canceled, meaning their requests may fail due to a closed session. Need to think about that a bit. We may need to explicitly shut down the reflectors somehow in the right order (garbage collection ensured the right order, but the gc `__del__` can't be async...).

kubespawner/clients.py
minrk: > With "our task", do you mean

I mean `_client_cache.pop`, that's the only resource we have to clean up, so that's the only code _of ours_ in the finally block.

> I'm focused on my own and others comprehension of this code only at this point

absolutely! I think what I have makes the most sense of anything so far, but I certainly want it to be clear.

I read this code:

```python
 async with client.api_client:
    while True:
        await asyncio.sleep(300)
```

as: "hold the api_client resource forever, then let it cleanup when this coroutine is finished, however and whenever that happens".

Putting `with` in the finally block reads like _acquiring_ resources after we're done, which seems odd (though as we  happen to know, acquisition doesn't do anything, so there's no practical difference, other than clarity). If we want to put the client cleanup explicitly in the finally block, I think calling `close` there is better.

consideRatio: Thanks for helping me understand this better, it reads more clear to me now after having reasoned about it with you!

consideRatio:
@minrk this looks good to me, are you happy with the changes as they are at this point? I'll re-run the test a few times to try catch a potential missed race condition and merge if you are.

minrk:
I _think_ the CancelledError handling will actually solve the races, but I can't be 100% sure. If tests are reliable, let's go ahead.

minrk:
Updated description.

consideRatio:
Wieeee @minrk amazing work! I appreciate seeing this complexity reduction in the project and I love learning from reviewing your work! :heart: :tada:

consideRatio:
I'm not sure how to label things. This was a bug, but it wasn't released, so, should it be marked as a bug for the changelog that is fixed? On the other hand, should it be labelled a bug to know that it solved a bug in case someone used the default branch version or a developer wanted insight about the purpose of this change?

Yikes... Well, I'm going for maintenance now.

PR #598, opened 2022-03-22
[KubeIngressProxy] Migrate to networking.k8s.io/v1 api for Ingress resources

consideRatio:
Closes #597. With the latest release of kubernetes_asyncio version 22 (mapping to k8s 1.22), we had a failure related to outdated k8s API references extensions/v1beta1. This PR makes KubeIngressProxy generate Ingress resources assuming the networking/v1 API  for Ingress resources instead.

Someone installing KubeSpawner 3.0.3 that includes this PR for KubeIngressProxy should clean up all the old Ingress resources for user servers, because they will be orphaned going onwards and could cause collisions etc. I encourage people using KubeIngressProxy to help out documenting more detailed migration steps etc - for example as comments in this PR as it will be linked from the changelog.

This pr should be merged after #599 I think, which it currently includes as well.

PR #599, opened 2022-03-22
ci: stop testinging 1.17-1.19, assume k8s 1.20+ going onwards

consideRatio:
While KubeSpawner 3 supports k8s 1.17-1.19, I'd like to actively stop tests of:
- k8s 1.17-1.18 as those don't include networking.k8s.io/v1 and would cause problems with resolving #598 that has shown up for KubeIngressProxy
- k8s 1.19 as it is no longer supported by cloud providers as discussed in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2591

Due to that, I'd like to drop testing of k8s 1.17-1.19 and rely on tests for k8s 1.20+.

---

My plan is to cut a 3.0.3 release that include a KubeIngressProxy fix for kubernetes_asyncio v22 along with an update to KuberSpawner 3.0.0 release note - adding an post-release edit note - saying that it may work with k8s 1.17-1.19, but use of KubeSpawner 3 should be for k8s clusters 1.20+.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,7 +43,7 @@ jobs:
         include:
           # Tests with oldest supported Python, k8s, and k8s client
           - python: "3.7"
-            k3s: v1.17
+            k3s: v1.20
             test_dependencies: kubernetes_asyncio==19.*
 
           # Test with modern python and k8s versions


yuvipanda:
Thanks @consideRatio

PR #600, opened 2022-03-23
Add changelog for 4.0.0

consideRatio:
This is a small major release, being major because we drop active testing against k8s 1.17-1.19 even though it may very well still function.

I want to cut a release now as there is a critical bugfix to KubeIngressProxy that broke entirely without it.


yuvipanda:
@consideRatio i don't want to retroactively update changelog if possible. Can we just cut 4.0? Major version numbers are free!

consideRatio:
Thanks for review @yuvipanda! I've updated it to cut a major release instead and included most recent PR merge.

yuvipanda (approved):
Thanks, @consideRatio!

I think it's ok for you to self-merge whenever you make the release!

PR #603, opened 2022-03-31
Ensure that shared client has an api_client attr

athornton:


athornton:
Closing; if shared_client shouldn't support being called with "ApiClient" then...that makes the issue go away.

PR #607, opened 2022-04-27
Support dropdown list choices for `profile_list` profiles via `profile_options`

yuvipanda:
Introduces a new 'profile_options' parameter to profile, that allows
rendering a set of `<select>` tags in the profile_list page for
any option the administrator might want to allow the user to
choose from. This allows further customization of the spawned
server without leading to a combinatorial explosion of
profiles.

Each 'choice' in the options has its own set of kubespawner_overrides,
as different choices often necessitate setting multiple kubespawner
traitlet properties - for example, image choices might want to
affect default_url, while memory / CPU limits might affect machine
type selection. The simplest way to deal with this was to introduce
this slightly nested and repetitive config structure, as it is
simple to reason about.

This might mean many hub admins might get by with just a single
profile with a bunch of 'options'. How profiles are designed should
definitely change and become more streamlined with this.

I also clean-up the visual appearance of the profile list a little,
aligning things here and there :)

image

Here's the config that produced this:

```
c.KubeSpawner.profile_list = [
    {
        'display_name': 'Training Env',
        'slug': 'training-python',
        'default': True,
        'profile_options': {
            'image': {
                'display_name': 'Image',
                'choices': {
                    'pytorch': {
                        'display_name': 'Python 3 Training Notebook',
                        'kubespawner_override': {
                            'image': 'training/python:2022.01.01'
                        }
                    },
                    'tf': {
                        'display_name': 'R 4.2 Training Notebook',
                        'default': True,
                        'kubespawner_override': {
                            'image': 'training/r:2021.12.03'
                        }
                    }
                }
            }
        },
        'kubespawner_override': {
            'cpu_limit': 1,
            'mem_limit': '512M',
        }
    }, {
        'display_name': 'Python DataScience',
        'slug': 'datascience-small',
        'profile_options': {
            'memory': {
                'display_name': 'Memory',
                'choices': {
                    '1Gi': {
                        'display_name': '1GB',
                        'kubespawner_override': {
                            'mem_limit': '1G'
                        }
                    },
                    '2Gi': {
                        'display_name': '2G',
                        'kubespawner_override': {
                            'mem_limit': '2G'
                        }
                    }
                }
            },
            'cpu': {
                'display_name': 'CPUs',
                'choices': {
                    '2': {
                        'display_name': '2 CPUs',
                        'kubespawner_override': {
                            'cpu_limit': 2,
                            'cpu_guarantee': 1.8,
                            'node_selectors': {
                                'node.kubernetes.io/instance-type': 'n1-standard-2'
                            }
                        }
                    },
                    '4': {
                        'display_name': '4 CPUs',
                        'kubespawner_override': {
                            'cpu_limit': 4,
                            'cpu_guarantee': 3.5,
                            'node_selectors': {
                                'node.kubernetes.io/instance-type': 'n1-standard-4'
                            }
                        }
                    }
                }
            },
        },
        'kubespawner_override': {
            'image': 'datascience/small:label',
        }
    }, {
        'display_name': 'DataScience - Medium instance (GPUx2)',
        'slug': 'datascience-gpu2x',
        'kubespawner_override': {
            'image': 'datascience/medium:label',
            'cpu_limit': 48,
            'mem_limit': '96G',
            'extra_resource_guarantees': {"nvidia.com/gpu": "2"},
        }
    }
]
```

Fixes https://github.com/jupyterhub/kubespawner/issues/307


yuvipanda:
Another option I considered is allowing items in `kubespawner_override` to be list - however, this doesn't distinguish between us wanting to give the user a list of options, vs when the overriden traitlet itself is a list! So I'm taking this different tack.

yuvipanda:
Although maybe directly implementing something like https://github.com/jupyterhub/kubespawner/issues/307#issuecomment-471653532 is better?

rabernat:
Thanks so much for working on this Yuvi! This would definitely meet the most urgent need in terms of separating hardware from software.


The way you have implemented this, from a UI and configuration perspective, as as a tree.

```mermaid
  graph TD;
      Profile_1 -->Image_A;
      Profile_1 -->Image_B;
      Profile_2 -->Image_C;
      Profile_2 -->Image_D;
```

However, you could invert the hierarchy like this

```mermaid
  graph TD;
      Image_A-->Profile_1;
      Image_A-->Profile_2;
      Image_B-->Profile_3;
      Image_B-->Profile_4;
```

There is no natural "top level" hierarchy here.

In the likely very common case where you want to expose _the same set of images for all profiles_, I still think what i suggested in https://github.com/jupyterhub/kubespawner/issues/307#issuecomment-471653532 is simpler. The main downside of this implementation is that it becomes tedious to maintain consistency when you have many different images and many different profiles. So while the UI may be simpler, the configuration work is only incrementally better than what we have today (explicitly enumerating each image / profile combination).

That said, if https://github.com/jupyterhub/kubespawner/issues/307#issuecomment-471653532 is deemed too complex, I am still very satisfied with this improvement.

yuvipanda:
@rabernat so the question there becomes how would you constrain some profiles to only some set of choices? For example, you don't want the ML image on the non-ML profiles. And some profiles may be only available to some users based on GitHub team choices, and that might further constrain what's made available - particularly around GPUs. So I was thinking of a more expansive tree structure that isn't constrained to just image, but not a graph.

rabernat:
> For example, you don't want the ML image on the non-ML profiles.

This is not exactly true for our case. The ml-notebook image works fine without a GPU. The reason not to prefer it always is just because of the larger image size. 

Perhaps we could have the matrix + one default image for each profile?

Maybe it would be good to get some input from other use cases besides mine.

yuvipanda:
I played around with some more config 'shapes' and settled on one I like that also isn't limited to just images. What do you think now, @rabernat?

kubespawner/spawner.py
```diff
@@ -1486,44 +1527,87 @@ def _validate_image_pull_secrets(self, proposal):
         - `description`: Optional description of this profile displayed to the user.
         - `kubespawner_override`: a dictionary with overrides to apply to the KubeSpawner
           settings. Each value can be either the final value to change or a callable that
-          take the `KubeSpawner` instance as parameter and return the final value.
+          take the `KubeSpawner` instance as parameter and return the final value. This can
+          be further overridden by 'options'
+        ' 'options': A dictionry of sub-options that allow users to further customize the
+          selected profile. Items should have a unique key representing the customization,
+          and the value is a dictionary with the following  keys:
```
consideRatio, line 1534:
```suggestion
        - 'options': A dictionary of sub-options that allow users to further customize the
          selected profile. Items should have a unique key representing the customization,
          and the value is a dictionary with the following keys:
```

kubespawner/spawner.py
```diff
@@ -1486,44 +1527,87 @@ def _validate_image_pull_secrets(self, proposal):
...
+            to set the value for this particular option. The key is an identifier for this
+            choice, and the value is a dictionary with the following possible keys:
+            - 'display_name': Human readable display name for this choice.
+            - 'default': (optional Bool) True if this is the default selected choice
+            - 'kubespawner_override': A dictionary with overrides to apply to the KubeSpawner
+              settings, on top of whatever was applied with the 'kubespawner_override' key
+              for the profile itself. The key should be the name of the kubespawner setting,
+              and value can be either the final value or a callable that returns the final
+              value when called with the spawner instance as the only parameter.
+            for this option, and value
```
consideRatio, line 1546:
```suggestion
```

kubespawner/spawner.py
```diff
@@ -1486,44 +1527,87 @@ def _validate_image_pull_secrets(self, proposal):
...
+          - 'choices': A dictionary containing list of choices for the user to choose from
+            to set the value for this particular option. The key is an identifier for this
+            choice, and the value is a dictionary with the following possible keys:
+            - 'display_name': Human readable display name for this choice.
+            - 'default': (optional Bool) True if this is the default selected choice
+            - 'kubespawner_override': A dictionary with overrides to apply to the KubeSpawner
+              settings, on top of whatever was applied with the 'kubespawner_override' key
+              for the profile itself. The key should be the name of the kubespawner setting,
+              and value can be either the final value or a callable that returns the final
+              value when called with the spawner instance as the only parameter.
```
consideRatio, line 1545: Perhaps this is a good place to clarify how we combine the two kubespawner_override dictionaries. I understand from the text the option is given precedence over the profile level kubespawner_override, but how in detail?

Situation:

1. profile specific overrides
2. option1 specific overrides based on choice a,b,c
3. option2 specific overrides based on choice x,y,z

Do we start out with profile, followed by overwriting the values of the override dictionaries keys from option1, and then option2? The key question to clarify an answer to in documentation is if there is an override or a merge process.

kubespawner/spawner.py
```diff
@@ -2775,9 +2859,24 @@ def _options_from_form(self, formdata):
...
+        # In the default template we use, all form options for a particular profile
+        # come through of the format 'option-{profile}-{option-slug}'
+        if profile:
+            option_formdata_prefix = f'option-{profile}-'
+            for k, v in formdata.items():
+                if k.startswith(option_formdata_prefix):
+                    options[k[: len(option_formdata_prefix)]] = v[0]
+
+        print(options)
+
```
consideRatio, line 2876: Seems like a remnant print statement.

```suggestion
```

consideRatio:
> [...] and settled on one I like that also isn't limited to just images.

YES!!!!!!!!!! This was quite crucial for me to feel comfortable with this PR. I would have feared a influx of requests following shortly after this for the more general functionality anyhow and I would be uneasy to accept the part-way solution.

---

I've opened a PR to fix some details in the CSS, see https://github.com/yuvipanda/kubespawner/pull/1.

rabernat:
👍  this more general approach seems like the right solution.

kubespawner/spawner.py
yuvipanda: @consideRatio I've updated the text here with the hierarchy of overrides.

consideRatio (approved):
Wiee LGTM!

yuvipanda:
Thank you so much for the comments and PR, @consideRatio. I've merged the PR and addressed the comment.

I'm trying to figure out if there is a better word to use than 'options' for these though, as 'options' is also a JupyterHub general feature already.

kubespawner/spawner.py
```diff
@@ -1486,44 +1528,95 @@ def _validate_image_pull_secrets(self, proposal):
         - `description`: Optional description of this profile displayed to the user.
         - `kubespawner_override`: a dictionary with overrides to apply to the KubeSpawner
           settings. Each value can be either the final value to change or a callable that
-          take the `KubeSpawner` instance as parameter and return the final value.
+          take the `KubeSpawner` instance as parameter and return the final value. This can
+          be further overridden by 'options'
+        - 'options': A dictionary of sub-options that allow users to further customize the
+          selected profile. Items should have a unique key representing the customization,
```
consideRatio, line 1534: Mentioning that this is a drop-down list interface is relevant to do here I think.

kubespawner/spawner.py
```diff
@@ -1486,44 +1528,95 @@ def _validate_image_pull_secrets(self, proposal):
...
+          and the value is a dictionary with the following keys:
+          - 'display_name': Name used to identify this particular option
+          - 'choices': A dictionary containing list of choices for the user to choose from
+            to set the value for this particular option. The key is an identifier for this
+            choice, and the value is a dictionary with the following possible keys:
+            - 'display_name': Human readable display name for this choice.
```
consideRatio, line 1540: We have several nested dictionaries: profile, profile option, and profile option choice.

We specify `display_name`, `default`, and `kubespawner_override` in multiple places within these nested dictionaries.

I think this could be the cause of confusion by how similar everything looks even though it is quite logical.

I can live with this as it is and I can't come up with another config schema that I like more but I want to highlight this in case someone comes up with an idea about this.

consideRatio:
> I'm trying to figure out if there is a better word to use than 'options' for these though, as 'options' is also a JupyterHub general feature already.

Hmm so we have `spawner.options_form` and `spawner.options_from_form` which `KubeSpawner.profile_list` is wrapping, and you want to avoid confusion with those mentioned options? I'm not so concerned about that overlap, but independently of my low concern I'll brainstorm a bit.

## Brainstorming

Profile, option, choice... A user choose a profile, and specify profile options in a set of dropdowns each having different profile option choices...

Currently we specify a profile_list where each profile has some details such as display_name, slug, kubespawner_override, and now `options` is introduced. I'm considering a name like `profile_options` just because its verbosity can be helpful, but also in a way thinks its silly because its redundant given that this is a key for a profile already. I'm also considering if we can convey that the options we provide here, are always going to be dropdown lists. I think overall, that is my biggest wish to clarify if anything, that we are limiting the user to provide dropdown lists specifically where one and only one element should be selected.

I think we should stick to the scope of just supporting only drop down lists at the moment, I think it may be a bad idea to let the `options` name be limited by being named to include drop down in some way. I think we should have the name `options` not include `dropdown`, but instead either now or in the future have each option define a type where only `dropdown` is supported currently.

## Conclusion

- I like `options`, but I'm okay with something like `profile_options` or `profile_choices` as well, not feeling confident on the trade offs between clarity by being more verbose vs being inconsistently redundant by adding `profile_`.
- I like if we clarify this the options we currently support is dropdown lists only, and think its reasonable that we adopt a more general name for what is currently called `choices`, such as `values`. and combines this with `type` or `option_type` which for now only can be something like `dropdown`.
- I can imagine the wish to extend this to have checkboxes, radioboxes, free-form input, for example. I think though that for everything besides free-form input, we will manage with dropdowns though. I also think this requires significantly more thought as with for example free-form input, we may end up introducing a lot of additional complexity related to client/server-side validation etc and suddenly this may get out of hand.

## Summary

This is acceptable to me as it is. If the documentation would clarify that the options supported are the drop-down list, it would be an improvement. I made a comment (https://github.com/jupyterhub/kubespawner/pull/607#pullrequestreview-961815057) about where that could make sense to do.

ctr26:
amazing

kubespawner/spawner.py
yuvipanda: Done

yuvipanda:
I've also called these profile_options now to separate them from JupyterHub's native options. I too don't like the duplication but better that than ambiguity I think.

Ready to merge!

PR #612, opened 2022-06-08
Add user_namespace_labels and user_namespace_annotations for use with enable_user_namespaces

zv0n:
I've added option to specify labels/annotations for created user namespaces.

At our organization we use Rancher to manage Kubernetes and we would like to have Jupyterhub user namespaces to be a part of a single Rancher project. This is accomplished via the namespaces' annotations, so it would be very beneficial to us if we could specify annotations for newly created namespaces.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3046,8 +3046,8 @@ class KubeSpawner(Spawner):
     async def _ensure_namespace(self):
         ns = make_namespace(
             self.namespace,
-            self._expand_all(self.user_namespace_labels),
-            self._expand_all(self.user_namespace_annotations),
+            labels=self._expand_all(self.user_namespace_labels),
+            annotations=self._expand_all(self.user_namespace_annotations),
         )
         api = self.api
         try:
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -324,6 +324,40 @@ class KubeSpawner(Spawner):
         """,
     )
 
+    user_namespace_labels = Dict(
+        config=True,
+        help="""
+        Kubernetes labels that user namespaces will get (only if
+        enable_user_namespaces is True).
+
+        Note that these are only set when the namespaces are created, not
+        later when this setting is updated.
+
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
+        found within strings of this configuration. The username and servername
+        come escaped to follow the [DNS label
+        standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).
+        """,
+    )
+
+    user_namespace_annotations = Dict(
+        config=True,
+        help="""
+        Kubernetes annotations that user namespaces will get (only if
+        enable_user_namespaces is True).
+
+        Note that these are only set when the namespaces are created, not
+        later when this setting is updated.
+
+        `{username}`, `{userid}`, `{servername}`, `{hubnamespace}`,
+        `{unescaped_username}`, and `{unescaped_servername}` will be expanded if
+        found within strings of this configuration. The username and servername
+        come escaped to follow the [DNS label
+        standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).
+        """,
+    )
+
     user_namespace_template = Unicode(
         "{hubnamespace}-{username}",
         config=True,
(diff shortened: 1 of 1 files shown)


kubespawner/spawner.py
```diff
@@ -324,6 +328,28 @@ async def async_method(self, *args, **kwargs):
...
+        help="""
+        Kubernetes labels that user namespaces will get.
+
+        Note that these are only set when the namespaces are created, not
+        later when this setting is updated.
+        """,
+    )
+
+    user_namespace_annotations = Dict(
+        {},
```
consideRatio, line 343: I'm not sure, but I think a single `{}` object could be the same object for all spawner instances, which is problematic if its modified later for one spawner instance but not others.

Maybe there is some example of using the `Dict` traitlets configuration object somewhere else you can mimic to avoid this kind of issue.

kubespawner/spawner.py
```diff
@@ -173,8 +173,12 @@ def __init__(self, *args, **kwargs):
         # By now, all the traitlets have been set, so we can use them to
         # compute other attributes
 
+        self.ns_labels = {}
+        self.ns_annotations = {}
         if self.enable_user_namespaces:
             self.namespace = self._expand_user_properties(self.user_namespace_template)
+            self.ns_labels = self.user_namespace_labels
+            self.ns_annotations = self.user_namespace_annotations
```
consideRatio, line 181: To allow expansion of user properties could be relevant, but hmmmm you can leave it out but a comment should be made in the docstring about the state of how this is implemented - will an expansion be made or not?

I guess, the use of `_template` indicated that expansion will take place, and that it really must include something to expand to be username unique.

If you are not expanding things, I suggest removing the intermediary variables `ns_labels` and instead relying on the configuration directly.

zv0n: Hmm, now that I think about it expansion could be very useful, I haven't thought of including that. I did the `ns_labels` and `ns_annotations` mainly because I thought it would look better than having an `if` statement in `_ensure_namespace`.

I think that the expansion could work by iterating through the configuration dictionary and for each key assign expanded value to `ns_labels` or `ns_annotations` under the same key.

Do you think this solution would be acceptable?

kubespawner/spawner.py
zv0n: I wasn't aware that `{}` could refer to the same object in all instances, fixed and thank you for the info!

kubespawner/spawner.py
zv0n: I've added expansion through the `_expand_all` function and use it when creating the namespace, so this should take care of the possibility of different logic based on users as you mentioned.

kubespawner/spawner.py
```diff
@@ -3010,7 +3044,14 @@ async def load_user_options(self):
...
     async def _ensure_namespace(self):
-        ns = make_namespace(self.namespace)
+        if self.enable_user_namespaces:
+            ns = make_namespace(
+                self.namespace,
+                self._expand_all(self.user_namespace_labels),
+                self._expand_all(self.user_namespace_annotations),
+            )
+        else:
+            ns = make_namespace(self.namespace)
```
consideRatio, line 3054: I expect that this function isn't called unless `enable_user_namespace`, right?

I think that the `else` clause will never be executed and that this logic should reduce to

```suggestion
        ns = make_namespace(
            self.namespace,
            self._expand_all(self.user_namespace_labels),
            self._expand_all(self.user_namespace_annotations),
        )
```

Can you check if this guess is correct?

zv0n: You're right! I didn't notice that it's only called when enable_user_namespaces is enabled.

I removed the if statement

kubespawner/spawner.py
```diff
@@ -3010,7 +3044,11 @@ async def load_user_options(self):
             )
 
     async def _ensure_namespace(self):
-        ns = make_namespace(self.namespace)
+        ns = make_namespace(
```
consideRatio, line 3047: Wait, shouldnt make namespace also be updated to support this?

consideRatio: The fact that I discover this like this, even though tests pass, makes me think we should have a test for `make_namespace` like we do for many other functions, like `make_pod`.

zv0n: `make_namespace` already supports labels and annotations

https://github.com/jupyterhub/kubespawner/blob/616f72c4aee26c3d2127c6af6086ec50d6cda383/kubespawner/objects.py#L955

All that was needed was to add config options that would utilize this functionality

kubespawner/spawner.py
```diff
@@ -3010,7 +3044,11 @@ async def load_user_options(self):
...
+            self.namespace,
+            self._expand_all(self.user_namespace_labels),
+            self._expand_all(self.user_namespace_annotations),
```
consideRatio, line 3050:
```suggestion
            labels=self._expand_all(self.user_namespace_labels),
            annotations=self._expand_all(self.user_namespace_annotations),
```

consideRatio (approved):
Thank you @zv0n for your PR and iterating on it with me! This LGTM!

zv0n:
Thank you @consideRatio for responding so quickly and helping me make the code better! :)

PR #614, opened 2022-06-20
Add name to service port for Istio support

ddebeau:
Istio treats all traffic as TCP if the service port name does not start with a protocol. This PR adds a name to the port with the proper protocol.


consideRatio (approved):
Seems fine to me to add the name http as a name to this service port.

PR #617, opened 2022-07-07
Spawning containers in different namespaces without having cluster rights 

schaefer-david:
KubeSpawner offers two modes:

* One namespace is used for all deployments. The service account need then only access rights in this namespace
* KubeSpawner spawns images in multiple namespaces. The service account needs cluster rights in this scenario.

The following scenario is not supported:

There is one major namespace and one namespace which is used, if a pod with special resources, like GPU cards is needed and it is not possible that the service account gets cluster access rights, but has access rights in some namespaces.

This should be addressed with this commit: beside one namespace a list of namespaes could be given which are used for creation of namespaces - depending on the configuration in the profile_list.

Please provide feedback about my approach and if it can be merged in to the kubespawner main repo


manics:
I think the general idea makes sense. From a security perspective having the ability to run the hub container in one namespace and all user containers in another single (or limited number of pre-existing) namespaces is a good idea.

consideRatio:
Hi @schaefer-david and welcome to the open-source world, I see this is the first ever made pull request you make on this GitHub account!

I think your _current proposal_ would be a breaking change, so that if it was accepted, existing installations of this software would break if they upgraded to this. That is a very big deal. But we could perhaps make your wished outcome happen without having a breaking change.

As a maintainer of this project, here are some aspects I consider with regards to accepting a PR:
- breaking changes must be very clearly and strongly motivated
- well documented
- sustainable to maintain long term
- add value motivating the added maintenance burden

---

@schaefer-david is it correct that you look to resolve point 2 mainly below, or is it mainly point 1 below?
1. Be able to split users into grouped namespaces
2. Be able to not use ClusterRole and ClusterRoleBinding at all when granting permissions to KubeSpawner, as they escalate permissions more than accepted, while still segregating user servers into their own namespaces or as a compromise, a few pre-defined namespace where possibly a few resides.

---

Technical deliberations:
- Without access to pods in all namespaces, can we ask the k8s api-server for "all pods in the namespaces we can access" or similar? If that is possible, this is probably far easier to implement than if we need to have N number of reflectors each asking the k8s api-server about things going on in the individual namespaces. This I think would lead to "429 too many requests" responses from the API server quite quickly unless we keep the number of namespaces to 1-3 perhaps.

schaefer-david:
Dear @manics and @consideRatio ,

thanks for your warm welcome! I try to answer your questions:

I found only the methods to retrieve resources for a named namespace or list_[resource_type]_for_all_namespaces. There is nothing between... We can try to implement something in this area:

-  check in which namespaces the service account has access
-  query these namespaces

My purpose for this pull request is not to split users in namespaces - it is more that in our environment certain resources are only available in given namespaces.

If you tell me, where I should documentation, I can do this. On the other side I do not see at the moment the breaking change - because the old behaviour should work as given.

consideRatio:
@schaefer-david documentation will depend on the implementation, but the implementation depends on the strategy to meet the goals etc, and I'm trying to clarify the goal and conclude if a certain implementation strategy seems viable for this project. I may have been wrong about the breaking change looking at this in more detail.

My understanding now is that your goal is to put users in certain namespaces based for unrelated k8s reasons, but are fine with providing KubeSpawner a k8s ServiceAccount with permissions granted via a ClusterRole and ClusterRoleBinding.

If that is the case, that could possibly be implemented differently. If you wish that certain users end up in different namespaces depending on what they choose when they spawn, there are significant additional separate complexities though - what about the storage allocated for them? PVC resources created are local to the namespace, and if they can start in two separate namespaces etc, that could be a problem.

schaefer-david:
Dear @consideRatio ,

thanks for your reply and sorry for my late response.  You are absolutely correct, that pvc resources are only inside one namespace. But this does not matter for this use case.

Maybe I describe the use case a little bit broader:

- normally all containers are spawned in one namespace. Here all pods and pvcs are created
- special resources, like GPU are available only in another namespace. It is not possible to combine both namespaces
- if users need GPU support, they know that they do not have their default pvc. The GPU container is a dedicated additional (named) container
- the service account has only access in this two namespaces...

Do you have another idea then my provided code change how to implement this scenario?

Thanks in advance!

consideRatio:
Thank you for thinking about this @schaefer-david! Does #678 resolve this for you do you think?

You would need to define Role and RoleBinding in the namespaces providing KubeSpawner with sufficient permissions, and let them reference the ServiceAccount in the namespace where jupyterhub/kubespawner runs.

PR #619, opened 2022-07-12
properly handle IPv6 IPs

### LINKED ISSUE #620: Issues with IPv6 only cluster
When trying to start a user session on an IPv6 only EKS cluster the session fails with the error 

Spawn failed: Port could not be cast to integer value as '1f18:171d:402:7f94::1:8888'

It seems to be hanging up on the non-integer values in the IPv6 address. The prefix seems to be missing as well. The hub and CHP are running in IPv6 mode and seems to be working as expected. 

The user session pod does stay running and everything looks OK  

I've tried Kubespawner 3.0.2 and 4.1.0 with both Jupyterhub 1.5.0 and 2.3.1. EKS Cluster is 1.22 with the AWS VPC CNI.

nikhiljha:
Currently, trying to launch JupyterHub on an IPv6 Primary / IPv4 Secondary K8S cluster (PodIP = IPv6) fails.

```
[E .. JupyterHub gen:623] Exception in Future <Task finished name='Task-284' coro=<BaseHandler.spawn_single_user.<locals>.finish_user_spawn() done, defined at /usr/local/lib/python3.8/dist-packages/jupyterhub/handlers/base.py:900> exception=ValueError("Port could not be cast to integer value as ‘ffff:ffff:ffff::ffff:ffff’”)> after timeout
      File "/usr/local/lib/python3.8/dist-packages/tornado/gen.py", line 618, in error_callback
        future.result()
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/handlers/base.py", line 907, in finish_user_spawn
        await spawn_future
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/user.py", line 736, in spawn
        raise e
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/user.py", line 653, in spawn
        port = urlinfo.port
      File "/usr/lib/python3.8/urllib/parse.py", line 174, in port
        raise ValueError(message) from None
    ValueError: Port could not be cast to integer value as 'ffff:ffff:ffff::ffff:ffff'
```

This PR resolves the issue by attempting to parse the PodIP as an IP Address using Python's built in (correct) IP address detection mechanism, and then outputting a correctly formatted URL for IPv6 addresses.

The issue is a regression from https://github.com/jupyterhub/jupyterhub/pull/3020 but the root cause is in kubespawner not formatting the URL correctly.

fixes #620

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1937,7 +1937,7 @@ class KubeSpawner(Spawner):
             hostname = pod["status"]["podIP"]
             if type(ipaddress.ip_address(hostname)) == ipaddress.IPv6Address:
                 hostname = f"[{hostname}]"
-            
+
         if self.pod_connect_ip:
             hostname = ".".join(
                 [
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1935,7 +1935,7 @@ class KubeSpawner(Spawner):
         else:
             proto = "http"
             hostname = pod["status"]["podIP"]
-            if type(ipaddress.ip_address(hostname)) == ipaddress.IPv6Address:
+            if isinstance(ipaddress.ip_address(hostname), ipaddress.IPv6Address):
                 hostname = f"[{hostname}]"
 
         if self.pod_connect_ip:
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -9,6 +9,7 @@ import os
 import string
 import sys
 import warnings
+import ipaddress
 from functools import partial, wraps
 from urllib.parse import urlparse
 
@@ -1944,12 +1945,21 @@ class KubeSpawner(Spawner):
                     )
                 ]
             )
-
-        return "{}://{}:{}".format(
-            proto,
-            hostname,
-            self.port,
-        )
+        
+        fmtstr = "{}://{}:{}"
+        # Parsing as an IP Address may fail if hostname is a DNS name.
+        try:
+            addr = ipaddress.ip_address(hostname)
+            if type(addr) == ipaddress.IPv6Address:
+                fmtstr = "{}://[{}]:{}"
+        except:
+            pass
+        
+        return fmtstr.format(
+                proto,
+                hostname,
+                self.port,
+            )
 
     async def get_pod_manifest(self):
         """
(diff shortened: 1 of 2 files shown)


nikhiljha:
I still need to test this but I cannot figure out where the `k8s-hub` image is built from. Once I do I'll give it a go.

kubespawner/spawner.py
```diff
@@ -1945,7 +1946,16 @@ def _get_pod_url(self, pod):
                 ]
             )
 
-        return "{}://{}:{}".format(
+        fmtstr = "{}://{}:{}"
+        # Parsing as an IP Address may fail if hostname is a DNS name.
+        try:
+            addr = ipaddress.ip_address(hostname)
```
minrk, line 1952: I think rather than try/excepting all possible hostnames, the place for this is in the `podIP` branch above on L1936. Then it's a simpler check `if ":" in hostname: hostname = f"[{hostname}]"`

nikhiljha: Okay, but I think we should still use the `ip_address` parsing thing, just because it's best practice. For example, did you know `::192.168.1.1` is a valid IPv6 address? Admittedly it still has a `:` in it, but... I think we should just throw this problem to the Python standard library.

```python
>>> ipaddress.ip_address("::192.168.1.1")
IPv6Address('::c0a8:101')
```

nikhiljha:
No idea why that one test is failing but it appears to be unrelated(?)

kubespawner/spawner.py
```diff
@@ -1934,6 +1935,8 @@ def _get_pod_url(self, pod):
         else:
             proto = "http"
             hostname = pod["status"]["podIP"]
+            if type(ipaddress.ip_address(hostname)) == ipaddress.IPv6Address:
```
minrk, line 1938: let's use a standard Python type check:

```suggestion
            if isinstance(ipaddress.ip_address(hostname), ipaddress.IPv6Address):
```

consideRatio:
Failures around kubectl exec or kubectl logs are unrelated intermittent issue in the k3s-io/k3s project! Tracked in the github action jupyterhub/action-setup-k3s-helm

minrk:
Thanks!

PR #635, opened 2022-08-26
catch errors in reflector.start

### LINKED ISSUE #627: Resilience needed against k8s master outages
### Bug description

During a kubernetes master outage, a few failed api calls leaves JupyterHub in an unrecoverable state. This causes user servers to fail to spin up.

#### Expected behaviour

JupyterHub can correctly recover itself after failed api calls to the k8s master.

#### Actual behaviour

JupyterHub is left in an unrecoverable state and user servers cannot start.

Some logs are available in the following issues:

- https://github.com/2i2c-org/infrastructure/issues/1248#issuecomment-1110326838
- https://github.com/2i2c-org/infrastructure/issues/1102#issuecomment-1200486502

### How to reproduce

This is an intermittent issue we are seeing on cloud-based deployments.

### Your personal set up

We are running z2jh on GCP clusters. We see this issue most frequently on **non-regional** clusters. Our clusters that are regional do not experience this as often, if at all, because the k8s master has high availability for these clusters.

You can find more information about how we deploy our hubs here: https://github.com/2i2c-org/infrastructure

### Related issues 

- https://github.com/2i2c-org/infrastructure/issues/1102
- https://github.com/2i2c-org/infrastructure/issues/1103
- https://github.com/2i2c-org/infrastructure/issues/1135
- https://github.com/2i2c-org/infrastructure/issues/1248
- https://github.com/2i2c-org/infrastructure/issues/1616

### Versions

- jupyterhub 2.3.1
- kubespawner 4.1.0

Established in this comment: https://github.com/jupyterhub/kubespawner/issues/627#issuecomment-1213322208

minrk:
ensures failure to start reflectors is fatal, rather than resulting in permanent hangs waiting for first_load_future

I _believe_ this closes #627 (it at least explains one such event).

Relevant bits:

- We already have [code](https://github.com/jupyterhub/kubespawner/blob/f3d44e6e51d2ca16e787963562e5568f2a5494e0/kubespawner/spawner.py#L2380-L2385) for reflectors failing to update after a few tries being fatal to the Hub
- reflector.start is [scheduled asynchronously](https://github.com/jupyterhub/kubespawner/blob/f3d44e6e51d2ca16e787963562e5568f2a5494e0/kubespawner/spawner.py#L2396), via `ensure_future`. Exceptions on this future are ignored.
- methods like start/stop/poll [await a first_load_future](https://github.com/jupyterhub/kubespawner/blob/f3d44e6e51d2ca16e787963562e5568f2a5494e0/kubespawner/spawner.py#L212-L213) to ensure the reflector has loaded info at least once
- first_load_future is only set [_on success_](https://github.com/jupyterhub/kubespawner/blob/f3d44e6e51d2ca16e787963562e5568f2a5494e0/kubespawner/reflector.py#L237), it's never set if the first load fails

All these combined mean that if the _first_ load fails, the exception is ignored (it will be logged by asyncio with `Task exception was never retrieved`), and all subsequent calls wait forever instead of raising, never attempting to restart the reflector.

The events described in #627 are:

- reflector starts having trouble talking to k8s (several `iohttp.client_exceptions.ClientConnectorError`)
- it keeps failing, so we get to `Pods reflector failed, halting Hub`
- Hub stops, restarts
- On restart, if we are still in the disrupted state, the _initial_ load fails again with the same `aiohttp.client_exceptions.ClientConnectorError`, but this time it's ignored (`ERROR:asyncio:Task exception was never retrieved`)
- since first_load_future is never set, all start/stop/poll calls are just infinite waits

The fix here is to:

1. catch (any) error in `Reflector.start()`, and
2. call the same fatal cleanup `on_reflector_failure` function
3. store the error in `first_load_future` so that anyone awaiting the first load will get the exception, rather than a forever-incomplete Future

The result should be that the Hub goes into CrashLoopBackoff while it can't talk to the k8s API, which I think is probably correct.

We could also give the first load more retry chances with `exponential_backoff` before it considers errors fatal, but I'm not sure if that's the right thing to do.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2393,7 +2393,16 @@ class KubeSpawner(Spawner):
                 on_failure=on_reflector_failure,
                 **kwargs,
             )
-            asyncio.ensure_future(self.__class__.reflectors[key].start())
+            f = asyncio.ensure_future(self.__class__.reflectors[key].start())
+
+            async def catch_reflector_start():
+                try:
+                    await f
+                except Exception as e:
+                    self.log.exception(f"Reflector for {kind} failed to start.")
+                    sys.exit(1)
+
+            asyncio.create_task(catch_reflector_start())
 
         if replace and previous_reflector:
             # we replaced the reflector, stop the old one
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -378,9 +378,16 @@ class ResourceReflector(LoggingConfigurable):
         and not afterwards!
         """
         if self.watch_task and not self.watch_task.done():
-            raise RuntimeError('Task watching for resources is already running')
+            raise RuntimeError(f"Task watching for {self.kind} is already running")
+        try:
+            await self._list_and_update()
+        except Exception as e:
+            self.log.exception(f"Initial list of {self.kind} failed")
+            if not self.first_load_future.done():
+                # anyone awaiting our first load event should fail
+                self.first_load_future.set_exception(e)
+            raise
 
-        await self._list_and_update()
         self.watch_task = asyncio.create_task(self._watch_and_update())
 
     async def stop(self):


kubespawner/spawner.py
```diff
@@ -2393,7 +2393,16 @@ def on_reflector_failure():
...
             )
-            asyncio.ensure_future(self.__class__.reflectors[key].start())
+            f = asyncio.ensure_future(self.__class__.reflectors[key].start())
+
+            async def catch_reflector_start():
+                try:
+                    await f
+                except Exception as e:
+                    self.log.exception(f"Reflector for {kind} failed to start.")
+                    sys.exit(1)
```
consideRatio, line 2403: I'm not feeling confident on the outcome of having `sys.exit(1)` here yet.

I looked for `sys.exit` within the source code to see if its not the first time this is done, and therefore probably not an issue. Doing that I found this commit which added a reason for not feeling confident that sys.exit should be called: https://github.com/jupyterhub/kubespawner/commit/c15ad869b6c60a79dfa0f727e429ab4750cf84ff

minrk: This is identical to the call directly above which is called for the same reason, but when the error occurs later rather than at start time. We know it works because this is the call that's restarting unhealthy 2i2c Hubs (working as intended).

It was more complicated when we had threads, but we don't anymore. Sys.exit from the main thread always unconditionally exits the process unless unless SystemExit is caught explicitly.

consideRatio: Ah, thanks!

consideRatio (approved):
Wieee nice work resolving this very tricky bug to catch! Thank you @minrk, @sgibson91, @abkfenris, @GeorgianaElena, and @gmanuch for working to resolve this and providing information leading to a resolution!

PR #640, opened 2022-09-28
Allow profile_options callable to be async

yuvipanda:
This allows network requests to be made here, which allows things like dynamically populating a list of tags for a docker image.

Ref https://github.com/2i2c-org/infrastructure/pull/1735#issuecomment-1261146440


GeorgianaElena (approved):

yuvipanda:
Thanks @GeorgianaElena! Think I can get you to hit merge when you get the chance? :)

PR #641, opened 2022-09-29
Add extra_services option

dolfinus:
KubeSpawner has `services_enabled` option which creates a service for the notebook pod. But it only creates service of type ClusterIP, and only opens 80 TCP port.

But sometimes there are running processes in the notebook pod which also need a separated service. For example:
1. [Apache Spark in k8](https://spark.apache.org/docs/latest/running-on-kubernetes.html#client-mode-networking) (client mode) - driver is running in the notebook container and opens several ports, which are then used by executor pods to communicate with a driver
2. Apache Spark with driver in k8s and executors in YARN - same ports are used, but it is required to use `NodePort` instead of `ClusterIP`, allowing access from external network (executors running in YARN cluster) to a driver running in the pod
3. Services running in a `extra_containers` of the same pod, which can listen for some port

Here I've added a new option `KubeSpawner.extra_services` where administrator can describe a list of services:
```
c.KubeSpawner.extra_services = [
  {
    "name": "jupyter-{username}-spark-driver--{servername}",
    "ports": [
      {
        "name": "driver",
        "port": 1000,
        "target_port": 1000,
      },
      {
        "name": "block-manager",
        "port": 1001,
        "target_port": 1001,
      },
      {
        "name": "ui",
        "port": 1002,
        "target_port": 1002,
      },
    ],
]
```

Just the same mechanism as in `services_enabled` is used - this list of services is created while spawning a new notebook, all the services are connected to a created pod using `owner_references` feature.

In addition to service name and port user can set service type, labels, annotations and extra spec options, like `clusterIP`.

Theoretically, values can be set dynamically using `pre_spawn_hook` - for example, in the case with Spark driver in k8s + executors in YARN pre spawn hook can generate random `nodePort`, pass it into a `extra_services` to create a `NodePort` service, and then pass the same port into an environment variable using `extra_container_config`.

Note: `LoadBalancer` service type is not tested is this PR as it requires a very complicated spec, and I don't see any use case for it. But theoretically, it is possible to use it.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -905,11 +905,13 @@ def make_secret(
 
 def make_service(
     name,
-    port,
+    ports,
     servername,
-    owner_references,
+    type="ClusterIP",
+    owner_references=None,
     labels=None,
     annotations=None,
+    spec=None,
 ):
     """
     Make a k8s service specification for using dns to communicate with the notebook.
@@ -919,13 +921,20 @@ def make_service(
     name:
         Name of the service. Must be unique within the namespace the object is
         going to be created in.
-    env:
-        Dictionary of environment variables.
+    ports:
+        List of dict with port description, e.g. `[{"name": "http", "port": 80, "target_port": 80}]`
     labels:
         Labels to add to the service.
     annotations:
         Annotations to add to the service.
-
+    spec:
+        Extra options should be added to a spec, like `{"clusterIP": "None"}`
+    servername:
+        Name of server user by selector
+    owner_references:
+        Pod UUID used by Kubernetes to remove service with the pod
+    type:
+        `ClusterIP` or `NodePort`
     """
 
     metadata = V1ObjectMeta(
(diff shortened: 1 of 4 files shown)


dolfinus:
@GeorgianaElena Could you please review?

consideRatio:
I worry that adding another configuration like this goes beyond what we can maintain sustainably in the KubeSpawner project.

Is there reasonable workarounds to not embedding logic like this in KubeSpawner perhaps?

Ping @GeorgianaElena @manics @minrk for further opinions.

manics:
@consideRatio I thought I'd commented but obviously hadn't.... I'm thinking exactly the same as you!

This might be a good time to consider how to bring in the extensibility mentioned in https://discourse.jupyter.org/t/jupyterhub-amalthea/12208 into KubeSpawner

dolfinus:
`modify_pod_hook` hook cannot be used here because it is called before creating the pod, so `owner_reference` cannot be fetched. There could be some new hook after creating a pod, but I'm not sure is that okay or not

dolfinus:
Ok, implemented `after_pod_created_hook` in #644

dolfinus:
I'm using a slightly modified version of `kubespawner.spawner.get_service_manifest` which calls `kubespawner.objects.make_service` with different port number and service name, and then call k8s api with a copy of this code:
https://github.com/jupyterhub/kubespawner/blob/cd4c08d5e175e3b6d58e279c27265e6e95e6197b/kubespawner/spawner.py#L2741-L2757

dolfinus:
Probably

PR #642, opened 2022-09-30
Expand environment variables

dolfinus:
In #644 spawner can add extra services after pod creation. To pass created service name, I've tried to set up notebook environment like this:
```python
c.KubeSpawner.environment = {"SPARK_DRIVER_HOST": "jupyter-{username}-spark-driver--{servername}"}
```
But unfortunately, environment variables does not support substitutions. Now they are

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ lib64/
 parts/
 sdist/
 var/
+venv/
 *.egg-info/
 .installed.cfg
 *.egg
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -4,6 +4,12 @@
 
 ## [Unreleased]
 
+#### Breaking changes
+
+- Now `c.KubeSpawner.environment` values supports substitution, just like other config options [#642](https://github.com/jupyterhub/kubespawner/pull/642) ([@dolfinus](https://github.com/dolfinus)).
+  For example, `{"MYVAR": "jupyterhub-{username}"}` will rendered as `{"MYVAR": "jupyterhub-sam"}` for a user named `sam`.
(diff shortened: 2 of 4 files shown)


yuvipanda:
I'm trying to understand if this would be a 'breaking' change, as users might already have things like `${SHELL}` in their env vars. Would we produce a syntax error if that were the case? I know they don't expand in kubernetes unless you use `$(SHELL)` as syntax and SHELL is explicitly set in the kubernetes spec, but what happens when these vars are set for `${SHELL}`? 

I think this is the right thing to do and we should land this PR - I just want to understand if this is a breaking change or not so we can label it appropriately.

Thanks a lot for working on this!

dolfinus:
I've changed current implementation to avoid breaking backward compatibility - instead of `.format` `_expand_all` uses `.replace`. For unknown substitutions, spawner will show a warning

yuvipanda:
@dolfinus ah, damn - I think this might break current escaping where you can escape variables with something like `{{username}}` - that renders as just `{username}` instead of being expanded. With this change, those *will* get replaced. So it's gonna break in a different way.

I think the right thing to do is to mark this as a breaking change, and setup `_expand_all` in basically as many places that need it in one go :) What do you think?

dolfinus:
Ok, reverted latest change back

manics (approved):
Looks great!

dolfinus:
Should I update CHANGELOG, or this will be done after merging?

manics:
Would you mind adding a paragraph to the Changelog now, ideally including brief upgrade/compatibility instructions? It'll help whoever does the next release.

dolfinus:
Ok, please check

dolfinus:
Any updates?

yuvipanda:
Thanks, @dolfinus!

Issue #523, opened 2021-08-06
Support to create the networkpolicy dynamically 


ripplehang:
### Proposed change

When spawn the new pod for the jupyter, I also want to configure the networkpolicy for the new spawned pod

### Alternative options

Currently the networkpolicy is static and doesn't exists the API to configure the networkpolicy dynamically.

### Who would use this feature?

the networkpolicy should be be important secure mechanism, since the python scripts run in jupyter notebook would be free style, i want to limit the impact by only allow the pod to access the specific range of the destnation.

dolfinus:
#644 can solve this

PR #644, opened 2022-10-12
Add after_pod_created_hook

dolfinus:
KubeSpawner has `modify_pod_hook` which allows to change pod spec before it will be launched. But there is not hook to access pod after it was created but before spawner's `start` method is already finished.

This hook is useful for creating other K8S objects, like services or ingress, with properly set up `owner_reference` which allows to automatically remove all created objects after pod removal.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -1022,6 +1022,26 @@ class KubeSpawner(Spawner):
         """,
     )
 
+    after_pod_created_hook = Callable(
+        None,
+        allow_none=True,
+        config=True,
+        help="""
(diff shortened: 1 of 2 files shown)


consideRatio:
@dolfinus are there issues that can be closed/resolved by getting this feature merged?

dolfinus:
It may fix #523, but I'm not sure

yuvipanda:
Thanks, @dolfinus!

sbernauer:
Thanks @dolfinus for implementing this! Helped us quite a lot setting up Services for distributed pyspark code

PR #645, opened 2022-10-13
Use jupyterhub.utils.maybe_future instead of tornado.get.maybe_future

### LINKED ISSUE #490: c.Kubespawner.modify_pod_hook does not support co-routines
### Bug description

I am facing an issue when providing a co-routine to `c.Kubespawner.modify_pod_hook`.

The Tornado docs for `gen.maybe_future()` say that `maybe_future` can only handle Futures and no other yieldable objects.

Does this mean, we can pass coros as a hook function?
I mean a coro which as been declared via `async def ...`

#### Expected behaviour

`c.Kubespawner.modify_pod_hook` should support co-routines.

#### Actual behaviour

`c.Kubespawner.modify_pod_hook` only supports normal functions or futures.

If passed a co-routine, I will pass the co-routine into the Kubernetes client which fails.

### How to reproduce

1. Define a co-routine via the `async def `syntax:

```python
import asyncio

async def my_pod_modify_hook(spawner):
     await asyncio.sleep(1)
     return

c.Kubespawner.modify_pod_hook = my_pod_modify_hook
```

### Your personal set up

A pretty standard Z2JH setup (master)

dolfinus:
`tornado.get.maybe_future` handles only functions decorated by `@gen.coroutine`, but not `async def` syntax added in Python 3.5. Using `jupyterhub.utils.maybe_future` instead fixes this issue.

Fixes #490.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     install_requires=[
         'escapism',
         'python-slugify',
-        'jupyterhub>=0.9',
+        'jupyterhub>=0.9.1',
         'jinja2',
         'kubernetes_asyncio>=19.15.1',
         'urllib3',
--- a/setup.py
+++ b/setup.py
@@ -11,10 +11,12 @@ if v[:2] < (3, 7):
 setup(
     name='jupyterhub-kubespawner',
     version='4.2.1.dev',
+    # NOTE: If lower bounds are updated, also update our test for the lower
+    #       bounds in .github/workflows/test.yaml.
     install_requires=[
         'escapism',
         'python-slugify',
-        'jupyterhub>=0.9.1',
+        'jupyterhub>=1.3.0',
         'jinja2',
         'kubernetes_asyncio>=19.15.1',
         'urllib3',
@@ -27,7 +29,7 @@ setup(
             'kubernetes>=11',
             'pytest>=5.4',
             'pytest-cov',
-            'pytest-asyncio>=0.11.0',
+            'pytest-asyncio>=0.17',
         ]
     },
     description='JupyterHub Spawner for Kubernetes',
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -41,10 +41,12 @@ jobs:
         # kubernetes_asyncio:  https://github.com/tomplus/kubernetes_asyncio/tags
         #
         include:
-          # Tests with oldest supported Python, k8s, and k8s client
+          # Tests with oldest supported Python, jupyterhub, k8s, and k8s client
           - python: "3.7"
             k3s: v1.20
-            test_dependencies: kubernetes_asyncio==19.*
+            test_dependencies: >-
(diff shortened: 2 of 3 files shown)


consideRatio (approved):
LGTM!

The approach seems fine and validated by @minrk in https://github.com/jupyterhub/kubespawner/issues/490#issuecomment-791313061. With this PR, there is no longer a trace of tornado left in kubespawner.

Remaining question, should we bump JupyterHub's lower version bound?

https://github.com/jupyterhub/kubespawner/blob/fa19acdbb056c8d105baab13bbaa2b8b59dfa962/setup.py#L14-L22

The `maybe_future` function has not changed since 1.3.0, but was changed in 1.3.0. See https://github.com/jupyterhub/jupyterhub/blame/5997614f452a3cfebfa541800012a914d37f3ec8/jupyterhub/utils.py#L572-L596 and https://github.com/jupyterhub/jupyterhub/pull/3242.

@minrk do you agree that we should bump to version 1.3.0 as a lower bound?

dolfinus:
Bumped Jupyterhub version to `0.9.1`, `maybe_awaitable` does not exist in `0.9.0`

consideRatio:
@dolfinus I'm considering if we need to bump to 1.3.0 as well, which I think would be perfectly fine to do. What do you think?

> The `maybe_future` function has not changed since 1.3.0, but was changed in 1.3.0. See [git blame](https://github.com/jupyterhub/jupyterhub/blame/5997614f452a3cfebfa541800012a914d37f3ec8/jupyterhub/utils.py#L572-L596) showing [this most recent change for 1.3.0](https://github.com/jupyterhub/jupyterhub/commit/3c071467bb97647661e775f8ccae3da6f6fafd91) as part of [this PR](https://github.com/jupyterhub/jupyterhub/pull/3242).
> 
> @minrk do you agree that we should bump to version 1.3.0 as a lower bound?

setup.py
```diff
@@ -14,7 +14,7 @@
     install_requires=[
         'escapism',
         'python-slugify',
-        'jupyterhub>=0.9',
+        'jupyterhub>=0.9.1',
```
consideRatio, line 17: Not sure this is what we want, but think it might be. See discussion in https://github.com/jupyterhub/kubespawner/pull/645#pullrequestreview-1140452434

```suggestion
        'jupyterhub>=1.3.0',
```

dolfinus, line 17: Isn't that conflict with jupyterhub version from `test_dependencies` of CI/CD workflow?

consideRatio: Hmmm I'm not sure what you mean.

consideRatio:
@dolfinus and @minrk I've verified by running tests that KubeSpawner didn't support use with JupyterHub <1.3.0 even before this PR.

consideRatio:
Thank you @dolfinus!!!! :heart: :tada: :sunflower:

Issue #638, opened 2022-09-24
KubeIngressProxy.add_route does not catch 404 exception


ealyn:
### Bug description
When I upgrade from 1.1.0 to 4.2.0 for using networking.k8s.io/v1 ingress, the jupyterhub can not startup successfully.
It seems that the reason is the k8s client switch from kubernetes to kubernetes-asyncio, but exception catching not follow, it still catch kubernetes client.rest.ApiException.

#### Expected behaviour

In old version, `delete_if_exists` works find

> [I 2022-09-24 07:03:28.010 JupyterHub proxy:347] Checking routes
> [I 2022-09-24 07:03:28.010 JupyterHub proxy:432] Adding route for Hub: /jupyterhub/ => http://jupyterhub-hub.jupyterhub-test:8081
> [W 2022-09-24 07:03:28.022 JupyterHub proxy:174] Could not delete endpoint/jupyter--2fjupyterhub-2f-route: does not exist

#### Actual behaviour

The error come up while jupyterhub startup adding route:

> [I 2022-09-24 06:57:45.072 JupyterHub proxy:347] Checking routes
> [I 2022-09-24 06:57:45.073 JupyterHub proxy:432] Adding route for Hub: /jupyterhub/ => http://jupyterhub-hub.jupyterhub-test:8081
> [I 2022-09-24 06:57:45.073 JupyterHub reflector:274] watching for ingresses with label selector='component=singleuser-server,hub.jupyter.org/proxy-route=true' in namespace jupyterhub-test
> [E 2022-09-24 06:57:45.088 JupyterHub app:2989]
>     Traceback (most recent call last):
>       File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 2987, in launch_instance_async
>         await self.start()
>       File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 2852, in start
>         await self.proxy.check_routes(self.users, self._service_map)
>       File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 70, in locked_method
>         return await method(*args, **kwargs)
>       File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 426, in check_routes
>         await asyncio.gather(*futures)
>       File "/usr/local/lib/python3.8/dist-packages/kubespawner/proxy.py", line 282, in add_route
>         delete_endpoint = await self.core_api.delete_namespaced_endpoints(
[...]

PR #646, opened 2022-10-13
[KubeIngressProxy] Fix 404 error in `add_route`

dolfinus:
Fix for #638

### DIFF
--- a/kubespawner/proxy.py
+++ b/kubespawner/proxy.py
@@ -279,7 +279,7 @@ class KubeIngressProxy(Proxy):
                 'Could not find endpoints/%s after creating it' % safe_name,
             )
         else:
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Any changes here are not scrutinized as well as changes in KubeSpawner because we lack end to end tests.

Is this tested to work manually?

dolfinus:
Yes, I've triesdto use KubeIngressProxy and got exception while hub startup, which has gone after this fix

ealyn:
@dolfinus I've view the PR and wonder if function delete_route need remove `await`?

ealyn:
> Why?

Oh, I found you have commit the fix #649. Now everything is find.

Issue #390, opened 2020-01-26
 KubeSpawner override for profile overrides existing spawner.environment


scivm:
I have a pre_spawn_start where I get the AD_TOKEN and put it to the spawner.environment dictionary.   My single user profile sets environment variables also.  kubespawner is using these variables to override the spawner environment which causes my single user notebook to fail because it requires a AD_TOKEN (own requirement).  I think the profile environment dictionary should append to the existing spawner environment variables if they are not the same?  

Also, if you try a second spawn just after the first failed one then both the AD_TOKEN and the override variables are set in spawner.environment and spawn works fine.

Single user override in z2h config.yaml:
   ```
- display_name: "Datascience environment"
      description: "Python, R, and Julia."
      default: true
      kubespawner_override:
        image: scivm/jupyter-datascience:latest
        cpu_guarantee: 1
        cpu_limit: 2
        mem_guarantee: "2096M"
        mem_limit: "8192M"
        environment:
             datasource_driverClassName: org.postgresql.Driver
```

It can be seen that in pre_spawn_start the AD_TOKEN is set, but then in modify_pod_hook, only the postgres environment variable is still set.
```
printing spawner.environment from pre_spawn_start
{'AD_TOKEN': 'eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsI...'

[D 2020-01-26 18:01:03.351 JupyterHub user:556] Calling Spawner.start for shamberger michael (ext)
[D 2020-01-26 18:01:03.351 JupyterHub spawner:1988] Applying KubeSpawner override for profile 'Datascience environment'
[D 2020-01-26 18:01:03.351 JupyterHub spawner:1995] .. overriding KubeSpawner value cpu_guarantee=1
[D 2020-01-26 18:01:03.351 JupyterHub spawner:1995] .. overriding KubeSpawner value cpu_limit=2
[D 2020-01-26 18:01:03.351 JupyterHub spawner:1995] .. overriding KubeSpawner value environment={'datasource_driverClassName': 'org.postgresql.Driver'}
[D 2020-01-26 18:01:03.352 JupyterHub spawner:1995] .. overriding KubeSpawner value image=scivm/jupyter-datascience:latest
[D 2020-01-26 18:01:03.352 JupyterHub spawner:1995] .. overriding KubeSpawner value mem_guarantee=2096M
[...]

PR #650, opened 2022-10-14
Let `kubespawner_override` merge instead of replace dictionaries

yuvipanda:
Currently, if you set any dictionary values in kubespawner_override, it completely *replaces* the existing dictionary. Rest of z2jh for the most part - particularly helm - *merge* dictionaries rather than replace it. This brings kubespawner in line with that.

Ref https://github.com/2i2c-org/infrastructure/issues/1530


consideRatio:
I think there is a real value to both be able to merge dictionaries, and to replace them. So switching one for the other doesn't seem like a pure win, making this change a trade off where I think conservatively. I would have preferred that the initial behavior to be like the helm charts.

Is there an API change or addition that could accomodate both needs without being breaking, or with a breaking change that end users easily can adjust to?

## Related

- https://github.com/jupyterhub/kubespawner/issues/390
- https://github.com/jupyterhub/kubespawner/pull/393
- Known workaround to a need to merge dictionaries: https://github.com/jupyterhub/kubespawner/pull/393#issuecomment-703309234

yuvipanda:
Thank you for linking to prior work, @consideRatio! I spent some time thinking about what configuration would make sense to replace a dictionary rather than merge, and I actually can’t think of any at all! Do you have examples you can share? I can think of places where you want to explicitly unset an item but not replace the entire dictionary. I agree that it would have been great if the original behavior matched helm, but I think we can still change this if there isn’t a strong set of examples where the existing replacement behavior is required.

yuvipanda:
I understand your concerns around breaking and complexity, @consideRatio. However, I think the current behavior is *purely* a bug, and folks don't actually notice until quite late - https://github.com/2i2c-org/infrastructure/issues/1775 and https://github.com/2i2c-org/infrastructure/issues/1774 was reported to us close to a year after they had users, and I don't know how long it took for you to notice https://github.com/2i2c-org/infrastructure/issues/1530! So it's a breaking change, but I think it's the right thing to do so it behaves like how rest of config behaves.

However, a lot of this is derived from the fact that I can't think of a single config property inside kubespawner where you want to *replace* the entire dictionary rather than merge. I do think there are cases where you want to *delete* an existing variable - if you want, I can find a way to add that functionality.

consideRatio:
I don't want to block this, but it would be nice to have another person agreeing on this breaking change.

> I do think there are cases where you want to delete an existing variable - if you want, I can find a way to add that functionality.

I think we should rely on the existing hooks for customizations beyond what is supported with what `kubespawner_override` does to reduce complexity growth. I figure appending or updating lists etc is a similar feature that I think we can't provide a dedicated API for.

consideRatio:
This is a breaking change, but its also seen as a bugfix. Should we go for a major version bump in KubeSpawner, and if we do, should we also go for a major version bump of z2jh when adopting this version of KubeSpawner?

@yuvipanda could you add a changelog entry about this to keep the release process lightweight?

consideRatio (approved):

minrk:
Given that this is really ~only used in z2jh, matching helm's dict merge behavior makes sense to me. As long as there's a way to 'clear' values (usually setting to None or an empty string is equivalent to clearing), I don't think there's a downside of merge behavior.

But I see that this doesn't appear to _quite_ do that - it only goes one level deep with `dict.update`. Do we have any dict config where _values_ can themselves be dicts (e.g. full object manifests)? If so, you'll need to do a _deep_ merge (i.e. recursively iterate through dicts, merging each _value_ if they are dicts as well). We have an implementation [in jupyter server](https://github.com/jupyter-server/jupyter_server/blob/fc0ac3236fdd92778ea765db6e8982212c8389ee/jupyter_server/config_manager.py#L14). Note that in that case, setting `None` is explicitly used to clear values. This may not always be correct in KubeSpawner.

clobber:

- pro: override config is 'what you get'
- pro: coming from a Jupyter config perspective, 'replace the config' is a logical expectation because it's the same behavior as `c.KubeSpawner.key = override_value`
- con: only way to merge is to duplicate what was there before (not always possible, always tedious)

merge:

- pro: simpler to override just what you need
- pro: coming from helm perspective (most users, I suspect), merge behavior is consistent with other dict config handling
- con: "just use this" clobber requires knowing what was there before (not always possible, very rare case, I suspect)

manics:
I don't have a strong feeling on merge vs overwrite.

> But I see that this doesn't appear to quite do that - it only goes one level deep with dict.update. Do we have any dict config where values can themselves be dicts (e.g. full object manifests)? If so, you'll need to do a deep merge (i.e. recursively iterate through dicts, merging each value if they are dicts as well).

I think we should be consistent though- if we go for merge then do a deep merge. Some config values represent K8s objects, I haven't checked whether they're nested Dicts (so far I've found a nested array in `SecurityContext`), but adding the deep merge now will avoid future unexpected behaviour if KubeSpawner is extended in future.

consideRatio:
:+1: on deep merge I think.

### Overview of influenced `Dict()` config

These are KubeSpawner's top level configurable `Dict()` traitlets, which are the only options possibly influenced I think.

```
common_labels
environment
extra_annotations
extra_container_config
extra_labels
extra_pod_config
extra_resource_guarantees
extra_resource_limits
lifecycle_hooks
node_selector
storage_extra_annotations
storage_extra_labels
storage_selector
user_namespace_annotations
user_namespace_labels
```

### Example situation where its a breaking change deployment

For reference, I have an example situation from me as an JupyterHub admin where an issue arise. I'll try adjust to this but want to highlight as an example of how it can break deployments for other users and that this can be practically breaking for people.

When I deploy a jupyterhub helm chart via 2i2c's helm chart that depends on the jupyterhub helm chart, it [comes with a predefined `singleuser.nodeSelector` configuration](https://github.com/2i2c-org/infrastructure/blob/001a0855118632ac10a93f3c1924fdf926d5318c/hub-templates/base-hub/values.yaml#L90-L91). I have currently [relied on `kubespawner_override`](https://github.com/2i2c-org/infrastructure/pull/436/files#diff-91aa1f4385c492c4adba448b741048dfa33db48b2468807638d7627fd97b7cd3R131-R140) to override the meta chart's provided `singleuser.nodeSelector` (`KubeSpawner.node_labels`) label that I [haven't set on my k8s nodes](https://github.com/2i2c-org/infrastructure/pull/436/files#diff-2913cb40ae1ad80f097b0f052b51771f17cde8d998b99b6bb924e8e0b18bf36fR185-R187).

I think I can adjust by defining this in my helm chart configuration file.

```yaml
# basehub is a 2i2c meta helm chart depending on the jupyterhub helm chart
basehub:
  jupyterhub:
    singleuser:
      nodeSelector: null
```

yuvipanda:
Ok, so how about we steal the implementation in https://github.com/jupyter-server/jupyter_server/blob/fc0ac3236fdd92778ea765db6e8982212c8389ee/jupyter_server/config_manager.py#L14 (it's small enough) and use it here. That sounds good?

consideRatio:
## Breaking change (proposal for changelog text)

A breaking change _could occur_ if `kubespawner_override` is used to override a dictionary based configuration listed below. It will only be a breaking change if there is a difference between merging and replacing previous dictionary though.

There could be a difference between merging and replacing if for example `common_labels` is configured ahead of time to be `{"a": "a"}`, and then `kubespawner_override` is used to set it to `{"b": "b"}`. Then previosly the result would be `{"b": "b"}` while now it will be `{"a": "a", "b": "b"}`.

Only `common_labels` has a default value provided by KubeSpawner or the JupyterHub Helm chart, making this mainly an issue for users that first have configured one of these values and then expect it to be entirely replaced in `kubespawner_override`. As an example, someone may have configured `node_selector` to be `{"a": "a"}` by default and then expect it to become `{"b": "b"}` after using `kubespawner_override`, but after this change it will end up being `{"a": "a", "b": "b"}`.

```
common_labels
environment
extra_annotations
extra_container_config
extra_labels
extra_pod_config
extra_resource_guarantees
extra_resource_limits
lifecycle_hooks
node_selector
storage_extra_annotations
storage_extra_labels
storage_selector
user_namespace_annotations
user_namespace_labels
```

yuvipanda:
Is there someone who would be willing to take this over from me? If not I can try to find time this / next week...

consideRatio:
I'm overall not at all confident about this bugfix due to its breaking in a way that many have started to rely on. I think working this to the discussed state when deep merges are made, and then merging this PR, will likely be followed by significant maintenance needs in this repo due to the breaking changes.

I believe people will have issues directly due to the breaking change, and that a few won't have a good workaround possible. Due to this, I'd like to not go for this.

manics:
I don't have any view on whether this feature is needed or not. If we do want to keep this one option is a new `kubespawner_merge` property that's applied after `kubespawner_override`.

yuvipanda:
It would make me pretty sad to not merge this *purely* for backwards compatibility reasons, as it makes using environment variables in kubespawner_overrides hard otherwise. I am also not sure that significant changes will be needed - this only affects people who are using overrides of things that are *dictionaries* in `kubespawner_override` *multiple .yaml files* being passed to `helm`. IMO that's not the huge majority of people - if you only have a single values.yaml, you will actually see no difference at all.

However, I also currently have no energy and am going on vacation in a few days (hopefully i will have energy when i am back in two weeks), and can implement https://github.com/jupyterhub/kubespawner/pull/650#issuecomment-1281815803.

consideRatio:
@yuvipanda I'm sorry for second guessing support for this PR, I've gone back and forth on this. Its related to my struggle to overview the consequences of this. If you get this to do deep merges, I'll work on documenting the breaking change thoroughly!

> if you only have a single values.yaml, you will actually see no difference at all

I think these are examples of a single chart config file where you will experience a breaking change. Let's go for it anyhow though!

```yaml
# breaking config
singleuser:
  nodeSelector:
    key_a: a
  profileList:
    - display_name: something
      kubespawner_override:
        node_selector:
          key_b: b
```

```yaml
# breaking config assuming the chart installed depends on the jupyterhub chart, 
# and that chart has provided some default value for jupyterhub.singleuser.nodeSelector
#
jupyterhub:
  profileList:
    - display_name: something
      kubespawner_override:
        node_selector:
          key_b: b
```

yuvipanda:
aaah, sorry I missed that @consideRatio. I still think it'll be a fairly small number of people, and it will be a positive change for most of them avoiding duplication. I appreciate the deep emphasis you place on documenting breaking changes, and on the empathy you have for operators.

I'll pick this back up once I'm back from vacation. I don't think this needs to block any releases - we can make another one when it's done if you want to make a kubespawner release before. However, if someone else wants to pick this up and drive it to completion, I'd be perfectly happy for them to do so!

yuvipanda:
ok, I've updated the code now and added a bit of docs to `kubespawner_override` at the appropriate place I think.

yuvipanda:
ok, I'll look at the test failures shortly.

kubespawner/utils.py
```diff
@@ -202,3 +202,26 @@ def host_matching(host: str, wildcard: str) -> bool:
...
+
+    None values will delete their keys.
+    """
+    for k, v in new.items():
+        if isinstance(v, dict):
+            if k not in target:
+                target[k] = {}
+            recursive_update(target[k], v)
+            if not target[k]:
+                # Prune empty subdicts
```
manics, line 220: Is this definitely the right behaviour? Does K8s consistently treat an absent key as `{}` as opposed to null/None?

yuvipanda: @manics I've simplified and removed that behavior.

yuvipanda:
ok I've fixed the failing tests!

consideRatio:
> In other repos we've been adding breaking changes to the changelog straightaway to help with whoever does the release, e.g.:
> https://github.com/jupyterhub/kubespawner/pull/642/files#diff-48cd8b08129999cc6aaf4078c85eb99efd0cf0ed88331178f1a784781e9ee4cb
> Do you mind adding something, then I think this is good to merge.

@yuvipanda and @manics I wrote https://github.com/jupyterhub/kubespawner/pull/650#issuecomment-1304568831 about this earlier - @manics / @yuvipanda what do you think? If acceptale, it can be copy pasted to the changelog under the unreleased section.

---

> A breaking change _could occur_ if `kubespawner_override` is used to override a dictionary based configuration listed below. It will only be a breaking change if there is a difference between merging and replacing previous dictionary though.
> 
> There could be a difference between merging and replacing if for example `common_labels` is configured ahead of time to be `{"a": "a"}`, and then `kubespawner_override` is used to set it to `{"b": "b"}`. Then previosly the result would be `{"b": "b"}` while now it will be `{"a": "a", "b": "b"}`.
> 
> Only `common_labels` has a default value provided by KubeSpawner or the JupyterHub Helm chart, making this mainly an issue for users that first have configured one of these values and then expect it to be entirely replaced in `kubespawner_override`. As an example, someone may have configured `node_selector` to be `{"a": "a"}` by default and then expect it to become `{"b": "b"}` after using `kubespawner_override`, but after this change it will end up being `{"a": "a", "b": "b"}`.
> 
> ```
> common_labels
> environment
> extra_annotations
> extra_container_config
> extra_labels
> extra_pod_config
> extra_resource_guarantees
> extra_resource_limits
> lifecycle_hooks
> node_selector
> storage_extra_annotations
> storage_extra_labels
> storage_selector
> user_namespace_annotations
> user_namespace_labels
> ```

yuvipanda:
Thanks for helping get it through, @manics and @consideRatio!

PR #648, opened 2022-10-13
[KubeIngressProxy] Add `ingress_class_name`, `[hub|service|server]_specification`, `common_labels`, `extra_[labels|annotations]`

dolfinus:
Alternative implementation of #406.

1. Different config options for `/hub` services (`/service/:name`), and server routes (`/user/:user/:server`), which allows to set up `hub.domain.com`, `{service}.hub.domain.com`  `{user}.hub.domain.com` hosts within the same config.
2. `host` and `tlsSecret` can be omitted to avoid breaking backward compatibility
3. Allow to pass custom `ingressClassName`, as well as set additional labels and annotations for created Endpoints, Service and Ingress (e.g. `nginx.ingress.kubernetes.io/ssl-redirect: "true"`)
4. Allow new options to contain substitutions just like KubeSpawner

It can be too much for one pull request, because of that it's in Draft

PR #653, opened 2022-10-21
[KubeIngressProxy] Add common_labels option and expand username etc

dolfinus:
This allows to add custom labels to all objects created by KubeIngressProxy, like Endpoints, Service and Ingress. For example, admin can add here helm release name.

First step of dividing #648 into smaller pieces

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -6,6 +6,7 @@ import ipaddress
 import json
 import operator
 import re
+from typing import Optional
 from urllib.parse import urlparse
 
 from kubernetes_asyncio.client.models import (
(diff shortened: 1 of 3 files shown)


docs/source/changelog.md
```diff
@@ -4,6 +4,8 @@
 
 ## [Unreleased]
 
+- Add `common_labels` option to KubeIngressProxy [#653](https://github.com/jupyterhub/kubespawner/pull/653) ([@dolfinus](https://github.com/dolfinus))
+
```
consideRatio, line 8:
```suggestion
```

We can easily generate these using the tool `github-activity`, it is only the breaking changes summarized that its critical to add manually to the changelog ahead of time.

So, this is better removed so that the tool can do it automatically, and use the labels of the PR to categorize it under a category etc.

dolfinus: Ok, removed this

consideRatio (approved):
LGTM, thanks for your thorough work on this @dolfinus!!

PR #654, opened 2022-10-22
Fix hard-coded component label for services_enabled=True

dolfinus:
I've set up KubeSpawner with following settings:
```python
c.KubeSpawner.component_label = "mylabel"
c.KubeSpawner.services_enabled = True
```

But created services were not bound to the notebook pod because 'component' label is hard-coded in `make_service` function, and always have `singleuser-server` value.

Fixed that by moving selector build outside the `make_service` function.
Added tests for custom `component_label` value passed with different values of `services_enabled`.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -55,6 +55,11 @@ class PodReflector(ResourceReflector):
 
     kind = "pods"
 
+    # The default component label can be over-ridden by specifying the component_label property
+    labels = {
+        'component': 'singleuser-server',
+    }
+
     @property
     def pods(self):
         """
@@ -113,25 +118,24 @@ class KubeSpawner(Spawner):
     spawned by a user will have its own KubeSpawner instance.
     """
 
-    # Reflectors are stored in class variable for performance reasons.
-    # If every pod will start its own reflector, Kubelet will not be happy
-    # with attaching 20k event watchers
-    reflectors = {}
+    # Reflectors keeping track of the k8s api-server's state for various k8s
+    # resources are singletons as that state can be tracked and shared by all
+    # KubeSpawner objects.
+    reflectors = {
+        "pods": None,
+        "events": None,
+    }
 
     # Characters as defined by safe for DNS
     # Note: '-' is not in safe_chars, as it is being used as escape character
     safe_chars = set(string.ascii_lowercase + string.digits)
-
-    def get_reflector_key(self, kind: str) -> tuple:
-        namespace = None if self.enable_user_namespaces else self.namespace
-        return (kind, self.component_label, namespace)
 
     @property
     def pod_reflector(self):
         """
         A convenience alias to the class variable reflectors['pods'].
         """
-        return self.__class__.reflectors.get(self.get_reflector_key('pods'))
+        return self.__class__.reflectors['pods']
 
     @property
     def event_reflector(self):
@@ -140,9 +144,7 @@ class KubeSpawner(Spawner):
         spawner instance has events_enabled.
         """
         if self.events_enabled:
-            return self.__class__.reflectors.get(self.get_reflector_key('events'))
-
-        return None
+            return self.__class__.reflectors['events']
 
     def __init__(self, *args, **kwargs):
         _mock = kwargs.pop('_mock', False)
@@ -2378,9 +2380,9 @@ class KubeSpawner(Spawner):
 
     def _start_reflector(
         self,
-        kind: str,
-        reflector_class: ResourceReflector,
-        replace: bool = False,
+        kind=None,
+        reflector_class=ResourceReflector,
+        replace=False,
         **kwargs,
     ):
         """Start a shared reflector on the KubeSpawner class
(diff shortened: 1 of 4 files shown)


dolfinus:
Is it OK that Reflector is saved into class attribute, but it is bound to a specific KubeSpawner instance?
https://github.com/jupyterhub/kubespawner/blob/b96c54f191aa2b475a4cf64af8405517167062db/kubespawner/spawner.py#L2410-L2415

kubespawner/spawner.py
```diff
@@ -2407,7 +2408,7 @@ def on_reflector_failure():
         previous_reflector = self.__class__.reflectors.get(key)
 
         if replace or not previous_reflector:
-            self.__class__.reflectors[key] = ReflectorClass(
+            self.__class__.reflectors[key] = reflector_class(
```
dolfinus, line 2411: Why reflector is stored in the class attribute? Is it caused by some performance issues with Kubernetes API?

dolfinus: Ok, added a comment

kubespawner/spawner.py
```diff
@@ -118,21 +113,25 @@ class KubeSpawner(Spawner):
...
+    # Reflectors are stored in class variable for performance reasons.
+    # If every pod will start its own reflector, Kubelet will not be happy
+    # with attaching 20k event watchers
+    reflectors = {}
 
     # Characters as defined by safe for DNS
     # Note: '-' is not in safe_chars, as it is being used as escape character
     safe_chars = set(string.ascii_lowercase + string.digits)
 
+    def get_reflector_key(self, kind: str) -> tuple:
```
dolfinus, line 125: I'm not sure about this implementation.

From the one hand, tests are working properly without calling any additional method to reset reflectors. From another, changes needed by tests are added into spawner class

consideRatio: Can you clarify where and why tests needed this change?

dolfinus: I've added test `test_spawn_component_label` which starts pod with different `component_label` value. But because reflectors use `component_label` to get events from k8s, and reflector instance is just one per KubeSpawner class, previous tests already set different `component_label`, which causes failure of `test_spawn_component_label`.

There are 3 ways to solve this:
1. Create reflector per spawner object instead of the whole class. Cannot be done because this reduces performance
2. Reset reflector object before and after test with a different `component_label`. I've tried to do this previously, but calling private class method in tests is quite strange for me.
3. Create reflector instance for each component_label (plus namespace, if reflector is watching for a specific namespace)

Now I implemented option 3, is that OK?

consideRatio: I think overall KubeSpawner's reflectors are the hardest thing to understand and maintain, so adding complexity in there to support a test makes me uneasy.

I wonder if updating tests to avoid re-using old configuration of reflectors or similar is possible. I'd like to push a bit more on strategy 2 before resorting to strategy 3, I'll see if I can figure something out.

From now on to ease my review process, do not make any force pushes but instead add commits if you make changes.

consideRatio: For reference, this is where I'm looking: https://github.com/jupyterhub/kubespawner/pull/654/files#diff-84f25ce0bea68e4626c0cc2f38be2ec945648eddce266b2bbac73519390756f9R209-R222

How and can we get the KubeSpawner class state be reset between these tests?

dolfinus: This was in previous implementation I've pushed to this PR:
https://github.com/jupyterhub/kubespawner/commit/89e645c9ddfaf1f268dea0b11d375fd7b6aa4015#diff-6fdb3a928956e8ab2b810679d51fae72870d45ed4a088e85e16814339aff8901R2429
https://github.com/jupyterhub/kubespawner/commit/89e645c9ddfaf1f268dea0b11d375fd7b6aa4015#diff-84f25ce0bea68e4626c0cc2f38be2ec945648eddce266b2bbac73519390756f9R225-R234

consideRatio: Thanks!

Okay I've pushed a commit to this PR, reverting a few things related to strategy 3 above in favor of my approach of strategy 2. What do you think?

kubespawner/spawner.py
```diff
@@ -2457,10 +2460,11 @@ def _start_watching_pods(self, replace=False):
         If replace=True, a running pod reflector will be stopped
         and a new one started (for recovering from possible errors).
         """
+        pod_reflector_class = PodReflector
+        pod_reflector_class.labels.update({"component": self.component_label})
```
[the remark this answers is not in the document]
consideRatio: > IMHO updating class variable is not necessary here and should be avoided

Hmmm you may be right, I made it like this to minimize the changes that seemed unrelated. Do you consider it related to this PR atm?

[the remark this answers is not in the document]
dolfinus: OK, I'll create separated PR for this

consideRatio:
@dolfinus if this looks good to you, I suggest we go for a merge!

Super thankful for your thorough work and wants to do what I can manage to help you retain a momentum!!

consideRatio (approved):

kubespawner/spawner.py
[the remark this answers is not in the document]
dolfinus: #672

consideRatio:
image

Rocket approval :)

PR #656, opened 2022-10-22
[KubeIngressProxy] Reuse service created by KubeSpawner.services_enabled

dolfinus:
If `KubeSpawner.services_enabled` option is set to `True`, `KubeIngressProxy` will get target URL in a form `http://jupyterhub-user--servername.pod-namespace.svc.cluster.local:8888`, but instead of reusing the existing service it creates ExternalName pointing to `jupyterhub-user--servername.pod-namespace.svc.cluster.local` DNS name.

Sometimes it is useful, for example if pod is created in another namespace, and there is no way to create Ingress with rule pointing to another namespace.
But in other cases `ExternalName` is redundant, and may cause errors. In my case, ingress paths with ExternalName are quire unstable, 50% of runs end with an error "500 Redirect loop detected".

Here I've added a special case for `make_ingress` - if `target` is a valid k8s service name, and it has the same namespace as ingress, Ingress -> External -> Service is reduced to Ingress -> Service, and Service is reused instead of creating new one.


kubespawner/objects.py
```diff
@@ -61,6 +61,10 @@
 
 from .utils import get_k8s_model, update_k8s_model
 
+SERVICE_DNS_PATTERN = re.compile(
```
consideRatio, line 64: An explanation here, or a reference to where its explained, seems very relevant. This influences KubeSpawner code as well and I care deeply that this project is possible to maintain long term - having comments like that can be essential.

consideRatio:
> This behavior is activated by option KubeIngressProxy.reuse_existing_services=True, which is disabled by default to avoid backward compatibility issues.

Next release is a major version of KubeSpawner, and I think you have worked KubeIngressProxy significantly so that it could be fine to include a breaking change. Is this behavior the better default in your mind? If so, maybe we should switch to it.

dolfinus:
Well, I'm not sure it this issue with accessing Service from ExternalName is related to my k8s instance or it is a common behavior. It's better to wait for other users to use this option before changing the defaults.

consideRatio (approved):

PR #657, opened 2022-10-24
Save the namespace between restarts

totycro:
Hi,

I'm using a setup of kubespawner, where the `namespace` depends on the jupyter profile which the users selects. So users can start sessions in different namespaces where they will have access to different resources on our cluster. This is implemented by setting `namespace` in `kubespawner_override`.

So far this works great, the only issue we've found is that when the hub is restarted, it doesn't find the pod any more because it looks for the pod in the default namespace.

This PR will serialize the namespace, so when the spawner is loaded, the `namespace` is set up correctly. For my use case, this is necessary, and it also shouldn't cause problems for other use cases, so is this something you could consider merging?

On an unrelated note, we are also customizing the `pod_name` for different jupyter profiles, so we actually depend on `pod_name` being serialized. This is already the case, but the documentation in the code says that this isn't really necessary.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2188,9 +2188,13 @@ class KubeSpawner(Spawner):
 
         It's also useful for cases when the `pod_template` changes between
         restarts - this keeps the old pods around.
+
+        We also save the namespace for use cases where the namespace is
+        calculated dynamically.
         """
         state = super().get_state()
         state['pod_name'] = self.pod_name
+        state['namespace'] = self.namespace
         return state
 
     def get_env(self):
@@ -2215,9 +2219,14 @@ class KubeSpawner(Spawner):
         but if the `pod_template` has changed in between restarts, it will no longer
         be the case. This allows us to continue serving from the old pods with
         the old names.
+
+        For a similar reason, we also save the namespace.
         """
         if 'pod_name' in state:
             self.pod_name = state['pod_name']
+
+        if 'namespace' in state:
+            self.namespace = state['namespace']
 
     @_await_pod_reflector
     async def poll(self):


yuvipanda:
Thanks @totycro! Yeah, the pod_name is serialized because the template used to create pod names could be changed between restarts. I think the same is true here for namespaces in your config, and so it makes sense to add it here. Where is the documentation specifying that it isn't necessary? We should make that more vague :)

kubespawner/spawner.py
```diff
@@ -2215,10 +2219,15 @@ def load_state(self, state):
         but if the `pod_template` has changed in between restarts, it will no longer
         be the case. This allows us to continue serving from the old pods with
         the old names.
+
+        For a similar reason, we also save the namespace.
         """
         if 'pod_name' in state:
             self.pod_name = state['pod_name']
 
+        if 'namespace' in state:
```
dolfinus, line 2228: Actually, a lot of other variables are already set in `__init__` with an old namespace value:
https://github.com/jupyterhub/kubespawner/blob/acfc4d5efc364b8eff07efbe83da3904c3a80435/kubespawner/spawner.py#L180-L188
Restoring only `namespace` could cause issues with them

totycro:
Thanks for merging this @yuvipanda!

The comment about not needing to serialize the pod name is here: https://github.com/jupyterhub/kubespawner/blob/main/kubespawner/spawner.py#L2187

kubespawner/spawner.py
totycro: Ah, you're right. I think in many cases, this doesn't manifest as an actual bug because some properties are only used to create the pod, and since we're talking about serialization here, the pod has already been created.

The values are still wrong after deserialization and might cause problems.

I guess we could either also save the dependent attributes, or recalculate them after restore.

I think saving those attribute could be more robust because e.g. the `dns_name` could have been changed since the pod has been created, but the existing pod will still have the old value no matter what.

What do you think?

dolfinus: Created #677 677

PR #662, opened 2022-11-02
Fix dict keys iteration

GeorgianaElena:
Follow-up to https://github.com/jupyterhub/kubespawner/pull/631
Fixes the bug introduced there


yuvipanda:
Thanks, @GeorgianaElena

PR #663, opened 2022-11-03
Add changelog for 4.3.0

GeorgianaElena:

PR #664, opened 2022-11-03
Make sure we run the publish workflow on every tag pushed

GeorgianaElena:
Without this, the publish workflow might fail to run for the tagged commits :/ 

see https://github.com/jupyterhub/kubespawner/pull/663#issuecomment-1302150658

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -19,6 +19,8 @@ on:
     branches-ignore:
       - "dependabot/**"
       - "pre-commit-ci-update-config"
+  tags:
+    - "**"
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -19,8 +19,8 @@ on:
     branches-ignore:
       - "dependabot/**"
       - "pre-commit-ci-update-config"
-  tags:
-    - "**"
+    tags:
+      - "**"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -19,6 +19,8 @@ on:
     branches-ignore:
       - "dependabot/**"
       - "pre-commit-ci-update-config"
+  tags:
+    - "**"
```
manics, line 23:
```suggestion
    tags:
      - "**"
```
Doesn't this need to be indented a bit more to come under `push`?

GeorgianaElena: Yes! Thanks!

consideRatio (approved):

consideRatio:
Woops that was me who broke the release job recently in https://github.com/jupyterhub/kubespawner/commit/9584c13803a4f03577806595ecc125ca7137b88d. Sorry!

I think what to do now is to not make any new commits etc, but delete the tag, and then re-push it.

consideRatio:
Or no... If we do that it won't trigger, as that commit doesn't have this fix...

@GeorgianaElena I created another changelog PR. Want to give it another go after merging it? Sorry for the trouble!

GeorgianaElena:
@consideRatio I triggered the workflow manually for that commit and it went ok. The release happened and the new version should be on pypi.

PR #666, opened 2022-11-04
maint: pyproject.toml, hatchling, tbump, .readthedocs.yaml updates

consideRatio:
This is just chore maintenance of our infrastructure surrounding the actual logic of kubespawner.

- I've iterated a bit on the RELEASE.md document, making it more succint etc and using `tbump` wich is what jupyterhub/jupyterhub use, and is a maintained project configurable via pyproject.toml. Like this, we drop use of bump2version during the release process.
- I've removed setup.py entirely in favor of using only pyproject.toml together with the hatchling build system. This is a build system that can help us in more tricky situations as well, such as when we bundle javascript to build etc.
- I've updated .readthedocs.yaml, just bumping misc versions for good measure and removed an outdated comment only relevant when we relied on a readthedocs CI runner with sphinx version 1 pre-installed etc.


consideRatio:
Got time to give this a review @yuvipanda?

yuvipanda:
What's the worst that could happen? :D Thanks for working on this, @consideRatio

consideRatio:
Thank you!!

PR #667, opened 2022-11-08
[KubeIngressProxy] Add KubeIngressProxy.ingress_specifications

### LINKED ISSUE #404: Ingress rules w/o host wont match

dolfinus:
This allows to add `host` field to ingress rules, and also set `tls` field in ingress specification. Alternative implementation of #406, should fix #404.

Next step of dividing #648 into smaller pieces

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -6,7 +6,7 @@ import ipaddress
 import json
 import operator
 import re
-from typing import List, Optional
+from typing import Dict, List, Optional
 from urllib.parse import urlparse
 
 from kubernetes_asyncio.client.models import (
@@ -60,7 +60,7 @@ try:
 except ImportError:
     from kubernetes_asyncio.client.models import V1EndpointPort as CoreV1EndpointPort
 
-from .utils import get_k8s_model, update_k8s_model
+from .utils import get_k8s_model, host_matching, update_k8s_model
 
 
 def make_pod(
@@ -739,7 +739,7 @@ def make_ingress(
     ingress_extra_labels: Optional[dict] = None,
     ingress_extra_annotations: Optional[dict] = None,
     ingress_class_name: Optional[str] = None,
-    ingress_specifications: Optional[dict] = None,
+    ingress_specifications: Optional[List[Dict]] = None,
 ):
     """
     Returns an ingress, service, endpoint object that'll work for this service
@@ -764,11 +764,12 @@ def make_ingress(
         labels=common_labels,
     )
 
-    if routespec.startswith('/'):
-        host = None
-        path = routespec
-    else:
-        host, path = routespec.split('/', 1)
+    # /myuser/myserver or https://myuser.example.com/myserver for spawned server
+    # /services/myservice or https://services.example.com/myservice for service
+    # / or https://example.com/ for hub
+    route_parts = urlparse(routespec)
+    routespec_host = route_parts.netloc or None
+    routespec_path = route_parts.path
 
     target_parts = urlparse(target)
     target_port = target_parts.port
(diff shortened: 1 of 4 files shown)


kubespawner/objects.py
```diff
@@ -829,31 +831,43 @@ def make_ingress(
...
+        if "tlsSecret" in spec
+    ]
+
     ingress = V1Ingress(
         kind='Ingress',
         metadata=ingress_meta,
         spec=V1IngressSpec(
-            rules=[
-                V1IngressRule(
-                    host=host,
```
consideRatio, line 838: It seems that `host` is a variable that we figure out based on `routespec`, but then later won't be using. I figure it should be cleaned up or similar if not needed.

Also related, now you are required to provide ingress_specifications, right? Otherwise you will attempt to create the V1IngressSpec with rules=None and tls=None - does that make sense to do?

If you are required to provide something, it would be good to error more verbosly if you aren't provided with that.

dolfinus: Ok, I've set `host` variable as a primary source of hostname, and `spec["host"]` as a fallback

dolfinus: `ingress_specifications` is not required:
```python
ingress_specifications = ingress_specifications or [{}]
```
If `ingress_specifications` is None or empty list, `[... for spec in ingress_specifications]` will return a list with one item

kubespawner/objects.py
```diff
@@ -829,31 +831,43 @@ def make_ingress(
         labels=ingress_labels,
     )
 
+    ingress_specifications = ingress_specifications or [{}]
+
+    rules = [
+        V1IngressRule(
+            host=host or spec.get("host"),
```
consideRatio, line 838: Ah now I see this involved a reference to `host`.

Should this be `spec.get("host") or host` instead though? I'm not sure at all, just wanted to make sure that at least one of us is sure about what makes sense.

consideRatio: Yeah I think this is currently weird, considering how `spec["host"]` is referenced in the tls section.

consideRatio: Since this only affects the KubeIngressProxy part etc, I figure we go with what you think makes sense here - but maybe you could add a comment about this as this is apparently quite tricky?

tests/test_objects.py
```diff
@@ -2178,6 +2178,271 @@ def test_make_ingress_external_name():
     }
 
 
+@pytest.mark.parametrize(
+    'target,ip',
+    [
+        ('http://192.168.1.10:9000', '192.168.1.10'),
+        ('http://[2001:db8::dead:babe]:9000', '2001:db8::dead:babe'),
```
consideRatio, line 2185: Could you update this to be something nobody ends up reflecting on?

kubespawner/objects.py
consideRatio: Overall, this PR LGTM, but this makes me uncertain.

I don't overview when host is None, or if there are implicit assumptions about what ingress_specification can and can't be, such as setting just `tlsSecret` and not the `host` etc, and if passing `host=None` is acceptable or not here.

If a comment clarifies it, even though its a comment indicating that we are not sure, that is a clear improvement on managing this complexity I think.

tests/test_objects.py
yuvipanda: And hopefully something slightly less morbid, at least for my sake? :D

dolfinus: If you're talking about `dead:babe` thing, this is just a copy of existing test: https://github.com/jupyterhub/kubespawner/blob/acfc4d5efc364b8eff07efbe83da3904c3a80435/tests/test_objects.py#L1980

kubespawner/objects.py
dolfinus: IMHO, the best way here is to merge host from `routespec` with items from `ingress_specifications`. But I'm still not sure about the way of merging these values.

For example, `routespec=https://my.example.com/me/service` and `ingress_specifications=[{"host": "*.example.com", "tlsSecretName": "some-secret"}]`. Do we need to create:
```yaml
spec:
  rules:
    - host: my.example.com
      path: /me/service
    - host: *.example.com
      path: /me/service
  tls:
    hosts:
      - *.example.com
    secretName: my-secret
```
or
```yaml
spec:
  rules:
    - host: *.example.com
      path: /me/service
  tls:
    hosts:
      - *.example.com
    secretName: my-secret
```
or
```yaml
spec:
  rules:
    - host: my.example.com
      path: /me/service
  tls:
    hosts:
      - *.example.com
    secretName: my-secret
```

tests/test_objects.py
consideRatio: :( can you update both?

dolfinus: Ok, done

kubespawner/objects.py
dolfinus: Options 1 and 2 can lead to the following issue.

Example:
```python
c.Jupyterhub.subdomain_host="example.com"
c.KubeIngressProxy.ingress_specifications = [{"host": "*.example.com", "tlsSecretName": "secret-name"}]
```
For user1 and user2 Jupyterhub will create `routespec="https://user1.example.com/default-server"` and routespec="https://user2.example.com/default-server"` respectively.

But then both ingress rules will look like:
```
spec:
  rules:
  - host: *.example.com
    path: /default-server
  tls:
    hosts:
    - *.example.com
    secretName: secret-name
```
and they cannot be distinct from each other.

So I've implemented option 3 which produces rules like:
```
spec:
  rules:
  - host: user1.example.com
    path: /default-server
  tls:
    hosts:
    - *.example.com
    secretName: secret-name
```
and added some tests covering these cases.

consideRatio:
:heart: :tada: :sunflower: thanks @dolfinus for helping me understand this better and providing comments so that maintenance in the future is simplified! This LGTM!

PR #672, opened 2022-11-10
Avoid class state by passing relevant config to PodReflector on instanciation

dolfinus:
Based on this discussion:
https://github.com/jupyterhub/kubespawner/pull/654#discussion_r1019049868

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -55,11 +55,6 @@ class PodReflector(ResourceReflector):
 
     kind = "pods"
 
-    # The default component label can be over-ridden by specifying the component_label property
-    labels = {
-        'component': 'singleuser-server',
-    }
-
     @property
     def pods(self):
         """
@@ -2387,8 +2382,8 @@ class KubeSpawner(Spawner):
 
     def _start_reflector(
         self,
-        kind=None,
-        reflector_class=ResourceReflector,
+        kind,
+        reflector_class,
         replace=False,
         **kwargs,
     ):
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks @dolfinus!! Retriggereing tests to run against main branch after merged #654

kubespawner/spawner.py
```diff
@@ -55,11 +55,6 @@ class PodReflector(ResourceReflector):
 
     kind = "pods"
 
-    # The default component label can be over-ridden by specifying the component_label property
-    labels = {
-        'component': 'singleuser-server',
-    }
-
```
consideRatio, line 62: This is an override of the base class ResourceReflector definition of labels, which is now being passed via `_start_reflector(labels=...)` instead. Presumably, the PodReflector or other base class then should have a constructor accepting that keyword argument...

I concluded that the class inheritance is like this: PodReflector -> ResourceReflector -> LoggingConfigurable -> Configurable -> HasTraits, and that HasTraits does indeed consume it. Okay, this makes sense then!

This is the kind of investigation I felt was needed, making this change something I appreciated having in a standalone PR.

consideRatio (approved):

kubespawner/spawner.py
dolfinus: I used this as a reference:
https://github.com/jupyterhub/kubespawner/blob/c6bb2a1e3a0ddec6044b4ef8cb31f8fa50571c7c/kubespawner/proxy.py#L271-L279

consideRatio: I appreciated that! It made us consistent, something I also checked for when reviewing this.

PR #673, opened 2022-11-14
docs: relocate docs/requirements.txt into pyproject.toml

consideRatio:
When we have autodoc_traits, inspecting a python package (kubespawner) to build documentation, it makes sense to put the requirements in the package itself rather than in a dedicated docs/requirements.txt file I figure.


choldgraf (approved):
LGTM

consideRatio:
I'll go for a self-merge, thinking this doesn't merit further review!

PR #678, opened 2022-11-17
Allow to watch multiple namespaces at the same time

dolfinus:
This should solve 2 issues:

- After #657, if someone started notebook pod in namespace `namespace1`, stopped Jupyterhub without stopping the notebook, and set `c.KubeSpawner.namespace=namespace2`, then Jupyterhub will start watching pods only in `namespace2` and forget about pods in `namespace1`, thinking that they are gone.
- #617. Profiles list can contain `kubespawner_overrides` with a custom namespace, allowing users to start some notebooks (for example, requiring GPU) in a dedicated namespace, when other notebooks will be started in the default one.

These issues can be solved by using `enable_user_namespaces=True`, so KubeSpawner will watch for all namespaces in the cluster. But this requires ClusterRole which may not be an option in some instances (for example, managed k8s).

Changes made in this PR requires just ServiceAccount (in JH namespace) + Role + RoleBinding (in the target namespace), allowing KubeSpawner to do everything it needs (except creating new namespaces).

This is achieved by creating a reflector per namespace. Reflector start function is also moved from the constructor to `_start()` method, allowing to set namespace after applying the `user_options`.

Added tests for cases:

- `profile_list` (in both list and callable form) returnes a profile with `kubespawner_overrides` containing different namespace than set in `c.KubeSpawner.namespace` or `c.KubeSpawner.user_namespace_template` (fails without changes from this PR).
- Restored pod state contains different namespace than set in `c.KubeSpawner.namespace` or `c.KubeSpawner.user_namespace_template` (#657).


dolfinus:
This PR is quite large and can contain some arguable changes, waiting for maintainers to discuss.

dolfinus:
@consideRatio is this approach valid, or I missed something?

minrk (approved):

consideRatio (changes requested):
Hey @dolfinus thanks for thorough work!

To align our understanding, do you agree that its not a ServiceAccount needed in the new namespace (said in PR description), but permissions granted to JupyterHub's/KubeSpawner's namespace local ServiceAccount via Role and RoleBinding from another namespace? It would be good if the PR description included a clear description as this relates to testing and use of this feature as well.

This starts to lgtm, and i figure this approach is reasonable to go for for use by advanced users.

kubespawner/spawner.py
```diff
@@ -2390,11 +2398,11 @@ async def progress(self):
                 break
             await asyncio.sleep(1)
 
-    def _start_reflector(
+    async def _start_reflector(
         self,
-        kind,
-        reflector_class,
-        replace=False,
+        kind: str,
```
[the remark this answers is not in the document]
dolfinus: `kind` is still string, only `key` is changed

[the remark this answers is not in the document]
consideRatio: Ooops right!

kubespawner/spawner.py
```diff
@@ -2455,31 +2482,64 @@ def _start_watching_events(self, replace=False):
...
             kind="events",
             reflector_class=EventReflector,
             fields={"involvedObject.kind": "Pod"},
             omit_namespace=self.enable_user_namespaces,
             replace=replace,
         )
 
-    def _start_watching_pods(self, replace=False):
-        """Start the pod reflector
+    async def _start_watching_pods(self, replace=False):
```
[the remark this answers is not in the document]
dolfinus: This is caused by changing `def _start_reflector` with `async def _start_reflector`.

Originally this method was called from constructor, so it used `asyncio.create_task` and `asyncio.ensure_future` to call async functions. Now reflectors are started in `_start`, so they there replaced with plain `await`

kubespawner/spawner.py
```diff
@@ -2409,44 +2417,63 @@ def _start_reflector(
...
+                    sys.exit(1)
+
+        else:
+            # Separate reflector for each namespace (does not require ClusterRole),
+            # avoid failing if it cannot be created.
+            # Namespace can be already deleted, or there can be missing access to only one namespace,
+            # and halting the entire Jupyterhub instance is not the best idea.
 
-        previous_reflector = self.__class__.reflectors.get(key)
+            on_reflector_failure = None
```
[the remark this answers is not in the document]
dolfinus: Updated the comment

kubespawner/spawner.py
```diff
@@ -206,26 +210,27 @@ def _await_pod_reflector(method):
...
-                and not self.event_reflector.first_load_future.done()
-            ):
-                await self.event_reflector.first_load_future
+            if not self.events_enabled:
+                return
+
+            await self._start_watching_pods()
             return await method(self, *args, **kwargs)
 
         return async_method
```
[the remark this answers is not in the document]
dolfinus: Done

kubespawner/spawner.py
```diff
@@ -116,21 +117,26 @@ class KubeSpawner(Spawner):
...
+        if self.enable_user_namespaces:
+            # one reflector fo all namespaces
+            return (kind, None)
+
+        return (kind, self.namespace)
+
     @property
     def pod_reflector(self):
         """
         A convenience alias to the class variable reflectors['pods'].
```
consideRatio, line 136: pod_reflector and event_reflector includes outdated docstrings

kubespawner/spawner.py
```diff
@@ -2455,31 +2463,64 @@ def _start_watching_events(self, replace=False):
...
+
+    async def _stop_watching_events(self):
+        """Stop events reflector"""
+        await self._stop_reflector("events")
+
+    async def _stop_watchers(self):
+        """Stop reflectors for current spawner instance"""
+        await self._stop_reflector()
+        if self.events_enabled:
+            await self._stop_watching_events()
```
consideRatio, line 2512: These functions were introduced but isn't referenced by anything besides themselves. I'd like to remove them to reduce complexity of the codebase.

kubespawner/spawner.py
```diff
@@ -2455,31 +2463,64 @@ def _start_watching_events(self, replace=False):
...
+
+    @classmethod
+    async def _stop_all_reflectors(cls):
+        """Stop reflectors for all instances"""
```
consideRatio, line 2516: I saw that this was soleley used for tests, maybe it belongs in the conftest.py, but it could live here as well - I figure it merits a clarification though.

```suggestion
        """Stop reflectors for all instances, a function used when running tests."""
```

consideRatio (commented):
The PR description mentions that a ServiceAccount needs to be setup in other namespaces. I think actually what's needed is to create a Role and RoleBinding in the other namespace, where the Role includes the relevant KubeSpawner permissions to act in that namespace, and the RoleBinding references the ServiceAccount available to KubeSpawner in the namespace where JupyterHub/KubeSpawner runs.

The Role and RoleBinding permissions granted to the JupyterHub Helm chart can be used as an example of setting up a Role and RoleBinding for KubeSpawner's ServiceAccount.
  - [Role](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/b88b1a2969eaec490e10ac956b745deb1e9588a3/jupyterhub/templates/hub/rbac.yaml#L2-L14)
  - [RoleBinding](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/b88b1a2969eaec490e10ac956b745deb1e9588a3/jupyterhub/templates/hub/rbac.yaml#L16-L29)
  
If you agree on this understanding, can you update the PR description to reflect this as well?

kubespawner/spawner.py
dolfinus: Done

kubespawner/spawner.py
dolfinus: Done

kubespawner/spawner.py
dolfinus: Fixed

dolfinus:
Yes, I've updated PR description, ServiceAccount should be created in the JH namespace, and RoleBinding in target (user) namespace.

.python-version
```diff
@@ -0,0 +1 @@
+3.9.16
```
consideRatio, line 1: File mistakenly committed

kubespawner/spawner.py
```diff
@@ -2455,31 +2463,50 @@ def _start_watching_events(self, replace=False):
...
 
+    async def _stop_reflector(self, kind: str):
+        """Stop a shared reflector on the KubeSpawner class"""
+
+        key = self._get_reflector_key(kind)
+        reflector = self.__class__.reflectors.pop(key, None)
+
+        if reflector:
+            await reflector.stop()
+
```
consideRatio, line 2499:
```suggestion
```

kubespawner/spawner.py
```diff
@@ -2409,44 +2397,64 @@ def _start_reflector(
...
+        if self.enable_user_namespaces:
+            # One reflector for all namespaces (requires ClusterRole).
+            # Stop JupyterHub if something went wrong
 
-        previous_reflector = self.__class__.reflectors.get(key)
+            def on_reflector_failure():
+                self.log.critical(
+                    "%s reflector failed, halting Hub.",
+                    key.title(),
+                )
```
consideRatio, line 2409: Since `key` is a tuple, title() will error.

```suggestion
                self.log.critical(f"{key} reflector failed, halting Hub.")
```

kubespawner/spawner.py
dolfinus: Fixed

consideRatio (commented):
I'd like to think about this a bit more before merging, but I'll get it done!

I'm thinking about:
- delaying calls to start functions until they are absolutely needed, possibly making the first spawn slower due to this, and errors show up later if things are misconfigured
- having both the `_await_pod_reflector` decorator used for `.poll()` and `.stop()`, but in-function logic representing that in `.start()` to run after a hook.
- performance of calling start functions

I'll need to context switch and focus on other things for a while now, but I'll try get back to these questions before the week ends

.python-version
dolfinus: Fixed

kubespawner/spawner.py
dolfinus: Using f-strings in logs is not recommended. Intergrations like Sentry or Elastic APM are using template strings to count number of errors caused by a specific issue, but f-strings produces different strings for each substitution values combination

consideRatio: Makes sense, I updated the code suggestion and mentioned your thought in an existing discussion at https://github.com/jupyterhub/team-compass/issues/625!

kubespawner/spawner.py
```diff
@@ -2409,44 +2397,64 @@ def _start_reflector(
...
+            await catch_reflector_start(self.__class__.reflectors[key].start())
 
         if replace and previous_reflector:
             # we replaced the reflector, stop the old one
-            asyncio.ensure_future(previous_reflector.stop())
+            await previous_reflector.stop()
+
+        reflector = self.__class__.reflectors[key]
+        if not reflector.first_load_future.done():
+            await catch_reflector_start(reflector.first_load_future)
```
consideRatio, line 2452: I think this section isn't needed.

In reflector.py I see the start function `await _list_and_update()` in a `try` block. I think that either the `start()` function will have raised an error, or it will have set `first_load_future` already.

consideRatio (changes requested):
I'm not confident about this implementation for low level technical details.

1. Race conditions
   It seems reflector objects `start()` function is supposed to be run only one time, or they raise errors. Are we sure this won't happen?

   Previously we only started reflectors on spawner construction and made checks that they had initialized in functions that relied on them. Now we instead call the function to start them repeatedly from each function that relies on them, and we are forced to ensure we have a very robust implementation here. I'm not confident we do yet.
2. Avoiding risk of a performance regression
   We now call the start reflector functions very often. Doing that we should ideally return very quickly.

I'm not sure on the implementation to go for, but I think it could make sense to let the `_start_watching_pods` and `_start_watching_events` functions contain logic like:

```python
key = self._get_reflector_key(kind)
reflector = self.__class__.reflectors.get(key)
if reflector:
    if not reflector.first_load_future.done():
        await reflector.first_load_future
    return reflector
else:
    return await self._start_reflector(
        kind="pods",
        reflector_class=PodReflector,
        labels={"component": self.component_label},
        omit_namespace=self.enable_user_namespaces,
        replace=replace,
    )
```

But actually, I realize that this won't work either, because its `_start_reflector` that currently does the step updating `self.__class__.reflectors[key]`, and that has to be done before awaiting the startup to avoid race conditions...

Argh I think this is very tricky and I think this could be the cause of a regression. I figure the action point would be to try ensure its not and convey that. I'm :+1: for this if I think its very unliklely it causes a race condition issue.

If you rework this, I think its perfectly fine btw if you want to either adopt or stop adopting the pattern of decorators with `_await_x_reflector` decorators. I think maybe it doesn't make as much sense any more to have those now that they are only partially usable since you sometime have to await reflectors mid way along a function and not directly before the function is called.

kubespawner/spawner.py
```diff
@@ -2409,44 +2397,64 @@ def _start_reflector(
...
+                    await func
+                except Exception:
+                    self.log.exception(f"Reflector for {kind} failed to start.")
+                    sys.exit(1)
+
+        else:
+            # Created a dedicated reflector for each namespace (does not require ClusterRole).
+            # Namespace can be already deleted, or there can be no access a specific namespace, but not others,
+            # so halting the entire Jupyterhub instance is not the best idea.
+            # Instead, just halt the failing spawner.
```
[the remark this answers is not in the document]
dolfinus: Done

dolfinus:
> It seems reflector objects start() function is supposed to be run only one time, or they raise errors. Are we sure this won't happen?

> We now call the start reflector functions very often. Doing that we should ideally return very quickly.

`_start_reflector()` returns current reflector instance if it has been created before and successfully started.

kubespawner/spawner.py
dolfinus: Fixed

kubespawner/spawner.py
```diff
@@ -2409,44 +2385,61 @@ def _start_reflector(
...
-                    self.log.exception(f"Reflector for {kind} failed to start.")
-                    sys.exit(1)
-
-            asyncio.create_task(catch_reflector_start())
+            await catch_reflector_start(self.__class__.reflectors[key].start())
 
         if replace and previous_reflector:
             # we replaced the reflector, stop the old one
-            asyncio.ensure_future(previous_reflector.stop())
+            await previous_reflector.stop()
```
consideRatio, line 2437: I think maybe we don't need to await this, but only ensure its scheduled to be done, what do you think?

```suggestion
            asyncio.ensure_future(previous_reflector.stop())
```

kubespawner/spawner.py
```diff
@@ -2409,44 +2385,61 @@ def _start_reflector(
...
-
-            async def catch_reflector_start():
-                try:
-                    await f
-                except Exception:
-                    self.log.exception(f"Reflector for {kind} failed to start.")
-                    sys.exit(1)
-
-            asyncio.create_task(catch_reflector_start())
+            await catch_reflector_start(self.__class__.reflectors[key].start())
```
consideRatio, line 2433: What do you think about:

1. Putting the key and previous_reflector assignments (line 2423-2424) in the beginning of this function
2. Using the key's namespace in the error message when failing to startup (i figure it can help help users setup the relevant permissions)
3. Removing the `if replace or not previous_reflector:` section on line 2427, and instead having an early `if previous_reflector and not replace:` that returns directly if nothing is to be done.

Maybe this isn't impacting performance at all, but I'm a bit extra cautious here because we are making changes that could cause a regression for users that may not use this feature.

kubespawner/spawner.py
```diff
@@ -2663,6 +2666,12 @@ async def _start(self):
         if self.enable_user_namespaces:
             await self._ensure_namespace()
 
+        # namespace can be changed via kubespawner_override, start watching pods only after
+        # load_user_options() is called
+        await self._start_watching_pods()
+        if self.events_enabled:
+            await self._start_watching_events()
```
consideRatio, line 2673: Previously we started up the reflectors concurrently, and waited for the startup when needed at a later time.

```python
        self._start_watching_pods()
        if self.events_enabled:
            self._start_watching_events()
```

We can at least do a slight improvement from waiting directly on A and then B, but starting both A and B and then waiting for both.

```suggestion
        start_futures = [self._start_watching_pods()]
        if self.events_enabled:
            start_futures.append(self._start_watching_events())
        await asyncio.wait(start_futures)
```

consideRatio: I note that I used `asyncio.wait` as compared to `asyncio.gather`, they are very similar I think but I'm not sure if either this code block should use `asyncio.gather` or the one in `_stop_all_reflectors` should use `asyncio.wait` - or if it doesn't matter much at all.

I understand that `.wait` doesn't care about the return values, which we don't here, but `.gather` does.

kubespawner/spawner.py
dolfinus: Fixed

kubespawner/spawner.py
dolfinus: Done

kubespawner/spawner.py
dolfinus: According to documentation:
https://docs.python.org/3/library/asyncio-task.html#asyncio.gather
https://docs.python.org/3/library/asyncio-task.html#asyncio.wait

`wait` does not propagate exception from a coroutine, but `gather` does:
```
If return_exceptions is False (default), the first raised exception is immediately propagated to the task that awaits on gather(). Other awaitables in the aws sequence won’t be cancelled and will continue to run.
```

So I used `gather` to propagate raised exception to the spawner, but with some wrapper to cancel tasks which are not done yet.

consideRatio (approved):
This LGTM, but I'm not as confident I'd like to be that this doesn't cause a regression in some way since its quite complicated change.

@minrk being very experienced with async/await and the event loop, could you review this as well?

consideRatio:
Thank you soo much for your very thorough work on this @dolfinus!!! :heart: :tada: :sunflower:

PR #680, opened 2022-11-22
Update docs on setting up a development environment

shaneknapp:
just updating the docs a bit to make running this locally a little easier.

@yuvipanda

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -151,13 +151,13 @@ rules on your computer. We will now walk you through the steps to get going:
 
     ```sh
     npm install configurable-http-proxy
-    export PATH=$(pwd)/node_modules/.bin:$PATH
+    export PATH=/<path>/<to>/node_modules/.bin:$PATH
     ```
 
 1.  Start JupyterHub
 
     ```sh
-    # Run this from the repo's root directory where the preconfigured
+    # Run this from the kubespawner repo's root directory where the preconfigured
     # jupyterhub_config.py file resides!
     jupyterhub
     ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -65,7 +65,7 @@ rules on your computer. We will now walk you through the steps to get going:
 
 1.  Install VirtualBox by [downloading and running an
     installer](https://www.virtualbox.org/wiki/Downloads), or
-    [Docker|https://www.docker.com/].
+    [Docker](https://www.docker.com/).
```
consideRatio, line 68: Ah from https://minikube.sigs.k8s.io/docs/drivers/ I understand that docker is available as a driver, and even preferred while virtualbox isn't preferred.

Have you tried getting kubespawner setup with minikube and docker as a driver to create the k8s cluster and verified that you could run the kubespawner test suite against it successfully?

CONTRIBUTING.md
```diff
@@ -151,13 +152,13 @@ rules on your computer. We will now walk you through the steps to get going:
 
     ```sh
     npm install configurable-http-proxy
-    export PATH=$(pwd)/node_modules/.bin:$PATH
+    export PATH=/<path>/<to>/node_modules/.bin:$PATH
```
consideRatio, line 155: If following the command `npm install configurable-http-proxy`, the path to node_modules will be from the local directory as known by `$(pwd)`.

Due to that I think what we have already is quite good.

consideRatio:
Thank you @shaneknapp !

shaneknapp:
ugh thank YOU for following up on this!  Real Life[tm] is hard!  XD

PR #682, opened 2022-11-23
Test restoring pod name from previos state

dolfinus:
Added test that if `c.KubeSpawner.pod_name_template` was changed with restarting a JupyterHub instance, the restored server state contains the same pod name as before restart.

Splitting #678 into smaller pieces.

### DIFF
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
     await spawner.stop()
 
 
+async def test_spawn_start_restore_pod_name(
+    kube_ns,
+    kube_client,
+    config,
+    hub,
+    exec_python,
+):
+    # Emulate stopping Jupyterhub and starting with different settings while some pods still exist
+
+    # Save state with old config
+    old_spawner = KubeSpawner(
+        hub=hub,
+        user=MockUser(name="start"),
+        config=config,
+        api_token="abc123",
+        oauth_client_id="unused",
+    )
+    old_spawner_pod_name = old_spawner.pod_name
+
+    # save state
+    await old_spawner.start()
+    old_state = old_spawner.get_state()
+    await old_spawner.stop()
+
+    # Change config
+    config.KubeSpawner.pod_name_template = 'new-{username}--{servername}'
+
+    # Load old state
+    spawner = KubeSpawner(
+        hub=hub,
+        user=MockUser(name="start"),
+        config=config,
+        api_token="abc123",
+        oauth_client_id="unused",
+    )
+    spawner.load_state(old_state)
+
+    # previous pod name is restored
+    assert spawner.pod_name == old_spawner_pod_name
+
+    # empty spawner isn't running
+    status = await spawner.poll()
+    assert isinstance(status, int)
+
+    pod_name = spawner.pod_name
+
+    # start the spawner
+    url = await spawner.start()
+
+    # verify the pod exists
+    pods = (await kube_client.list_namespaced_pod(kube_ns)).items
+    pod_names = [p.metadata.name for p in pods]
+    assert pod_name in pod_names
+
+    # pod should be running when start returns
+    pod = await kube_client.read_namespaced_pod(namespace=kube_ns, name=pod_name)
+    assert pod.status.phase == "Running"
+
+    # verify poll while running
+    status = await spawner.poll()
+    assert status is None
+
+    # make sure spawn url is correct
+    r = await exec_python(check_up, {"url": url}, _retries=3)
+    assert r == "302"
+
+    # stop the pod
+    await spawner.stop()
+
+    # verify pod is gone
+    pods = (await kube_client.list_namespaced_pod(kube_ns)).items
+    pod_names = [p.metadata.name for p in pods]
+    assert pod_name not in pod_names
+
+    # verify exit status
+    status = await spawner.poll()
+    assert isinstance(status, int)
+
+
 async def test_get_pod_manifest_tolerates_mixed_input():
     """
     Test that the get_pod_manifest function can handle a either a dictionary or
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -557,7 +557,7 @@ async def test_spawn_start_restore_pod_name(
     )
     spawner.load_state(old_state)
 
-    # previous pod name is restored
+    # previous pod name is restored by the load_state call
     assert spawner.pod_name == old_spawner_pod_name
 
     # empty spawner isn't running
--- a/tests/test_spawner.py
+++ b/tests/test_spawner.py
@@ -589,7 +589,7 @@ async def test_spawn_start_restore_pod_name(
     # stop the pod
     await spawner.stop()
 
-    # verify pod is gone
+    # verify pod with old name is gone
     pods = (await kube_client.list_namespaced_pod(kube_ns)).items
     pod_names = [p.metadata.name for p in pods]
     assert pod_name not in pod_names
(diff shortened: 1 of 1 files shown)


tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+
+    # Load old state
+    spawner = KubeSpawner(
+        hub=hub,
+        user=MockUser(name="start"),
+        config=config,
+        api_token="abc123",
+        oauth_client_id="unused",
+    )
+    spawner.load_state(old_state)
```
consideRatio, line 558: Is everything above representing the emulation of JupyterHub being stopped, configured, and started?

tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+
+    # previous pod name is restored
+    assert spawner.pod_name == old_spawner_pod_name
```
consideRatio, line 561:
```suggestion
    # previous pod name is restored by the load_state call
    assert spawner.pod_name == old_spawner_pod_name
```

tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+
+    # empty spawner isn't running
+    status = await spawner.poll()
+    assert isinstance(status, int)
```
consideRatio, line 565: Hmmmmm, is this a claim being tested? You assert that spawner.poll() is an int here - is that to assert it isn't running?

tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+
+    pod_name = spawner.pod_name
+
+    # start the spawner
+    url = await spawner.start()
+
+    # verify the pod exists
+    pods = (await kube_client.list_namespaced_pod(kube_ns)).items
+    pod_names = [p.metadata.name for p in pods]
+    assert pod_name in pod_names
```
consideRatio, line 575:
```suggestion
    # verify pod with old name now exists
    pods = (await kube_client.list_namespaced_pod(kube_ns)).items
    pod_names = [p.metadata.name for p in pods]
    assert pod_name in pod_names
```

tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+    await spawner.stop()
+
+    # verify pod is gone
+    pods = (await kube_client.list_namespaced_pod(kube_ns)).items
+    pod_names = [p.metadata.name for p in pods]
+    assert pod_name not in pod_names
+
+    # verify exit status
+    status = await spawner.poll()
+    assert isinstance(status, int)
```
consideRatio, line 599: Verify that you got any form of exit status - 0 or >0?

tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+    assert status is None
+
+    # make sure spawn url is correct
+    r = await exec_python(check_up, {"url": url}, _retries=3)
+    assert r == "302"
+
+    # stop the pod
+    await spawner.stop()
+
+    # verify pod is gone
```
consideRatio, line 592:
```suggestion
    # verify pod with old name is gone
```

tests/test_spawner.py
```diff
@@ -520,6 +520,85 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+    pod = await kube_client.read_namespaced_pod(namespace=kube_ns, name=pod_name)
+    assert pod.status.phase == "Running"
+
+    # verify poll while running
+    status = await spawner.poll()
+    assert status is None
+
+    # make sure spawn url is correct
+    r = await exec_python(check_up, {"url": url}, _retries=3)
+    assert r == "302"
```
consideRatio, line 587: This seems like something relevance to test not captured by the title of the test. Can you provide a note about how spawn url relate to the restored state etc?

consideRatio (commented):
Thank you for your work on this @dolfinus !!!

I'm not so confident about this in general so it led to quite a few comments =/

tests/test_spawner.py
dolfinus: This test is just a copy of `test_spawn_start`, and it has the same asserts:
https://github.com/jupyterhub/kubespawner/blob/7e09fbb173319570e72436659f066e7dce189ac9/tests/test_spawner.py#L172

this comment and others from below are not related to this PR.

tests/test_spawner.py
dolfinus: `user.spawn()` calls `Spawner.get_state()` and saves it to the database:
https://github.com/jupyterhub/jupyterhub/blob/73a33ed5fc90f60974141516c84f317f5cebe5e5/jupyterhub/user.py#L859-L860

Jupyterhub startup process calls `jupyterhub.initialize()`, it calls `jupyterhub.init_spawners`:
https://github.com/jupyterhub/jupyterhub/blob/7c5662ee52c98251d6311fd90df6e3542f3d1c71/jupyterhub/app.py#L2873

This method creates `Spawner` instances by fetching all the rows from the database:
https://github.com/jupyterhub/jupyterhub/blob/7c5662ee52c98251d6311fd90df6e3542f3d1c71/jupyterhub/app.py#L2578

`Spawner` constructor calls `spawner._new_spawner()`:
https://github.com/jupyterhub/jupyterhub/blob/73a33ed5fc90f60974141516c84f317f5cebe5e5/jupyterhub/user.py#L241-L248
https://github.com/jupyterhub/jupyterhub/blob/73a33ed5fc90f60974141516c84f317f5cebe5e5/jupyterhub/user.py#L273

Which loads state of each spawner from the database:
https://github.com/jupyterhub/jupyterhub/blob/73a33ed5fc90f60974141516c84f317f5cebe5e5/jupyterhub/user.py#L478

And then all initialized spawners state is checked by calling `spawner.poll()`:
https://github.com/jupyterhub/jupyterhub/blob/7c5662ee52c98251d6311fd90df6e3542f3d1c71/jupyterhub/app.py#L2494

This test is emulating the process of loading spawner state from the database, and using it for starting the pod.

If `c.JupyterHub.cleanup_servers` is `True`, then all spawners' pods are stopped before JupyterHub is shutted down:
https://github.com/jupyterhub/jupyterhub/blob/7c5662ee52c98251d6311fd90df6e3542f3d1c71/jupyterhub/app.py#L2935-L2938

That's why `spawner.stop()` is used here.

Why there is no test with `c.JupyterHub.cleanup_servers=False`?
Well, then JupyterHub is stopped, old spawner instance is gone because the entire JupyterHub process is down. And new instance with the same name and state is created only after JupyterHub process is started again.

So there cannot be 2 spawners with the same parameters, this situation cannot be achieved in JupyterHub, and there is no need to test that.

tests/test_spawner.py
dolfinus: This test is just a copy of `test_spawn_start`, and it has the same asserts:
https://github.com/jupyterhub/kubespawner/blob/7e09fbb173319570e72436659f066e7dce189ac9/tests/test_spawner.py#L172

This particular assert checks that pod is started and responses for status check calls. This is done to be sure that nothing is broken in the pod if its state was restored from the database

dolfinus:
Any updates?

dolfinus:
Ping

consideRatio (approved):
@dolfinus you are great, thank you for breaking apart things to make them easier to review and think about!

Sorry for being slow on reviewing, there is a lot of things on my plate and I'm limited in time.

PR #700, opened 2023-02-08
[POC] Auth aware profile list

npapapietro:
# Motivaion

The overall goal of this PR is to allow auth logic in KubeSpawner.profile_list depending on the instance of OAuthenticator chosen. This PR is meant to continue the discussions [here](https://discourse.jupyter.org/t/tailoring-spawn-options-and-server-configuration-to-certain-users/8449) as well as a sibling PR to https://github.com/jupyterhub/oauthenticator/pull/571. 

# Changes

* `_options_form_default`: Returns callable if `self.profile_list` is not empty
* `_render_options_form_dynamically`: Checks if `self.profile_list` is callable, and generates the list, then proceeds to filter out profiles (and profile options) if the authenticator class meets the auth criteria
* `_filter_profile_options_form`: New async function that filters profiles based on user configured `oauthenticator_override`

Per reviewer feedback I will proceed with extending some test coverage to this new logic. No auth is currently being tested beyond Dummy, so this will require some review as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,13 +19,13 @@ repos:
 
   # Autoformat: Python code
   - repo: https://github.com/pycqa/isort
-    rev: 5.11.4
+    rev: 5.12.0
     hooks:
       - id: isort
 
   # Autoformat: Python code
   - repo: https://github.com/psf/black
-    rev: 22.12.0
+    rev: 23.1.0
     hooks:
       - id: black
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -10,7 +10,7 @@ import os
 import string
 import sys
 import warnings
-from functools import partial, wraps
+from functools import lru_cache, partial, wraps
 from urllib.parse import urlparse
 
 import escapism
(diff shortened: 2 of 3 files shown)


consideRatio:
@npapapietro can you try to clarify the problem you wish to solve, and the limitations of for example handling management of the profile list using hooks as examplified in [the discourse post](https://discourse.jupyter.org/t/tailoring-spawn-options-and-server-configuration-to-certain-users/8449)?

If I can understand that clearly, its far easier to consider the strategy you have come up with on a high level, and then also on a more detailed level looking at code changes.

npapapietro:
I have been using the methods described in that post since it was suggest in my customer's jupyterhub installations :). It works well, but I think that pulling the key auth changes into the kubespawner might improve usability, deployment and security of kubernetes jupyterhub installation (and perhaps other installations as well).

1. Configuration as Code: The `kubespawner_override` sets a good precedent on how to define overrides (such as labels, nodeselectors, service accounts, etc) in configuration (ArgoCD or Flux) for helm releases of jupyterhub. Increasing this extra configurability by `oauthenticator_override` would allow for auth based profiles by configuration.
2. Security: Certain customers of mine have environments that utilize jupyterhub but run on closed or tightly controlled networks/systems with policies that forbid inlining or `exec` type behaviors (which is how I'm currently using it with `.Values.hub.extraConfig`.

This is a QOL feature in my opinion and by no means a hard requirement for most of my jupyterhub installations, but I have bandwidth to take a whack at it and see if the community would benefit from this as well.

Thank you

PR #718, opened 2023-04-10
Drop support for k8s 1.20-1.22 (stop testing against it)

consideRatio:
Our next release is a major release, and z2jh will drop support for k8s 1.22 as well when including kubespawner and making its major release - see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3040

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -6,8 +6,8 @@
 
 ### Breaking changes
 
-- Support for use against k8s 1.20-1.22 is no longer maintained, please upgrade
-  to k8s 1.23+ to ensure function.
+- Versions of K8s older than 1.23 are no longer supported, they may work but
+  this is not guaranteed.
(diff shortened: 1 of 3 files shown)


docs/source/changelog.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking changes
 
+- Support for use against k8s 1.20-1.22 is no longer maintained, please upgrade
+  to k8s 1.23+ to ensure function.
```
manics, line 10:
```suggestion
- Support for K8s 1.20-1.22 is no longer maintained, please upgrade
  to k8s 1.23+ to ensure continued working.
```

consideRatio: Continued working, sounds weird, is this an acceptable sentence?

manics: Another option, "Versions of K8s older than 1.23 are no longer supported, they may work but this is not guaranteed"?
.
Anyway, go ahead and merge whatever wording you prefer.

PR #721, opened 2023-04-18
docs: add changelog for 5.0.0

consideRatio:
- Closes #717 

Note tests failure are intermittent and unrelated, caused by k3s having made a release but not provided binaries etc. to the `latest` release channel.

PR #724, opened 2023-04-27
Allow building more complex profile_list templates

yuvipanda:
The profile_list HTML / CSS was a single string embedded into the python code,
and modifications were allowed as a single traitlet that had to be a fully complete jinja2
template.

This becomes difficult to manage very quickly, and means we lose out on all the compositional
benefits of using jinja2 templates.

This PR will:
- Split the existing HTML / CSS string into a separate HTML & CSS file
- Switch to `FilesystemLoader`, so we can split out the one big HTML / CSS file into
   multiple smaller files and compose them together
- Add the config `additional_profile_form_template_paths` to allow admins to set extra paths where templates can be found. This
   allows admins to customize the profile_list form easily without having to shove everything
   into one file. This is stolen from JupyterHub's `JupyterHub.template_paths`.
- Add a pre-commit hook to autoformat the default templates we ship
- Add some docs on how admins can continue to customize the profile_list template by setting
  the string, but also recommends using the directory based approach for more complex changes.
  This allows the change to be backwards compatible.


manics:
https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime recommends using [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources) instead of relying on direct filesystem access.

yuvipanda:
@manics it looks like that would only work on python 3.10+ without adding an external dependency. From https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime

> [importlib.resources](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources) was added to Python 3.7. However, the API illustrated in this code (using files()) was added only in Python 3.9, [[3]](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#files-api) and support for accessing data files via namespace packages was added only in Python 3.10 [[4]](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#namespace-support) (the data subdirectory is a namespace package under the root package mypkg). Therefore, you may find this code to work only in Python 3.10 (and above). For other versions of Python, you are recommended to use the [importlib-resources](https://pypi.org/project/importlib-resources) backport which provides the latest version of this library.

I'd say it's ok to let this be, maybe add a note to remove it once we are python 3.10+ only?

yuvipanda:
@manics the alternative is to add the additional dependency, which we currently don't have. I just checked to see if any jupyterhub or kubespawner dependency transitively includes that - turns out the answer is no. If you strongly prefer, I will add that dependency?

manics:
If there's no easy fix then leave it as it is unless anyone else has thoughts

yuvipanda:
hmm, not sure if the test failures are related.

yuvipanda:
Per @consideRatio the test failures are unrelated https://github.com/jupyterhub/kubespawner/pull/721#issue-1673545073****

consideRatio:
> Per @consideRatio the test failures are unrelated https://github.com/jupyterhub/kubespawner/pull/721#issue-1673545073****

@yuvipanda the failures I meant would have shown up during the action-k3s-helm step, not in pytest. It seems that the failure seen here is recurring for the py37 env with oldest dependencies.

But, they seem to be indepent of this PR as when I triggered the test workflow in the main branch, I saw the failures as well: https://github.com/jupyterhub/kubespawner/actions/runs/4824271639

yuvipanda:
@consideRatio makes sense. Let's investigate that separately from this PR?

consideRatio:
@yuvipanda yes I agree, separate investigation.

I'm considering an conditional dependency for `importlib_resources`. I've seen Min do similar things in several places.

[Python version constrained requirement](https://github.com/jupyterhub/jupyterhub/blob/3578001fab54c927fb43ec5d5116d046ec458b66/requirements.txt#L4), that [is conditionally imported](https://github.com/jupyterhub/jupyterhub/blob/3578001fab54c927fb43ec5d5116d046ec458b66/jupyterhub/traitlets.py#L8-L12).

Reading this not confident on the implications or meaning, I would feel at ease knowing we do that.

> It is strongly recommended that, if you are using data files, you should use [importlib.resources](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources) to access them.

yuvipanda:
@consideRatio ok, I've expanded the scope of this PR a little bit, and in the process got rid of the manual template loading :)

kubespawner/spawner.py
```diff
@@ -2951,9 +2909,20 @@ def _env_keep_default(self):
...
+
+        loader = FileSystemLoader(
+            self.additional_profile_form_template_paths
+            + [os.path.join(os.path.dirname(__file__), 'templates')]
         )
+
+        if self.profile_form_template != "":
+            # Admin has custom set the profile_form_template as a templated string
+            # so we use that directly
+            profile_form_template = Environment(loader=loader).from_string(
```
consideRatio, line 2923: This was previously `Environment(loader=BaseLoader)` and not `FileSystemLoader`.

Should it be that the FileSystemLoader should be initialized in the `else` clause and the `if` clause declares `loader=BaseLoader` like before?

yuvipanda:
Alright, I've completely removed the usage of `__file__` from here!

pyproject.toml
```diff
@@ -121,3 +121,7 @@ tag_template = "{new_version}"
 
 [[tool.tbump.file]]
 src = "kubespawner/_version.py"
+
```
consideRatio, line 124:
```suggestion

# djlint is used for autoformatting jinja templates
#
# ref: https://www.djlint.com/docs/formatter/
#
```

consideRatio (commented):
I love seeing these as standalone files, and I'm happy to see a jinja autoformatter we can use with pre-commit!

I think the `additional_profile_form_template_paths` sounds reasonable, thinking about the option of simply having `profile_form_template_paths` and letting its `@default(...)` function evaluate something. Maybe not though, then its far easier to override it when you want to append to it when configuring this via YAML based config rather than Python based config. So, maybe this is the way to go as it is.

Are the .html / .css files solely extracted and autoformatted - but not changed?

kubespawner/spawner.py
```diff
@@ -1554,6 +1489,14 @@ def _validate_image_pull_secrets(self, proposal):
         """,
     )
 
+    @default("profile_form_template")
+    def _profile_form_template(self):
```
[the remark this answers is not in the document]
yuvipanda: I got rid of this one :)

kubespawner/spawner.py
yuvipanda: This is necessary to use the new loaders, as it allows admins to continue to use composition of the default elements we ship. For example, in https://github.com/jupyterhub/kubespawner/pull/724/files#diff-47559ea38f5c75d2661f1c86c79b031011275298348501fcec8a7d80f23f0dbb, we `{% include "style.css" %}`, and that only works because it's the `ChoiceLoader` rather than the `BaseLoader`. The `BaseLoader` is not useful when we want to use `include`, so I think this is the right thing to do.

I've moved the instantiation to be out of the conditional though, just to be clearer.

yuvipanda:
@consideRatio I have not changed the contents of the templates, so there should be no visual differences. Just extracted and autoformatted.

consideRatio (approved):
I think this LGTM. @manics what do you think?

---

@yuvipanda can you update the PR description to mention that you add the new config option, and maybe consider if the title can highlight what is new for the jupyterhub admin reading this changelog? It sounds like a bugfix/maintenance PR in a way as it mentions "robustly".

yuvipanda:
@consideRatio how about now for the PR title & description?

manics:
I haven't reviewed the latest code but I'm happy to go with @consideRatio 

Are we happy that the wheel package correctly contains these extra files (from `python -mbuild .`), or can we rely on pyproject.toml/the build system to just "do the right thing"?

yuvipanda:
@manics turns out hatch doesn't read MANIFEST.in, so I've included it separately in the wheel now. Good catch!

consideRatio:
Nice work @yuvipanda !!!

yuvipanda:
THANKS A LOT, @consideRatio!

PR #726, opened 2023-04-28
Require jupyterhub 4+, currently latest kubernetes_asyncio, and stop testing k8s 1.23

### LINKED ISSUE #725: ci: test failures in our oldest dependencies test
I don't get why yet, but [we have test failures in the main branch](https://github.com/jupyterhub/kubespawner/actions/workflows/test.yaml?query=branch%3Amain).

### Maybe related

- https://github.com/jupyterhub/binderhub/issues/1681

### LINKED ISSUE #727: Require JupyterHub 3+, stop tests with k3s 1.223, and release 6.0.0?
We have released kubespawner 5.0.0 just now, but our tests testing our lower bounds of dependencies has started failing reliably for jupyterhub 1.5.1, even though they passed before, and I just can't figure out what changed.

I'd like to drop support for jupyterhub 1 and 2 for KubeSpawner, and go for `jupyterhub>=3.1.1` going onwards to not need to look into what is going wrong.

I'd also like to drop testing of k8s 1.23. We have an action to setup k3s clusters that would benefit from dropping support for testing against k8s 1.23 to reduce the complexity.

### Related

- #725 
- https://github.com/jupyterhub/action-k3s-helm/issues/94

consideRatio:
I repeatedly failed to figure out why our test suite started failing in our oldest dependencies test using JupyterHub 1 in #725. I spend a lot of time looking into this as it was also a clue for test failures in binderhub that probably was unrelated it turns out. With this background of investing significant time into maintenance for backward compatibility, I suggested in #727 we drop support for the previous major JupyterHub versions in KubeSpawner 6.0.0.

In this PR I'm declaring it a requirement to use `jupyterhub>=4` directly even though `jupyterhub>=2` practically works to my knowledge. I motivate going straight to `>=4` to simplify future maintenance and reduce the amounts of tests we need to run in our test suite. My assumption is that there is little value to support jupyterhub 2, 3 and 4 in kubespawner 6+ released after jupyterhub 4, because the z2jh distribution and other users of kubespawner would upgrade to jupyterhub 4 at the same time they would update to a new version of kubespawner.

---

In the spirit of reducing some maintenance load, I'd like to also stop testing against k8s 1.23 and let KubeSpawner 6 come without guarantees it will work, even though it currently is known to work.

I suggest KubeSpawner 6 and z2jh 3 both ship with a disclaimer of "k8s 1.23 is tested to work, but going onwards we won't test against k8s 1.23".

Note that dropping testing against k8s 1.23 is motivated by wanting to make a new breaking release of action-k3s-helm that drops support for k8s 1.23 testing which is hard to maintain as well, see https://github.com/jupyterhub/action-k3s-helm/issues/94.

---

- Closes #725 
- Closes #727

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -39,12 +39,15 @@ jobs:
         #
         include:
           # Tests with oldest supported Python, jupyterhub, k8s, and k8s client
+          #
+          # NOTE: If lower bounds are updated, also update our test for the
+          #       lower bounds in pyproject.toml.
+          #
           - python: "3.7"
             k3s: v1.23
             test_dependencies: >-
-              jupyterhub==1.3.0
-              kubernetes_asyncio==23.6.0
-              traitlets==4.3.2
+              jupyterhub==4.0.0
+              kubernetes_asyncio==24.2.3
 
           # Test with modern python and k8s versions
           - python: "3.9"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,13 +9,13 @@ on:
       - "docs/**"
       - "**.md"
       - ".github/workflows/*"
-      - "!.github/workflows/test-chart.yaml"
+      - "!.github/workflows/test.yaml"
   push:
     paths-ignore:
       - "docs/**"
       - "**.md"
       - ".github/workflows/*"
-      - "!.github/workflows/test-chart.yaml"
+      - "!.github/workflows/test.yaml"
     branches-ignore:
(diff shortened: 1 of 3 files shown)


yuvipanda:
> Note that dropping testing against k8s 1.23 is motivated by wanting to make a new breaking release of action-k3s-helm that drops support for k8s 1.23 testing which is hard to maintain as well, see [jupyterhub/action-k3s-helm#94](https://github.com/jupyterhub/action-k3s-helm/issues/94).

👍🏾

PR #729, opened 2023-05-02
Select profile if any of its choices are interacted with

### LINKED ISSUE #696: Interacting with a profile_list's nested server options should select the server
@jbusecke reported this experience, and I think it makes for a good improvement of KubeSpawner's `profile_list` `profile_options` feature.

>Currently I can select an image on e.g. the medium server, but the default radio button selection will still be on a small server. 
>
>I think it would make a lot of sense to automatically select whichever instance the drop down is attached to, once the drop down is selected?

Each entry in the `profile_list` can have additional options as configured under the `profile_list` list entry's `profile_options`. Interaction with such options should switch the radio button to select the `profile_list` entry having those `profile_options`.

profile-list-option

batpad:
This builds on top of this PR: https://github.com/jupyterhub/kubespawner/pull/724

Fixes #696 

Does:
 - Adds `js-` prefixed classes to the Select boxes and the Label containers - this is not strictly necessary, but I find it good practice to separate out classes used by Javascript
 - Adds a `<script>` tag to the end of the template with simple jQuery to react to changes to the select tag and ensure that the parent Input Radio is selected. If there's a strong preference, can move to a separate file right away, or can do that if and when the JS gets more complex.

Stylistic choices:

 - I've gone ahead and used jQuery since it existed - happy to remove the dependency, but it would mean a bit more verbose code.
 - There's probably a few different ways to handle traversing the DOM and selecting the correct radio, probably with different trade-offs. I've gone for the simplest approach here: this does mean that this function would need to be looked at in case the HTML structure of the page drastically changes, but hopefully it's simple enough to tell what's going on here that it would be easy to fix.

Let me know if this works / if there's anything else that would be good to do here.

cc @yuvipanda

### DIFF
--- a/kubespawner/templates/form.html
+++ b/kubespawner/templates/form.html
@@ -17,11 +17,10 @@
           <div>
             {%- for k, option in profile.profile_options.items() %}
               <div class="option">
-                <label for="profile-option-{{ profile.slug }}-{{ k }}" class="js-profile-option-label">
+                <span class="profile-options-label">
                   {{ option.display_name }}
-                </label>
-                <select name="profile-option-{{ profile.slug }}-{{ k }}"
-                        class="form-control js-profile-option-select">
+                </span>
+                <select name="profile-option-{{ profile.slug }}-{{ k }}">
                   {%- for k, choice in option['choices'].items() %}
                     <option value="{{ k }}" {% if choice.default %}selected{% endif %}>
                       {{ choice.display_name }}
@@ -37,9 +36,8 @@
   {%- endfor %}
 </div>
 <script>
-  $('.js-profile-option-select, .js-profile-option-label').click(function() {
-    // if the user clicks on either the select or label for the select,
-    // make sure to select the associated profile radio.
+  $('.js-profile-option-select').click(function() {
+    // we need this bit of JS to select the profile when a <select> inside is clicked.
     $(this).parents('.js-profile-label')
       .find('input[type=radio]')
       .prop('checked', true);
--- a/kubespawner/templates/form.html
+++ b/kubespawner/templates/form.html
@@ -17,7 +17,7 @@
           <div>
             {%- for k, option in profile.profile_options.items() %}
               <div class="option">
-                <label for="profile-option-{{ profile.slug }}-{{ k }}">{{ option.display_name }}</label>
+                <label for="profile-option-{{ profile.slug }}-{{ k }}" class="js-profile-option-label">{{ option.display_name }}</label>
                 <select name="profile-option-{{ profile.slug }}-{{ k }}" class="js-profile-option-select form-control">
                   {%- for k, choice in option['choices'].items() %}
                     <option value="{{ k }}" {% if choice.default %}selected{% endif %}>{{ choice.display_name }}</option>
@@ -32,7 +32,7 @@
   {%- endfor %}
 </div>
 <script>
-  $('.js-profile-option-select').click(function() {
+  $('.js-profile-option-select, .js-profile-option-label').click(function() {
     // if the user changes a select box of a profile option,
     // make sure we also select the radio button for that option
     $(this).parents('.js-profile-label')
--- a/kubespawner/templates/form.html
+++ b/kubespawner/templates/form.html
@@ -2,7 +2,8 @@
 <div class="form-group" id="kubespawner-profiles-list">
   {%- for profile in profile_list %}
     {#- Wrap everything in a label tag so clicking anywhere selects the option #}
-    <label for="profile-item-{{ profile.slug }}" class="profile js-profile-label">
+    <label for="profile-item-{{ profile.slug }}"
+           class="profile js-profile-label">
       <div class="radio">
         <input type="radio"
                name="profile"
@@ -17,14 +18,10 @@
           <div>
             {%- for k, option in profile.profile_options.items() %}
               <div class="option">
-                <span class="profile-options-label">
-                  {{ option.display_name }}
-                </span>
+                <span class="profile-options-label">{{ option.display_name }}</span>
                 <select name="profile-option-{{ profile.slug }}-{{ k }}">
                   {%- for k, choice in option['choices'].items() %}
-                    <option value="{{ k }}" {% if choice.default %}selected{% endif %}>
-                      {{ choice.display_name }}
-                    </option>
+                    <option value="{{ k }}" {% if choice.default %}selected{% endif %}>{{ choice.display_name }}</option>
                   {%- endfor %}
                 </select>
               </div>
--- a/kubespawner/templates/form.html
+++ b/kubespawner/templates/form.html
@@ -2,7 +2,7 @@
 <div class="form-group" id="kubespawner-profiles-list">
   {%- for profile in profile_list %}
     {#- Wrap everything in a label tag so clicking anywhere selects the option #}
-    <label for="profile-item-{{ profile.slug }}" class="profile">
+    <label for="profile-item-{{ profile.slug }}" class="profile js-profile-label">
       <div class="radio">
         <input type="radio"
                name="profile"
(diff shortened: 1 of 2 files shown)


yuvipanda:
Hmm, I was trying to see why pre-commit.ci couldn't auto-fix https://results.pre-commit.ci/run/github/46604988/1683023436.r2OoP9CCQH6uz9H7a_tv_g.

image

Wonder what is the cause here.

consideRatio:
@yuvipanda pre-commit.ci and we as maintainers doesn't have rights to push commits to this branch and that is whats caused the failure in pre-commit.ci.

I figure I'll go for a merge and fix this retroactively to expedite this.

consideRatio:
Wieee thank you @batpad for working this!!

This is an improvement, but doesn't fix the issue fully - this only works when we make a change to something as compared to selecting something.

change-only

Do you have ideas on how this can be solved?

consideRatio:
@batpad looking into this a bit, I think the issue is that we define nested <label> tags, which doesn't seem to be something thats considered okay to do, see https://stackoverflow.com/a/20842000/2220152.

I don't think we need the <label> tag on `Image` / `Memory` / `CPU` above. Maybe its sufficient with `<span>` or `<p>` or similar?

batpad:
@consideRatio thanks for looking at this!

Ah, I see, so if you click on the option but dont actually "change" it in the Select, it doesn't select the Profile, when it probably should.

This is because we're currently binding to the "change" event on the `select`, and since it does not change, the event does not get fired.

Do you think it's okay to trigger the selection of the profile just when the user clicks the Select box? i.e. don't wait for Select to open and click on an Option. (I'll submit a quick commit so we can see what this looks like)

We can try this with binding to the `click` event on the Select - if that seems awkward, let me see if there's any way to have an event on clicking an `option`, even if the option does not change the selection.

batpad:
@consideRatio I pushed a commit to change the profile selection to happen on the click event. So now, the profile should be selected when you click on the Select box, regardless if you change the option or not. Thanks for that catch.

Interestingly, the behaviour for me is slightly different on Chrome and Firefox.

On Firefox: Profile Radio gets selected AFTER clicking on the option.

On Chrome: Profile Radio gets selected immediately when clicking on the Select.

Let me know if it helps to attach a video of the behaviour - although I think this fulfills the purpose of just making sure the Profile radio gets selected whenever someone clicks on the profile or interacts with it.

batpad:
> I also think pressing the dropdown labels should work

Ah, another really good catch :-) - lemme get to that. Yes, ideally this can be fixed with the `for` attribute, otherwise will add another click handler for the label or so.

kubespawner/templates/form.html
```diff
@@ -18,7 +18,7 @@ <h3>{{ profile.display_name }}</h3>
             {%- for k, option in profile.profile_options.items() %}
               <div class="option">
                 <label for="profile-option-{{ profile.slug }}-{{ k }}">{{ option.display_name }}</label>
-                <select name="profile-option-{{ profile.slug }}-{{ k }}" class="form-control">
+                <select name="profile-option-{{ profile.slug }}-{{ k }}" class="js-profile-option-select form-control">
```
consideRatio, line 21: Line 1: Update `for` attribute of label to point to the radio button rather than the select item
Line 2-3: Autoformatting to split lines, and consistently putting js-x classes last.

```suggestion
                <label for="profile-item-{{ profile.slug }}">{{ option.display_name }}</label>
                <select name="profile-option-{{ profile.slug }}-{{ k }}"
                        class="form-control js-profile-option-select">
```

batpad:
@consideRatio hmm I pushed a commit to add the click handler to the `label` - this "works" and I guess is okay, but just trying to take a step back on the problem a bit --

Also reading the StackOverflow issue you linked. Right, so the fundamental problem here was that we were trying to make the click work just using the `label` and `for` attributes. This will not work for nested form elements - it seems like a nice way to avoid some Javascript :-) - but I think what we might actually want to do here:

 - Change the structure to not wrap the entire element in a `label` - just have a `div` enclosing each profile element, and remove this nested `label` structure, also because it results in invalid HTML.
 - Just attach a single click handler to the parent div element to select the input radio inside.

This seems to me like it would be a bit cleaner - result in cleaner markup, and if we need to have some JS anyways to make this work, it seems cleaner to have a single click event on the container.

Let me know if this makes sense and this should actually be pretty quick to do.

kubespawner/templates/form.html
batpad: @consideRatio did editing the `for` attribute work for you? It didn't seem to work for me when I tried, but I can try again / maybe I messed something up.

consideRatio: Yepp, this is making the `for` change from the main branch, and testing on chrome of the latest/very recent version from ubuntu 22.04

change

consideRatio:
This is fine as it is, if you want you can go for the `for` change and reduce what's done by the JS but I'm not sure I think that is better as the label really is for the select box - its just that we change it to hack the selection logic when using nested labels which we shouldn't.

So, I suggest we stay with what we have here.

But! Autoformatting - run do `pip install pre-commit` and `pre-commit run -a`, then commit those changes. You can also do `pre-commit install --install-hooks` to get autoformatting changes made before you make commits going onwards.

batpad:
> I don't think we need the tag on Image / Memory / CPU above. Maybe its sufficient with <span> or <p> or similar?

So I tried this with a `span` and that works so that the input gets selected when clicking the label.

Since label inside label is invalid according to the spec, I can just make that change, and maybe this is okay to go? I think it might be better to stay away from larger restructuring of the HTML?

consideRatio:
Btw I don't think we need to fix the nested label tag issue, it seems to work after all - so if it comes to retaining the styling vs stop nesting labels - then I favor retaining the styling as it was.

batpad:
> Btw I don't think we need to fix the nested label tag issue, it seems to work after all - so if it comes to retaining the styling vs stop nesting labels - then I favor retaining the styling as it was.

Haa, apologies for not catching the styling regression in Chrome - so strange that the styling is fairly different in Firefox and now it's driving me a bit mad trying to figure out what's happening exactly :-)

And apologies that what should have been a simple fix has spiraled a bit, but thanks again for your feedback and watchful eye. I'm going to play around a bit with the styling mostly to satisfy my own curiosity on what's happening here, but I think you're right: it's better to revert to the nested labels, since that all seemed to work. I do prefer adding the explicit click handler rather than adding the `for` for the parent radio for two reasons:

 - Am pretty sure am seeing inconsistent behaviour when clicking on it between browsers, and since this is undefined in the spec, I'd rather not depend on non-standard browser behaviour for that to work.
 - It does seem like a misuse of the `label` tag and seems like it would really mess with things like screen readers, with the label claiming to be a label for something it's not really the label for.

I have to say that there's a part of me that still really does not like the nested labels, and inconsistencies I'm seeing between chrome and firefox are worrying me a bit. But I think we can take that up separately and focus this PR on just fixing the issue it set out to fix, which is simply selecting the profile when the "Select" is interacted with.

Will revert to nested labels with the click handler on the label and push in a bit.

Thanks again @consideRatio !

yuvipanda:
Thanks a lot for engaging on this issue, @consideRatio!

batpad:
Reverted to using the `label` tag instead of `span` and adding the click handler.

@consideRatio turns out the styling was totally an Oops on my part: I accidentally removed the `form-control` class on the \<select\>. So we could do either `label` or `span` here - switching to `span` will allow us to remove the additional click handler.

Have left it at `label` for now based on our last discussion. If this looks okay to you, I'd be fine with merging this and happy to chat about the nested label stuff separately if it's something we think is worth changing down the line.

batpad:
> so strange that the styling is fairly different in Firefox and now it's driving me a bit mad trying to figure out what's happening exactly

Just to clarify - this was just my brain on too little coffee or so - I accidentally removed a class and then my brain got confused - have tested between FF and Chrome again and things look consistent / fine. Sorry about that, and thanks again for catching.

consideRatio (approved):
Wieee this looks exactly like before testing on chrome, and works exactly as expected to fix #696.

I opened #734 about avoiding nested `<label>` tags as it seems you have a quick followup fix for that without introducing a styling change! Nice!

Thank you for your thorough work on this @batpad!! :heart: :tada: :sunflower:

Issue #699, opened 2023-02-07
Allow for profile_options to have more than `choices` - a text based input field


consideRatio:
In [`profile_list`](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list) documentation, we write about the `profile_options` section introduced in #607.

Could we make this support a free form input, allowing users to specify something arbitrary? That could be used for input to choosing an image for example.

yuvipanda:
yep i think this should be doable!

yuvipanda:
Another option is to say we support an 'other', which will add an 'Other' entry to the end of the dropdown, and selecting that will have a textbox show up.

jbusecke:
Either one of those would be great options IMO. Thanks for working on this.

batpad:
@yuvipanda I can take a stab at this.

We loop through the profiles here: https://github.com/yuvipanda/kubespawner/blob/profile_template/kubespawner/templates/form.html#L16 - I'm guessing there is some sort of `profile_option` that would define whether the Select for that profile needs an "Other" option or not? Could you help me with what that: basically, how do we know whether to render an extra "Other" option in the Select drop-down for a particular Profile.

This should then ideally be pretty quick work to hook up.

PR #735, opened 2023-05-18
Allow end user to select a choice different from list of available choices in profile_list

batpad:
Opening a Draft PR working toward #699 .

This should eventually:

 - Allow specifying an option on profiles to allow a users to specify free-form text for an image (currently a single boolean called `allow_other`, but I think @yuvipanda had some ideas to make that a bit nicer.
 - If `allow_other` is true, show an additional option called "Other" in the dropdown and handle showing a text input when Other is selected
 - Handle the text input value on the backend to actually pull the image onto the cluster, etc.

Currently, have added a simple `allow_other` option and handling it just on the frontend.

@yuvipanda hopefully this is a good starting point for us to work together on implementing handling the value on the backend and improve the semantics of the options.

cc @consideRatio

### DIFF
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -46,10 +46,10 @@ c.KubeSpawner.profile_list = [
         'profile_options': {
             'image': {
                 'display_name': 'Image',
-                'free_form': {
+                'other_choice': {
                     'enabled': True,
                     'display_name': 'Image Location',
-                    'match_regex': '^pangeo/.*$',
+                    'validation_match_regex': '^pangeo/.*$',
                     'validation_message': 'Must be a pangeo image, matching ^pangeo/.*$',
                     'kubespawner_override': {'image': '{value}'},
                 },
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -49,7 +49,7 @@ c.KubeSpawner.profile_list = [
                 'other_choice': {
                     'enabled': True,
                     'display_name': 'Image Location',
-                    'validation_match_regex': '^pangeo/.*$',
+                    'validation_regex': '^pangeo/.*$',
                     'validation_message': 'Must be a pangeo image, matching ^pangeo/.*$',
                     'kubespawner_override': {'image': '{value}'},
                 },
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -46,7 +46,13 @@ c.KubeSpawner.profile_list = [
         'profile_options': {
             'image': {
                 'display_name': 'Image',
-                'allow_other': True,
+                'free_form': {
+                    'enabled': True,
+                    'display_name': 'Image Location',
+                    'match_regex': '^pangeo/.*$',
+                    'validation_message': 'Must be a pangeo image, matching ^pangeo/.*$',
+                    'kubespawner_override': {'image': '{value}'},
+                },
                 'choices': {
                     'pytorch': {
                         'display_name': 'Python 3 Training Notebook',
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -46,6 +46,7 @@ c.KubeSpawner.profile_list = [
         'profile_options': {
             'image': {
                 'display_name': 'Image',
+                'allow_other': True,
                 'choices': {
                     'pytorch': {
                         'display_name': 'Python 3 Training Notebook',
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -46,7 +46,7 @@ c.KubeSpawner.profile_list = [
         'profile_options': {
             'image': {
                 'display_name': 'Image',
-                'other_choice': {
+                'unlisted_choice': {
                     'enabled': True,
                     'display_name': 'Image Location',
                     'validation_regex': '^pangeo/.*$',
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2926,10 +2926,9 @@ class KubeSpawner(Spawner):
     def _env_keep_default(self):
         return []
 
-    _profile_list = None
 
     def _render_options_form(self, profile_list):
-        self._profile_list = self._populate_profile_list_defaults(profile_list)
+        profile_list = self._populate_profile_list_defaults(profile_list)
 
         loader = ChoiceLoader(
             [
@@ -2945,7 +2944,7 @@ class KubeSpawner(Spawner):
             profile_form_template = env.from_string(self.profile_form_template)
         else:
             profile_form_template = env.get_template("form.html")
-        return profile_form_template.render(profile_list=self._profile_list)
+        return profile_form_template.render(profile_list=profile_list)
 
     async def _render_options_form_dynamically(self, current_spawner):
         profile_list = await maybe_future(self.profile_list(current_spawner))
@@ -3218,13 +3217,12 @@ class KubeSpawner(Spawner):
         Override in subclasses to support other options.
         """
 
-        if self._profile_list is None:
-            if callable(self.profile_list):
-                profile_list = await maybe_future(self.profile_list(self))
-            else:
-                profile_list = self.profile_list
+        if callable(self.profile_list):
+            profile_list = await maybe_future(self.profile_list(self))
+        else:
+            profile_list = self.profile_list
 
-            self._profile_list = self._populate_profile_list_defaults(profile_list)
+        profile_list = self._populate_profile_list_defaults(profile_list)
 
         selected_profile = self.user_options.get('profile', None)
         selected_profile_user_options = dict(self.user_options)
@@ -3232,7 +3230,7 @@ class KubeSpawner(Spawner):
             # Remove the 'profile' key so we are left with only selected profile options
             del selected_profile_user_options['profile']
 
-        if self._profile_list:
+        if profile_list:
             await self._load_profile(selected_profile, selected_profile_user_options)
         elif selected_profile:
             self.log.warning(
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3033,9 +3033,7 @@ class KubeSpawner(Spawner):
                         raise ValueError(
                             f'Expected option {option_name} for profile {profile["slug"]} or {unlisted_choice_form_key}, not found in posted form'
                         )
-                    unlisted_choice = selected_options[
-                        unlisted_choice_form_key
-                    ]
+                    unlisted_choice = selected_options[unlisted_choice_form_key]
 
                     # Validate value of 'unlisted_choice' against validation regex
                     if profile.get('profile_options')[option_name][
@@ -3055,6 +3053,7 @@ class KubeSpawner(Spawner):
                     raise ValueError(
                         f'Expected option {option_name} for profile {profile["slug"]}, not found in posted form'
                     )
+
     async def _load_profile(self, slug, selected_profile_user_options):
         """Load a profile by name
 
@@ -3105,7 +3104,9 @@ class KubeSpawner(Spawner):
                 setattr(self, k, v)
 
         if profile.get('profile_options'):
-            self._validate_posted_profile_options(profile, selected_profile_user_options)
+            self._validate_posted_profile_options(
+                profile, selected_profile_user_options
+            )
             # Get selected options or default to the first option if none is passed
             for option_name, option in profile.get('profile_options').items():
                 unlisted_choice_form_key = f'{option_name}--unlisted-choice'
@@ -3129,7 +3130,9 @@ class KubeSpawner(Spawner):
                     ]
                     for k, v in chosen_option_overrides.items():
                         chosen_option_overrides[k] = v.format(
-                            value=selected_profile_user_options[unlisted_choice_form_key]
+                            value=selected_profile_user_options[
+                                unlisted_choice_form_key
+                            ]
                         )
                 else:
                     chosen_option_overrides = option['choices'][chosen_option][
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2963,8 +2963,11 @@ class KubeSpawner(Spawner):
         if not self.profile_list:
             return ''
         if callable(self.profile_list):
+            # Return the function dynamically, so JupyterHub will call this when the
+            # form needs rendering
             return self._render_options_form_dynamically
         else:
+            # Return the rendered string, as it does not change
             return self._render_options_form(self.profile_list)
 
     @default('options_from_form')
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3054,31 +3054,38 @@ class KubeSpawner(Spawner):
                         f'Expected option {option_name} for profile {profile["slug"]}, not found in posted form'
                     )
 
+    def _get_profile(self, slug: str):
+        """
+        Get the configured profile for given profile slug
+
+        Raises an error if no profile exists for the given slug.
+
+        If slug is empty string, return the default profile
+        """
+        profile_list = self._populate_profile_list_defaults(self.profile_list)
+        if slug != "":
+            for profile in profile_list:
+                if profile['slug'] == slug:
+                    return profile
+
+            # A slug is specified, but not found
+            # name specified, but not found
+            raise ValueError(
+                "No such profile: %s. Options include: %s"
+                % (slug, ', '.join(p['slug'] for p in profile_list))
+            )
+        else:
+            # slug is not specified, let's find the default and return it
+            # default is guaranteed to be set in at least one profile
+            return next(p for p in profile_list if p.get('default'))
+
     async def _load_profile(self, slug, selected_profile_user_options):
         """Load a profile by name
 
         Called by load_user_options
         """
-
         # find the profile
-        default_profile = self._profile_list[0]
-        for profile in self._profile_list:
-            if profile.get('default', False):
-                # explicit default, not the first
-                default_profile = profile
-
-            if profile['slug'] == slug:
-                break
-        else:
-            if slug:
-                # name specified, but not found
-                raise ValueError(
-                    "No such profile: %s. Options include: %s"
-                    % (slug, ', '.join(p['slug'] for p in self._profile_list))
-                )
-            else:
-                # no name specified, use the default
-                profile = default_profile
+        profile = self._get_profile(slug)
 
         self.log.debug(
             "Applying KubeSpawner override for profile '%s'", profile['display_name']
@@ -3173,6 +3180,8 @@ class KubeSpawner(Spawner):
           from display_name
         - If profile_options are present with choices, but no choice is set
           as the default, set the first choice to be the default.
+        - If no default profile is set, the first profile is set to be the
+          default
 
         The profile_list passed in is mutated and returned.
         """
@@ -3190,6 +3199,11 @@ class KubeSpawner(Spawner):
                         # No explicit default is set
                         default_choice = list(option_config['choices'].keys())[0]
                         option_config['choices'][default_choice]["default"] = True
+
+
+        if not any(p.get("default") for p in profile_list):
+            # No profile has 'default' explicitly set, we set it for the first profile in the List
+            profile_list[0]["default"] = True
 
         return profile_list
 
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3054,15 +3054,15 @@ class KubeSpawner(Spawner):
                         f'Expected option {option_name} for profile {profile["slug"]}, not found in posted form'
                     )
 
-    def _get_profile(self, slug: str):
+    def _get_profile(self, slug: str, profile_list: list):
         """
         Get the configured profile for given profile slug
 
         Raises an error if no profile exists for the given slug.
 
         If slug is empty string, return the default profile
+        profile_list should already have all its defaults set.
         """
-        profile_list = self._populate_profile_list_defaults(self.profile_list)
         if slug != "":
             for profile in profile_list:
                 if profile['slug'] == slug:
@@ -3106,13 +3106,12 @@ class KubeSpawner(Spawner):
             else:
                 setattr(self, k, v)
 
-    async def _load_profile(self, slug, selected_profile_user_options):
+    async def _load_profile(self, slug, profile_list, selected_profile_user_options):
         """Load a profile by name
 
         Called by load_user_options
         """
-        # find the profile
-        profile = self._get_profile(slug)
+        profile = self._get_profile(slug, profile_list)
 
         self.log.debug(
             "Applying KubeSpawner override for profile '%s'", profile['display_name']
@@ -3230,7 +3229,9 @@ class KubeSpawner(Spawner):
             del selected_profile_user_options['profile']
 
         if profile_list:
-            await self._load_profile(selected_profile, selected_profile_user_options)
+            await self._load_profile(
+                selected_profile, profile_list, selected_profile_user_options
+            )
         elif selected_profile:
             self.log.warning(
                 "Profile %r requested, but profiles are not enabled", selected_profile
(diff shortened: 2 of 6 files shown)


consideRatio:
Ooooh great to see you work towards this @batpad!!!

### Reflection on config naming

I think `allow_other` could be renamed to be a bit more self-explanatory, as something related to the `choices` config.

With a free-form input field, front-end validation of the input is probably going to end up being another feature. I don't think we need to think about implementing that as part of this PR, but I'm considering it already when thinking on what configuration to provide. For example `include_other_choice` as a boolean may need to be augmented with `other_choice_validation` or similar, which makes me consider if we want `other_choice` as a dictionary with `included` under it or similar, allowing `other_choice.included` and `other_choice.some_validation_related_config`.

Hmmm hmmm, I'm currently leaning towards a flat structure to configure this, like `other_choice_included`, to be complemented with `other_choice_some_validation_related_config`. There are pros and cons of using a flat config structure (`a_b`, `a_c`) vs a nested structure (`a["b"]`, `a["c"]`), but I think staying flat is fine here but its good if we can find a common prefix in the name such as `other_choice_`.

Wdyt about renaming `allow_other` (boolean) to `other_choice_included` (boolean)?

batpad:
@consideRatio I think @yuvipanda had a good idea for how this config should be structured, making it an object, that also contained a regex to use to validate the user input. Let's wait for him to weigh in on the structure for options - sorry, he had pasted that to me somewhere, but I can't find it :-/ .

batpad:
> There are pros and cons of using a flat config structure (a_b, a_c) vs a nested structure (a["b"], a["c"]), but I think staying flat is fine here but its good if we can find a common prefix in the name such as other_choice_.

Hmm, this is an interesting one - yea, I do think this should allow for multiple "sub-options" like the regex to validate input, so either prefixing with `other_choice_` or making it a nested structure sounds good to me - I don't really have a preference on nested vs flat - happy to go with whatever @yuvipanda and you feel is best here.

yuvipanda:
Thanks for your comments, @consideRatio!

I left a comment in https://github.com/batpad/kubespawner/commit/cc0037b9753d6905cd1d98a0a2acaded905fd84d#r111675499 earlier, which I will repro here.

Let's expand this a little bit. How about something like:

```yaml
freeform:
   enabled: true
   match_regex: (optional regex to match)
   kubespawner_override:
       some_config_key: some_value-with-{value}-substituted-with-what-user-wrote
```

So for an image where we only want to allow pangeo images, it would be something like

```yaml
freeform:
   enabled: true
   match_regex: ^pangeo/.*$
   kubespawner_override:
      image: "{value}"
```

https://github.com/jupyterhub/kubespawner/blob/ff8dcf15205208fb25a7255563b59f53f07aca88/kubespawner/spawner.py#L3016 is what gets the form data, converts them into 'user options' that are then passed on to https://github.com/jupyterhub/kubespawner/blob/ff8dcf15205208fb25a7255563b59f53f07aca88/kubespawner/spawner.py#L3091, which interprets them and applies the overrides.

yuvipanda:
With respect to nested vs flat, @consideRatio, I think nested is better here! Because we already will have three things:

1. Wether to enable this or not
2. How it should actually override the kubespawner traitlet config. Currently, with choices, we set a 'key' and override it with with specific config that is set in the `jupyterhub_config.py`. With free-form text, we should allow admins the option to interpolate the value entered into an override. We also need to *store* the raw user entered value in user_options, and re-setup the kubespawner_override whenever the user server is stopped / started again.
3. Simple regex based frontend validation, that @batpad tells me can be done with HTML5 form validation built into browsers, no extras needed.

Given this, I'd suggest the nested config structure.

batpad:
To add:

 - We would also want something like a `validation_message` to show a human-readable error message / describe the input format required, like the human readable value of the `regex` field.

yuvipanda:
Re: nested config, I also opened https://github.com/jupyterhub/kubespawner/issues/736 to help us mitigate the complexity of validation in the nested config in the future. Does not need to block this PR though. I would like us to make sure we get frontend validation in here with this though.

jupyterhub_config.py
```diff
@@ -46,6 +46,13 @@
         'profile_options': {
             'image': {
                 'display_name': 'Image',
+                'free_form': {
```
consideRatio, line 49: @yuvipanda @batpad are you okay naming this `other_choice`?

I think the "other choice" name provides intuition that:
- its one additional choice to `choices`
- of a special "free input" kind
- that its just one single choice addition

I don't think naming this `free_form` would provide as much intuition on what the configuration is about compared to `other_choice`. And if not, it would making this config harder to understand intuitively, which could make it harder to and setup and update over time.

batpad: > @yuvipanda @batpad are you okay naming this other_choice?

`other_choice` sounds great to me :-)

Will wait a bit to see if @yuvipanda has any strong opinions, but I can go ahead and make that change.

Do the other variable names inside the object sound okay? Maybe be nice to make all naming changes in one go.

Thanks again for the thoughtfulness around this.

consideRatio: - `enabled` :+1:
- `display_name` :+1: - consistent with other naming
- `match_regex` :thinking: --- works, but I can also see value in having `validation_` be involved in this naming, if it is `validation_match_regex` or `validation_regex`
- `validation_message` :+1: 
- `kubespawner_override` :+1: - consistent with other naming

batpad: Nice!

@consideRatio I have made a commit renaming `free_form` to `other_choice` and `match_regex` to `validation_match_regex`.

> match_regex 🤔 --- works, but I can also see value in having validation_ be involved in this naming, if it is validation_match_regex or validation_regex. I think I favor validation_match_regex.

I definitely like prefixing with `validation` since we also have `validation_message` and also STRONGLY agree that `validation` should not be further nested. Personally, I think I'd prefer the slightly less verbose `validation_regex` to `validation_match_regex` as `validation_regex` sounds pretty unambiguous to me. But also not a hard opinion, and I'm not as familiar with general naming conventions here, so went with your suggestion here. If you do agree / feel that `validation_regex` is fine, am happy to change that.

Thank you again.

batpad: Ah, forgot to mention - also changed the `name` for the input field that's submitted with the POST request from `profile-option-{{ profile.slug }}-{{ k }}-other` to `profile-option-{{ profile.slug }}-{{ k }}-other-choice` to keep that consistent with the name of the option.

@consideRatio would be great to get your 👀 and 👍 / 👎 on the naming of this as well, since that's the interface between the front and backend and it would be nice to be thoughtful about what we call this.

ranchodeluxe:
Hey folks, just checking in here. There's a demo on June 26th that Aimee will be doing and I'm wondering if maybe this work could be wrapped up and exposed on `nasa-veda` cluster by then? No worries if it's a stretch. Is there anyway I can help?

batpad:
Starting to look at the backend integration for this.

Just pasting over @yuvipanda's comments from elsewhere about where this needs to happen to have it easy to refer to:

https://github.com/jupyterhub/kubespawner/blob/ff8dcf15205208fb25a7255563b59f53f07aca88/kubespawner/spawner.py#L3016 is what gets the form data, converts them into 'user options' that are then passed on to https://github.com/jupyterhub/kubespawner/blob/ff8dcf15205208fb25a7255563b59f53f07aca88/kubespawner/spawner.py#L3091, which interprets them and applies the overrides.

I'll start familiarizing myself here and then lean on you @yuvipanda to get through the actual implementation.

batpad:
@yuvipanda @consideRatio there's probably some discussion to be had around details of the implementation here, but this is now "complete" and ready for review:

 - Adds an `other_choice` option which handles showing a free text input for that choice
 - Handles validating and overriding the appropriate profile value on the backend
 - Has basic tests for the above, including a test to handle the `choices` parameter which did not have a test

Overall, I think there's some ideas on how we could do this whole thing a bit better, but that might be a bit of a larger refactor, that hopefully we can discuss separately.

I'll give this whole thing another good look tomorrow, but do let me know anything that pops out / that should definitely be fixed.

(thanks again @yuvipanda for helping me figure out most of the backend integration code)

jupyterhub_config.py
consideRatio: > Personally, I think I'd prefer the slightly less verbose `validation_regex` to `validation_match_regex` as `validation_regex` sounds pretty unambiguous to me.

I think you are right that it sounds pretty unambiguous and I couldn't figure out how to misunderstand it!

> If you do agree / feel that `validation_regex` is fine, am happy to change that.

Thank you for thinking about naming things so thoroughly!!! Let's go for it!

consideRatio:
> * Currently, the value for the "other" option in the drop-down is hard-coded as "other". This means you can't have a different option in the drop-down that has a value "other". Hopefully this is fine, but just a heads-up - we can change this hard-coded "other" to something else if this seems like it might be a problem.

I think its the right call to hardcode this specific choice for a given profile, and "other" seems like a fine choice but also one that could risk collide with other choices. I think either we should make the name a bit more obscure, like `other-choice` / `other_choice`, or force the other choices slugs to not overlap with the help of validation logic.

jupyterhub_config.py
batpad: > Thank you for thinking about naming things so thoroughly @batpad!!! Let's go for it!

Made this change :-)

batpad:
>  I think either we should make the name a bit more obscure, like other-choice / other_choice, or force the other choices slugs to not overlap with the help of validation logic.

Speaking to @yuvipanda a bit here -- have gone with naming the form field with `--other-choice` i.e. adding an additional `-` to prevent accidental naming conflicts.

It would also be good to have comprehensive validation of the config via JSON schema or similar, and then we can just prohibit `choices` with `--` in them. We can do that as a separate PR?

batpad:
> I think either we should make the name a bit more obscure, like other-choice / other_choice, or force the other choices slugs to not overlap with the help of validation logic.

Apologies, I mis-spoke in my comment above. Two different things needed to be done:

 - Change the value of the `name` field, affecting the key for the value being POSTed, made the `--other-choice` instead of `-other-choice`
 - Changing the `value` in the drop-down for the `Other` option. Changed that currently to just `other-choice` from `other`.

kubespawner/spawner.py
```diff
@@ -1537,6 +1538,16 @@ def _validate_image_pull_secrets(self, proposal):
...
 
           - `display_name`: Name used to identify this particular option
+          - `other_choice`: Object to specify if there should be a free-form field if the user
+            select "Other" as a choice:
+            - `enabled`: Boolean, whether the free form input should be enabled
+            - `display_name`: String, label for input field
+            - `validation_regex`: Optional, regex that the free form input should match - eg. ^pangeo/.*$
+            - `validation_message`: Optional, validation message for the regex. Should describe the required
+               input format in a human-readable way.
+            - `kubespawner_override`: Object specifying what key:values should be over-ridden
```
yuvipanda, line 1548: Specify that you can use `{value}` here to substitute whatever the user entered value was.

kubespawner/spawner.py
```diff
@@ -3048,14 +3068,39 @@ async def _load_profile(self, slug, selected_profile_user_options):
...
-                    user_selected_option_name
-                    not in profile.get('profile_options').keys()
-                ):
-                    raise ValueError(
-                        f'Expected option {user_selected_option_name} for profile {profile["slug"]}, not found in posted form'
-                    )
+
+            for option_name, option in profile.get('profile_options').items():
+                if option_name not in selected_profile_user_options:
+                    # other_choice in Enabled:
```
yuvipanda, line 3074:
```suggestion
                    # other_choice is enabled
```

kubespawner/spawner.py
```diff
@@ -3048,14 +3068,39 @@ async def _load_profile(self, slug, selected_profile_user_options):
...
+                    if option.get('other_choice', {}).get('enabled', False):
+                        if (
+                            f'{option_name}--other-choice'
```
yuvipanda, line 3077: Given `{option_name}--other-choice` is used multiple times, can we make it into a variable for clarity? `other_choice_form_key` or something?

kubespawner/spawner.py
```diff
@@ -3048,14 +3068,39 @@ async def _load_profile(self, slug, selected_profile_user_options):
...
+                        ]
+
+                        # Validate value of 'other-choice' against validation regex
+                        if profile.get('profile_options')[option_name][
+                            'other_choice'
+                        ].get('validation_regex', False):
+                            other_choice_validation_regex = profile.get(
+                                'profile_options'
+                            )[option_name]['other_choice']['validation_regex']
+                            regex = re.compile(other_choice_validation_regex)
```
yuvipanda, line 3094: You can directly call `re.match` here instead. If you look at the cpython source code, it gets compiled automatically (https://github.com/python/cpython/blob/904aef994262383ae916545908f0578c2d53cf31/Lib/re/__init__.py#L163). And compile also caches it (https://github.com/python/cpython/blob/904aef994262383ae916545908f0578c2d53cf31/Lib/re/__init__.py#L273) so it isn't recompiled unless necessary. Just having a .match here is cleaner than a compile and match.

kubespawner/templates/style.css
```diff
@@ -27,3 +27,7 @@
   margin-right: 8px;
   min-width: 96px;
 }
+
+.hide {
```
yuvipanda, line 31: bootstrap 3, which is used here, has a .hidden that we can use instead https://getbootstrap.com/docs/3.4/css/#helper-classes

tests/test_spawner.py
```diff
@@ -902,6 +958,84 @@ async def test_user_options_set_from_form():
         assert getattr(spawner, key) == value
 
 
+async def test_user_options_set_from_form_choices():
```
yuvipanda, line 961: A docstring describing what this is testing would be great!

tests/test_spawner.py
```diff
@@ -902,6 +958,84 @@ async def test_user_options_set_from_form():
...
+    assert spawner.user_options == {
+        'image': 'pytorch',
+        'profile': _test_profiles[3]['slug'],
+    }
+    assert spawner.cpu_limit is None
+    await spawner.load_user_options()
+    assert getattr(spawner, 'image') == 'pangeo/pytorch-notebook:master'
+
+
+async def test_user_options_set_from_form_other_choice():
```
yuvipanda, line 980: A docstring here would be great too

tests/test_spawner.py
```diff
@@ -902,6 +958,84 @@ async def test_user_options_set_from_form():
...
+    assert spawner.user_options == {
+        'image--other-choice': 'pangeo/test:latest',
+        'profile': _test_profiles[3]['slug'],
+    }
+    assert spawner.cpu_limit is None
+    await spawner.load_user_options()
+    assert getattr(spawner, 'image') == 'pangeo/test:latest'
+
+
+async def test_user_options_set_from_form_invalid_regex():
```
yuvipanda, line 999: Docstring <3

tests/test_spawner.py
```diff
@@ -902,6 +958,84 @@ async def test_user_options_set_from_form():
...
+        'image--other-choice': 'invalid/foo:latest',
+        'profile': _test_profiles[3]['slug'],
+    }
+    assert spawner.cpu_limit is None
+
+    with pytest.raises(ValueError):
+        await spawner.load_user_options()
+
+
+async def test_user_options_set_from_form_no_regex():
```
yuvipanda, line 1019: Thank you for adding all these tests! A docstring would be great here too.

yuvipanda (changes requested):
Some minor feedback, but otherwise this LGTM!

kubespawner/spawner.py
batpad: I agree.

The problem here is that we still need to define the variable in two places, since we have these two different loops. So, I had to create the variable twice: https://github.com/jupyterhub/kubespawner/pull/735/commits/f55e32302a9e80a967c9531fbf10068ca6d7970f#diff-6fdb3a928956e8ab2b810679d51fae72870d45ed4a088e85e16814339aff8901R3074 and https://github.com/jupyterhub/kubespawner/pull/735/commits/f55e32302a9e80a967c9531fbf10068ca6d7970f#diff-6fdb3a928956e8ab2b810679d51fae72870d45ed4a088e85e16814339aff8901R3107 .

This feels really not nice. I think options:

 - Make get_other_choice_key a function or so and call the function from both places so at least we define the `--other-choice` constant in only one place.
 - Refactor to not use two loops and loop only once
 - Leave as is in my latest commit
 - Something else .. ?

Thanks much for the review!

echarles:
How will this be used in the z2jh helm chart? Looking at https://z2jh.jupyter.org/en/stable/resources/reference.html#singleuser-profilelist I don't see a direct/obvious relation/impact, so wanted to ask.

yuvipanda:
@echarles once this is merged, released, and bumped in z2jh, someone would probably add a documentation PR there so it can be used.

jupyterhub_config.py
```diff
@@ -46,6 +46,13 @@
         'profile_options': {
             'image': {
                 'display_name': 'Image',
+                'other_choice': {
```
yuvipanda, line 49: `other_choice` always seemed a bit *too* generic to me, as `other` feels super generic a term. I spent some more time thinking about this, and I think I'd prefer calling it `unlisted_choice`, as it makes it clearer there is a *list* of choices from the admin, and this enables the end user to make a choice that is *unlisted*, but still possibly supported with some validation.

yuvipanda (commented):
This looks good to me, but I'd love for the name of the config to be different. I've made a suggestion, but otherwise this all seems fine.

Thanks for working through this, @batpad

echarles:
> @echarles once this is merged, released, and bumped in z2jh, someone would probably add a documentation PR there so it can be used.

Maeks sense, I have opened https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3152 to follow that.

jupyterhub_config.py
consideRatio: I'm not sure what I think is the best naming yet. I agree there is benefit of understanding that `profile_options` can involve a dropdown list of choices, but it seems a bit out of scope to let this name, which may or may not show up in a config, help convey that we have a list of choices.

I realize I've had a hidden assumption: that this new feature can be used without entries in the `choices` list. I think this is related to the naming discussion because it influences how coupled this choice is to the `choices` list.

The new config in this PR can be used when no entries are in `choices`, it just comes with some redundant UI:

### `choices` with _two_ entries, `other_choice` enabled

current

### `choices` with _no_ entries, `other_choice` enabled

current-with-no-listed-choices

---

@yuvipanda and @batpad here are some decisions I think would be helpful to make before deciding on the naming:

1. If the new config `other_choice`/`unlisted_choice` be allowed when `choices` is empty
2. What UI to go for in those situations
3. If it should say "Other..." or something else by default in the dropdown list when using `other_choice`/`unlisted_choice`
4. If that "Other..." text should be configurable or not

I think (1) it would be great to support use of the new config when `choices` is empty, and (2) that it would be great to not show a single-value dropdown list in the case that the new config is enabled with `choices` empty. I'm not sure about (3) and (4).

yuvipanda: @consideRatio I agree with you on 1 and 2, and should probably be configurable as well (4). I think (Other) is a reasonable default when we do have a list of choices. However, I think none of these should block this PR getting merged, and can be done as follow up PRs.

I think this being usable without any listed choices and the ability to configure the “other” text makes the case stronger for calling it “unlisted_choice” than “other_choice”.

consideRatio:
Hey @batpad I'm going on vacation for three weeks! Below are two documentation related review points, but besides those lean on other maintainers guidance - I'd love to see this approved and merged! Thank you for your work on this @batpad, I think it will be a very appreciated feature!!!

1. Is the new config `help` string documenting the functionality well enough via the configuration reference and reflecting the state to be merged?
   - Reviewable via [PR preview of the built documentation](https://jupyterhub-kubespawner--735.org.readthedocs.build/en/735/spawner.html#kubespawner.KubeSpawner.profile_list)
   - It could be relevant to involve `.. versionchanged:: 6.1` to clarify new functionality was added in the `profile_list` config was added at that point in time.
   - Note that the content of the `help` string is rST formatted, not MyST markdown.
   - Note that documentation can be built and viewed locally with `pip install -r docs/requirements.txt`, `cd docs`, `make devenv`. I'm not sure if it auto-refreshes correctly and re-generates built documentation from the source code, maybe. Otherwise do `rm -rf _build` and run again.
2. Is the PR title and description reflecting the final state to be merged?

kubespawner/spawner.py
```diff
@@ -3048,17 +3069,42 @@ async def _load_profile(self, slug, selected_profile_user_options):
...
-            for user_selected_option_name in selected_profile_user_options.keys():
-                if (
-                    user_selected_option_name
-                    not in profile.get('profile_options').keys()
-                ):
-                    raise ValueError(
-                        f'Expected option {user_selected_option_name} for profile {profile["slug"]}, not found in posted form'
-                    )
+
+            for option_name, option in profile.get('profile_options').items():
```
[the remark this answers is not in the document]
GeorgianaElena: Actually, isn't `selected_profile_user_options` what's passed over by the user in the UI through the form (or though API) and `profile.get('profile_options')` all the options available to them?

If that's the case, then shouldn't we check that the selected options are valid options and not the other way around? Hmm, I'm a bit confused about this now.

cc @yuvipanda

GeorgianaElena (commented):
Thanks a lot for the PR @batpad and thanks for adding tests <3

I left a comment/question about the validation done in the backend, but that's mostly me making sure I understood the implementation. This generally looks good 👍🏼 Thank you ✨

jupyterhub_config.py
GeorgianaElena: If it makes any difference, I think `unlisted_choice` it's more accurate with respect to what it represents.
But I don't think strongly about naming, as long as we have docs that do a good job describing what it does.

kubespawner/spawner.py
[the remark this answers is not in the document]
yuvipanda: @GeorgianaElena By having them in different loops, we don't do any processing at all of the options if *any* of the options provided were invalid. We could split them up into different functions, but I think they should probably just be two loops.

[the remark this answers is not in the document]
yuvipanda: > If that's the case, then shouldn't we check that the selected options are valid options and not the other way around? Hmm, I'm a bit confused about this now.

So the behavior change of this is that options that aren't configured by the admin but sent over will be ignored, while previously they will be errored. Is that right? I do think (and suggested!) the new code, as I think it's much clearer than what was there before. Do you think ignoring extra values while making sure whatever the admin has sent is validated is problematic?

jupyterhub_config.py
yuvipanda: I'd love for us to go with `unlisted_choice`!

yuvipanda:
Unfortunately, @batpad is a little sick, so I'm taking this on and trying to push it forward to completion now.

kubespawner/spawner.py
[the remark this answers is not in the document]
yuvipanda: I've moved the validation to a different function and added a docstring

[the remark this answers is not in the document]
yuvipanda: @GeorgianaElena I've adapted your PR into https://github.com/batpad/kubespawner/pull/2, what do you think?

GeorgianaElena (approved):
I believed this should be merged, after https://github.com/batpad/kubespawner/pull/2 gets merged as well! 🚀

yuvipanda (approved):

yuvipanda:
@GeorgianaElena ok this is ready to go now i think!

After this, I think it'd be useful to do a refactor to move all the profile handling code into a separate file. But I don't want to make this PR too much longer...

GeorgianaElena:
@yuvipanda, I agree! But for now, let's merge this! 🚀 

Many thanks to everyone for all the thought and effort that went into this PR! 🎉

jupyterhub_config.py
yuvipanda: I opened https://github.com/jupyterhub/kubespawner/issues/757 and https://github.com/jupyterhub/kubespawner/issues/756 to track these.

PR #738, opened 2023-05-30
Stop use of nested label tags in HTML

### LINKED ISSUE #734: We have nested label tags in form.html and it works, but its undefined behavior
I understand it as `<label>` tags in HTML aren't meant to be nested in each other, because then the behavior isn't defined so browsers may behave differently in how the present and behave with such content.

Due to that I'm opening this issue to figure out if we can avoid it, and if so, work to avoid it.

consideRatio:
- Closes #734 

There is a slight adjustment in the alignment as part of changing the nested `<label>` tag into a `<span>` tag, but I think that is preferred. In the gif below we see the difference switched back and forth. With this PR it becomes as it is when the label is put lower compared to what it was before, aligning better with the text in the dropdown lists.

alignment

### DIFF
--- a/kubespawner/templates/form.html
+++ b/kubespawner/templates/form.html
@@ -18,8 +18,8 @@
           <div>
             {%- for k, option in profile.profile_options.items() %}
               <div class="option">
-                <label for="profile-option-{{ profile.slug }}-{{ k }}"
-                       class="js-profile-option-label">{{ option.display_name }}</label>
+                <span for="profile-option-{{ profile.slug }}-{{ k }}"
+                      class="js-profile-option-label">{{ option.display_name }}</span>
                 <select name="profile-option-{{ profile.slug }}-{{ k }}"
                         class="form-control js-profile-option-select">
                   {%- for k, choice in option['choices'].items() %}
--- a/kubespawner/templates/style.css
+++ b/kubespawner/templates/style.css
@@ -22,7 +22,7 @@
   padding-bottom: 12px;
 }
 
-#kubespawner-profiles-list .profile .option label {
+#kubespawner-profiles-list .profile .option span {
   font-weight: normal;
   margin-right: 8px;
   min-width: 96px;


minrk:
I don't think this is quite the right solution. We do have two form elements here (the profile and the image), so each should have its label. What I think is incorrect is that the image control is _inside_ the.

Instead, I think the right thing to do is to make sure the profile item label does not enclose the image form element and label.

consideRatio:
@minrk the background is observing that nesting `<label>` tags isn't a well defined HTML it seems.

For example we had a label for each radio button, and when click on a label with an associated radio button, they are supposed to be selected. But, what if the label has a label inside it - then what?

In https://stackoverflow.com/a/20842000/2220152, a person leads with:

> As you know, it is not allowed to nest [label](http://www.w3.org/TR/2013/CR-html5-20130806/forms.html#the-label-element) elements,

And another post sais:

>> The [html5 spec](http://www.w3.org/TR/html-markup/label.html) explicitly says:
>
>The label element must not contain any nested label elements.

I have not found something in the html5 specification or similar saying this explicitly, but I've seen a validator warning about it and these stackoverflow responses.

minrk:
Thanks! I do understand the background, and that there shouldn't be a label inside another. What I mean is that the second label, which is for a real control, shouldn't be inside the first. Not that the second label shouldn't be a label at all. There also [shouldn't be any interactive elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#interactive_content) inside a label. That ought to mean the image select element to which the inner label applies also shouldn't be inside the outer label.

kubespawner/templates/form.html
```diff
@@ -18,8 +18,8 @@ <h3>{{ profile.display_name }}</h3>
           <div>
```
minrk, line 18: I think moving the `</label>` [on L34](https://github.com/jupyterhub/kubespawner/pull/738/files#diff-47559ea38f5c75d2661f1c86c79b031011275298348501fcec8a7d80f23f0dbbR34) to before opening this `<div>` for the properties should be the right fix. Then the label for the profile radio button ends before the labels and controls for the profile options.

consideRatio:
Closing as stale and probably not essential enough to prioritize

PR #742, opened 2023-06-07
return poll status after first load finish

ivyxjc:


kubespawner/spawner.py
```diff
@@ -2185,7 +2185,8 @@ async def poll(self):
         just Falsy, to determine that the pod is still running.
         """
 
-        await self._start_watching_pods()
+        reflector = await self._start_watching_pods()
+        await reflector.first_load_future
```
consideRatio, line 2189: This is the pattern I've seen elsewhere in the codebase, and I think that avoids some performance penalty of using `await`.

```suggestion
        if not reflector.first_load_future.done():
            await reflector.first_load_future
```

I'm not sure if it makes sense to do this here though, thinking maybe it belongs to `_start_watching_pods` or `_start_reflector`.

What do you think @ivyxjc?

consideRatio (commented):
I think its super tricky to follow how KubeSpawner works with this, so I'm struggling to review this.

Did you run into issues with this using `enable_user_namespaces` set to True @ivyxjc ?

### Related

- https://github.com/jupyterhub/kubespawner/pull/576

ivyxjc:
I ran into the issue with `enable_user_namespaces` set to default value False.

I am encountering the following problem:
When I restart the hub pod, some opening notebooks ran into `Service Unavailable` .
And I found out that when hub starts, hub will delete some users' server form database due to `kubespawner.poll()` does not return correct status.

The root cause is that `kubespwaner.poll()` return status even the first load doest not finish.

https://github.com/jupyterhub/jupyterhub/blob/be07c7ef31ea586fcb49143e8dc6b942aadbb8ec/jupyterhub/app.py#L2527

kubespawner/spawner.py
```diff
@@ -2185,7 +2185,9 @@ async def poll(self):
         just Falsy, to determine that the pod is still running.
         """
 
-        await self._start_watching_pods()
+        reflector = await self._start_watching_pods()
+        if not reflector.first_load_future.done():
+            await reflector.first_load_future
```
consideRatio, line 2190: @minrk I think this makes sense, but would appreciate your review help.

I'm especially thinking about if we should put this code in `_start_watching_pods` or `_start_reflector`.

Currently, `self._start_watching_pods` is awaited from `_start`, `stop`, and `poll`, where `_start` also awaits `await self._start_watching_events`.

Should do the await on `first_load_future` if needed from `_start_reflector`?

minrk: I think awaiting it in `_start_reflector` makes sense. If I were to guess based on my foggy memory of the distant past, I think perhaps `_start_reflector` maybe used to have a requirement to not be async, so it couldn't wait for things? That doesn't appear to be the case anymore.

danilopeixoto:
We implemented a copy of `KubeSpawner` with minor changes. We also noticed the `Hub` was deleting the `spawner` object of running servers because it couldn't find the server resources in the `reflector` at startup (`init_spawners`). The solution presented in the pull request solved our problem. Now the `poll` method waits for the reflector to flood for the first time.

We did not test the implementation in the original `KubeSpawner`.

minrk:
Thanks for the comment @danilopeixoto! I think that this bug may be the cause of https://github.com/jupyterhub/mybinder.org-deploy/issues/2686 leaving orphan pods taking up space  on mybinder.org.

I moved the await of first_load to inside _start_reflector, so it's always awaited and hopefully less likely to get missed.

PR #744, opened 2023-06-08
add 'safe' slug scheme

### LINKED ISSUE #498: Spawning fails for user names starting with non-alphanumeric chars
### Bug description
When a user with a username that starts with a non alphanumeric character (like '-' or '_') tries to spawn, it fails with:

```json
{
    "kind":"Status",
    "apiVersion":"v1",
    "metadata":{
        
    },
    "status":"Failure",
    "message":"Pod \"jupyter--5fusername\" is invalid: metadata.labels: Invalid value: \"-5fusername\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')",
    "reason":"Invalid",
    "details":{
        "name":"jupyter--5fusername",
        "kind":"Pod",
        "causes":[
            {
                "reason":"FieldValueInvalid",
                "message":"Invalid value: \"-5fusername\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')",
                "field":"metadata.labels"
            }
        ]
    },
    "code":422
}
```

#### Expected behaviour

A pod starts for the user

#### Actual behaviour
A user server spawns a pod with a label for the user's name. Label values can't start with non-alphanumeric chars, so this fails.

### How to reproduce

Setup simple kubespawner instance and try logging in with a username starting with '_'

### LINKED ISSUE #737: Server names starting with upper letter fail to start
### Bug description

Named server that start with an upper letter produces escaped labels with invalid characters in label annotations. E.g. for "RStudioServerOption" I get this error from kubernetes:

```
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod \"jupyter-enol-2efernandez---52-53tudio-53erver-4fption\" is invalid: metadata.labels: Invalid value: \"-52-53tudio-53erver-4fption\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')","reason":"Invalid","details":{"name":"jupyter-enol-2efernandez---52-53tudio-53erver-4fption","kind":"Pod","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: \"-52-53tudio-53erver-4fption\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')","field":"metadata.labels"}]},"code":422}
```

#### Expected behaviour

Named servers work independently of the name used

### How to reproduce

1. Enable Named Servers
2. Spawn a Named server starting with capital letter
3. See error

### Your personal set up

zero-to-jupyterhub with several customisations, jupyterhub-kubespawner v5.0.0

Issue is similar to #730 and #498

minrk:
closes #737
closes #498 

implements scheme [described here](https://github.com/jupyterhub/kubespawner/issues/498#issuecomment-1578423254).

- always produces valid values
- uses names, etc. as-is, when valid (no more `-2d`)
- uses stripped `name---hash` when needed
- new behavior is enabled by default, opt-out for backward compatibility via `KubeSpawner.slug_scheme = 'escape'`

Potential issues to address:

- escaping is applied to the final template result, rather than individual strings. This ensures correct values, but means the hash is a function of the full template string, not just the username (or servername). Dealing with length is tricky if we need to hash bits at a time.
- Not everything is a kubernetes field, e.g. working directory should probably get special handling? It has different rules than the label fields
- a breaking change for users to change schemes (just like changing the template config), but at least we don't make the change for them

### DIFF
--- a/docs/source/templates.md
+++ b/docs/source/templates.md
@@ -1,0 +1,156 @@
+(templates)=
+
+# Templated fields
+
+Several fields in KubeSpawner can be resolved as string templates,
+so each user server can get distinct values from the same configuration.
+
+String templates use the Python formatting convention of `f"{fieldname}"`,
+so for example the default `pod_name_template` of `"jupyter-{user_server}"` will produce:
+
+| username         | server name | pod name                                       |
+| ---------------- | ----------- | ---------------------------------------------- |
+| `user`           | `''`        | `jupyter-user`                                 |
+| `user`           | `server`    | `jupyter-user--server`                         |
+| `user@email.com` | `Some Name` | `jupyter-user-email-com--some-name---0c1fe94b` |
+
+## templated properties
+
+Some common templated fields:
+
+- [pod_name_template](#KubeSpawner.pod_name_template)
+- [pvc_name_template](#KubeSpawner.pvc_name_template)
+- [working_dir](#KubeSpawner.working_dir)
+
+## fields
+
+The following fields are available in templates:
+
+| field                    | description                                                                                                                                                  |
+| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `{username}`             | the username passed through the configured slug scheme                                                                                                       |
+| `{servername}`           | the name of the server passed through the configured slug scheme (`''` for the user's default server)                                                        |
+| `{user_server}`          | the username and servername together as a single slug. This should be used most places for a unique string for a given user's server (new in kubespawner 7). |
+| `{unescaped_username}`   | the actual username without escaping (no guarantees about value, except as enforced by your Authenticator)                                                   |
+| `{unescaped_servername}` | the actual server name without escaping (no guarantees about value)                                                                                          |
+| `{pod_name}`             | the resolved pod name, often a good choice if you need a starting point for other resources (new in kubespawner 7)                                           |
+| `{pvc_name}`             | the resolved PVC name (new in kubespawner 7)                                                                                                                 |
+| `{namespace}`            | the kubernetes namespace of the server (new in kubespawner 7)                                                                                                |
+| `{hubnamespace}`         | the kubernetes namespace of the Hub                                                                                                                          |
+
+Because there are two escaping schemes for `username`, `servername`, and `user_server`, you can explicitly select one or the other on a per-template-field basis with the prefix `safe_` or `escaped_`:
+
+| field                   | description                                                                                                                                         |
+| ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `{escaped_username}`    | the username passed through the old 'escape' slug scheme                                                                                            |
+| `{escaped_servername}`  | the server name passed through the 'escape' slug scheme                                                                                             |
+| `{escaped_user_server}` | the username and servername together as a single slug, identical to `"{escaped_username}--{escaped_servername}".rstrip("-")` (new in kubespawner 7) |
+| `{safe_username}`       | the username passed through the 'safe' slug scheme (new in kubespawner 7)                                                                           |
+| `{safe_servername}`     | the server name passed through the 'safe' slug scheme (new in kubespawner 7)                                                                        |
+| `{safe_user_server}`    | the username and server name together as a 'safe' slug (new in kubespawner 7)                                                                       |
+
+These may be useful during a transition upgrading a deployment from an earlier version of kubespawner.
+
+The value of the unprefixed `username`, etc. is goverend by the [](#KubeSpawner.slug_scheme) configuration, and always matches exactly one of these values.
+
+## Template tips
+
+In general, these guidelines should help you pick fields to use in your template strings:
+
+- use `{user_server}` when a string should be unique _per server_ (e.g. pod name)
+- use `{username}` when it should be unique per user, but shared across named servers (sometimes chosen for PVCs)
+- use `{escaped_}` prefix if you need to keep certain values unchanged in a deployment upgrading from kubespawner \< 7
+- `{pod_name}` can be re-used anywhere you want to create more resources associated with a given pod,
+  to avoid repeating yourself
+
+## Changing template configuration
+
+Changing configuration should not generally affect _running_ servers.
+However, when changing a property that may need to persist across user server restarts, special consideration may be required.
+For example, changing `pvc_name` or `working_dir` could result in disconnecting a user's server from data loaded in previous sessions.
+This may be your intention or not! KubeSpawner cannot know.
+
+`pvc_name` is handled specially, to avoid losing access to data.
+If `KubeSpawner.remember_pvc_name` is True, once a server has started, a server's PVC name cannot be changed by configuration.
+Any future launch will use the previous `pvc_name`, regardless of change in configuration.
+If you _want_ to change the names of mounted PVCs, you can set
+
+```python
+c.KubeSpawner.remember_pvc_name = False
+```
+
+This handling isn't general for PVCs, only specifically the default `pvc_name`.
+If you have defined your own volumes, you need to handle changes to these yourself.
+
+## Upgrading from kubespawner \< 7
+
+Prior to kubespawner 7, an escaping scheme was used that ensured values were _unique_,
+but did not always ensure fields were _valid_.
+In particular:
+
+- start/end rules were not enforced
+- length was not enforced
+
+This meant that e.g. usernames that start with a capital letter or were very long could result in servers failing to start because the escaping scheme produced an invalid label.
+To solve this, a new 'safe' scheme has been added in kubespawner 7 for computing template strings,
+which aims to guarantee to always produce valid object names and labels.
+The new scheme is the default in kubespawner 7.
+
+You can select the scheme with:
+
+```python
+c.KubeSpawner.slug_scheme = "escape"  # no changes from kubespawner 6
+c.KubeSpawner.slug_scheme = "safe"  # default for kubespawner 7
+```
+
+The new scheme has the following rules:
+
+- the length of any _single_ template field is limited to 48 characters (the total length of the string is not enforced)
+- the result will only contain lowercase ascii letters, numbers, and `-`
+- it will always start and end with a letter or number
+- if a name is 'safe', it is used unmodified
+- if any escaping is required, a truncated safe subset of characters is used, followed by `---{hash}` where `{hash}` is a checksum of the original input string
+- `-` shall not occur in sequences of more than one consecutive `-`, except where inserted by the escaping mechanism
+- if no safe characters are present, 'x' is used for the 'safe' subset
+
+Since length requirements are applied on a per-field basis, a new `{user_server}` field is added,
+which computes a single valid slug following the above rules which is unique for a given user server.
+The general form is:
+
+```
+{username}--{servername}---{hash}
+```
+
+where
+
+- `--{servername}` is only present for non-empty server names
+- `---{hash}` is only present if escaping is required for _either_ username or servername, and hashes the combination of user and server.
+
+This `{user_server}` is the recommended value to use in pod names, etc.
+In the escape scheme, `{user_server}` is identical to the previous value used in default templates: `{username}--{servername}`,
+so it should be safe to upgrade previous templated using `{username}--{servername}` to `{user_server}` or `{escaped_user_server}`.
+
+In the vast majority of cases (where no escaping is required), the 'safe' scheme produces identical results to the 'escape' scheme.
+Probably the most common case where the two differ is in the presence of single `-` characters, which the `escape` scheme escaped to `-2d`, while the 'safe' scheme does not.
+
+Examples:
+
+| name | escape scheme | safe scheme |
+| username | username | username |
+| has-hyphen | has-2dhyphen | has-hyphen |
+| Capital | `-43apital` (error) | `capital---1a1cf792` |
+| user@email.com | 'user-40email-2ecom' | 'user-email-com---0925f997' |
+| 'a-very-long-name-that-is-too-long-for-sixty-four-character-labels' | 'a-2dvery-2dlong-2dname-2dthat-2dis-2dtoo-2dlong-2dfor-2dsixty-2dfour-2dcharacter-2dlabels' (error) | 'a-very-long-name-that-is-too-long-for---29ac5fd2' |
+| ALLCAPS | '-41-4c-4c-43-41-50-53' (error) | 'allcaps---27c6794c'|
+
+Most changed names won't have a practical effect.
+However, to avoid `pvc_name` changing even though KubeSpawner 6 didn't persist it,
+on first launch (for each server) after upgrade KubeSpawner checks if:
+
+1. `pvc_name_template` produces a different result with `scheme='escape'`
+1. a pvc with the old 'escaped' name exists
+
+and if such a pvc exists, the older name is used instead of the new one (it is then remembered for subsequent launches, according to `remember_pvc_name`).
+This is an attempt to respect the `remember_pvc_name` configuration, even though the old name is not technically recorded.
+We can infer the old value, as long as configuration has not changed.
+This will only work if upgrading KubeSpawer does not _also_ coincide with a change in the `pvc_name_template` configuration.
--- a/docs/source/templates.md
+++ b/docs/source/templates.md
@@ -42,8 +42,8 @@ Because there are two escaping schemes for `username`, `servername`, and `user_s
 
 | field                   | description                                                                                                                                         |
 | ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `{escaped_username}`    | the username passed through the old 'escape' slug scheme                                                                                            |
-| `{escaped_servername}`  | the server name passed through the 'escape' slug scheme                                                                                             |
+| `{escaped_username}`    | the username passed through the old 'escape' slug scheme (new in kubespawner 7)                                                                     |
+| `{escaped_servername}`  | the server name passed through the 'escape' slug scheme (new in kubespawner 7)                                                                      |
 | `{escaped_user_server}` | the username and servername together as a single slug, identical to `"{escaped_username}--{escaped_servername}".rstrip("-")` (new in kubespawner 7) |
 | `{safe_username}`       | the username passed through the 'safe' slug scheme (new in kubespawner 7)                                                                           |
 | `{safe_servername}`     | the server name passed through the 'safe' slug scheme (new in kubespawner 7)                                                                        |
--- a/docs/source/templates.md
+++ b/docs/source/templates.md
@@ -135,13 +135,14 @@ Probably the most common case where the two differ is in the presence of single 
 
 Examples:
 
-| name | escape scheme | safe scheme |
-| username | username | username |
-| has-hyphen | has-2dhyphen | has-hyphen |
-| Capital | `-43apital` (error) | `capital---1a1cf792` |
-| user@email.com | 'user-40email-2ecom' | 'user-email-com---0925f997' |
-| 'a-very-long-name-that-is-too-long-for-sixty-four-character-labels' | 'a-2dvery-2dlong-2dname-2dthat-2dis-2dtoo-2dlong-2dfor-2dsixty-2dfour-2dcharacter-2dlabels' (error) | 'a-very-long-name-that-is-too-long-for---29ac5fd2' |
-| ALLCAPS | '-41-4c-4c-43-41-50-53' (error) | 'allcaps---27c6794c'|
+| name                                                                | escape scheme                                                                                       | safe scheme                                        |
+| ------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------- | -------------------------------------------------- |
+| `username`                                                          | `username`                                                                                          | `username`                                         |
+| `has-hyphen`                                                        | `has-2dhyphen`                                                                                      | `has-hyphen`                                       |
+| `Capital`                                                           | `-43apital` (error)                                                                                 | `capital---1a1cf792`                               |
+| `user@email.com`                                                    | `user-40email-2ecom`                                                                                | `user-email-com---0925f997`                        |
+| `a-very-long-name-that-is-too-long-for-sixty-four-character-labels` | `a-2dvery-2dlong-2dname-2dthat-2dis-2dtoo-2dlong-2dfor-2dsixty-2dfour-2dcharacter-2dlabels` (error) | `a-very-long-name-that-is-too-long-for---29ac5fd2` |
+| `ALLCAPS`                                                           | `-41-4c-4c-43-41-50-53` (error)                                                                     | `allcaps---27c6794c`                               |
 
 Most changed names won't have a practical effect.
 However, to avoid `pvc_name` changing even though KubeSpawner 6 didn't persist it,
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -196,9 +196,7 @@ class KubeSpawner(Spawner):
         # for transitive values (pod_name, dns_name)
         # but not persistent values (namespace, pvc_name)
         if self.enable_user_namespaces:
-            self.namespace = self._expand_user_properties(
-                self.user_namespace_template, slug_scheme="safe"
-            )
+            self.namespace = self._expand_user_properties(self.user_namespace_template)
             self.log.info(f"Using user namespace: {self.namespace}")
 
         self.pod_name = self._expand_user_properties(self.pod_name_template)
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -710,6 +710,10 @@ class KubeSpawner(Spawner):
         {
             'app.kubernetes.io/name': 'jupyterhub',
             'app.kubernetes.io/managed-by': 'kubespawner',
+            # app and heritage are older variants of the modern
+            # app.kubernetes.io labels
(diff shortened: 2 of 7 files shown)


Ph0tonic:
Hi,
Any status on this PR ? It would be great to have this merged, can I help with it ?
Thanks

kubespawner/spawner.py
```diff
@@ -1809,18 +1836,32 @@ def _expand_user_properties(self, template):
...
+            servername = raw_servername
+        else:
+            username = safe_username
+            servername = safe_servername
+
         rendered = template.format(
             userid=self.user.id,
-            username=safe_username,
+            username=username,
+            escaped_username=safe_username,
```
Ph0tonic, line 1851: Do we really need this extra parameter ?

kubespawner/spawner.py
```diff
@@ -1809,18 +1836,32 @@ def _expand_user_properties(self, template):
...
             unescaped_username=self.user.name,
             legacy_escape_username=legacy_escaped_username,
-            servername=safe_servername,
+            servername=servername,
             unescaped_servername=raw_servername,
+            escaped_servername=safe_servername,
```
Ph0tonic, line 1856: Same, do we really need this extra parameter ?

kubespawner/spawner.py
```diff
@@ -1809,18 +1836,32 @@ def _expand_user_properties(self, template):
         safe_username = escapism.escape(
```
Ph0tonic, line 1836: Shall we remove this `legacy_escaped_username` now, as it is not used anymore ?

kubespawner/spawner.py
```diff
@@ -1809,18 +1836,32 @@ def _expand_user_properties(self, template):
...
             self.user.name, safe=safe_chars, escape_char='-'
         ).lower()
+
+        if self.slug_scheme == "safe":
+            # safe slug scheme escapes _after_ rendering the template
+            username = self.user.name
+            servername = raw_servername
+        else:
+            username = safe_username
+            servername = safe_servername
```
Ph0tonic, line 1846: Maybe we could also move the code generating `username` and `servername` inside the if, we do not need to compute these steps if we are not using this non `safe` slug.

Ph0tonic (commented):
A few suggestions an questions.

Ph0tonic:
Hi,
Any status update on this PR ? May I help with anything ?
\cc @minrk

consideRatio:
@minrk I looked at this in order to help nudge it towards merge, but I see that its a very breaking PR to enable this for an existing hub as for example `pvc_name` is gets changed.

```python
            self.namespace = self._expand_user_properties(self.user_namespace_template)
        self.dns_name = self.dns_name_template.format(namespace=self.namespace, name=self.pod_name)
        self.secret_name = self._expand_user_properties(self.secret_name_template)
        self.pvc_name = self._expand_user_properties(self.pvc_name_template)
        if self.working_dir:
            self.working_dir = self._expand_user_properties(self.working_dir)
```

The changes to the labels are far less breaking as it won't affect kubespawner/z2jh/binderhub/grafana-dashboards etc - and only be breaking for external software considering the labels.

minrk:
@consideRatio yeah, I think that's exactly why I didn't finish this. I think the new scheme is fine and works better, but figuring out the transition is the hard part.

minrk:
I'm going to try to pick this one up with #835 / #836 in mind as the transition path:

- leave old labels as they are by default
- make old labels opt-out
- use safe scheme unconditionally in new labels
- use new labels in all selectors, don't rely on old labels for anything

The hard part is still going to be transitioning things like PVC, pod name, etc. We need to make sure that _once established_, they don't change even if the scheme changes. I think this is true already for pod name. Persisting chosen values in Spawner state _should_ do this for us, but will need to think it through and make sure everything is where it's supposed to be, and everything is persisted that should be. [Looking at it](https://github.com/jupyterhub/kubespawner/blob/862f6aabf08e194c3d00a79705f99e9ae25e5c62/kubespawner/spawner.py#L2138-L2152), unfortunately pvc name is not persisted, so that's going to be hard.

We may have to keep both schemes alive and check at runtime:

```
if new name exists:
   use new name
else:
  if old name exists:
    use old name, warn
  else:
    use new name
```

It may turn out that the best we can do is a breaking change with a documented _manual_ migration step.

PVCs can be renamed through the somewhat [frightening process](https://stackoverflow.com/questions/48403542/how-to-rename-persistence-volume-claim) of:

- get pv id from pvc
- set volume retention to Retain
- delete old pvc
- patch claimRef on volume
- create new pvc with specified volumeName

or use [krew rename-pvc](https://github.com/stackitcloud/rename-pvc) (which does the same thing, but has the benefit of being written by not us).

kubespawner/spawner.py
minrk: I don't think so, since it is only unused for the opt-in new scheme.

kubespawner/spawner.py
minrk: to ensure the same fields can still be used in the new scheme, yes if anyone needs to make sure names are preserved across schemes

kubespawner/spawner.py
minrk: same answer. Probably not, but this is the only way to preserve capabilities to produce the same strings.

yuvipanda:
Thank you for picking this back up, @minrk!

consideRatio:
Thank you for working this @minrk!!

These points may already be known and considered, but i wanted to highlight it when considering the paths onwards just in case.

**A external resource complexity to be aware of**

A key complexity for any change to username expansion is external resources, such as just creating a volume mount on the pod with subPath including username. We can't know if the external nfs volume includes the old username expansion or not, so we can't determine if its ok to switch over or not etc.

**Kubespawner itself only use component label i think**

I think the only use of labels currently by kubespawner itself is to reduce the set of resources to look at by the reflector using the component label. A performance aspect, because then the actual resource names are looked up among that set.

minrk:
Thanks for the notes!

> I think the only use of labels currently by kubespawner itself is to reduce the set of resources to look at by the reflector using the component label.

Close. I think there's one more place: when services are enabled, labels are also used to route service traffic to the pod, which also [uses the full label set](https://github.com/jupyterhub/kubespawner/blob/593405d24af017bc759b6763a0a2af8af4de0f38/kubespawner/spawner.py#L2096) (I'm not sure that's the right choice, but it should at least include user and server names).

Now that this mostly works, I think there are some problems in the switch to handling escape of whole strings at a time:

- variable expansion is done _all over the place_, with wildly different rules:
  - pod/pvc_name use object naming rules
  - labels have different rules (it's fine to use a strict subset of objects and labels for both)
  - subdir mount paths (totally different)
  - `pod_connect_ip` where hostname rules apply, the pod name may want to be used, but the rest of the string definitely must _not_ be escaped
- some strings should _include_ other strings, but whole-string escaping makes that impossible unless all expanded results can also be referenced by name:
  - `pod_name_template "jupyter-{username}--{servername}"` -> `jupyter-escape--server--abc123`
  - `pod_connect_ip "jupyter-{username}--{servername}.svc.local" -> `jupyter-escape--serversvclocal--def456`

That said, we _cannot guarantee correct strings_ if we don't handle the full slug. I think it's simply not feasible to have one `expand_user_properties` that's used all over the place for pod names, URLs, templates, paths and have it produce correct, consistent results.

I don't really know how to address this. I think it only makes sense to apply the 'safe' slug scheme to kubernetes resource names and labels, and not in other contexts. It _usually_ doesn't make sense to apply the escaping to mounts or subdirectories, but _sometimes_ it will. The rules, however, are completely different and not really knowable. Applying the new scheme only to resource names 

I'm going to give this a read through, and instead of changing what `expand_user_properties` does globally, I'm going to try to:

- only apply the new scheme to k8s resource names and labels (drastically reduce potential impact of the change)
- make `pod_name` (or perhaps dedicated `{k8s_name}`) available in the template namespace, so it can be referenced by other template strings without modification

yuvipanda:
Thank you for diligently working through this hard problem, @minrk.

Let's bump a major version whenever we release this.

minrk:
Getting closer, but still tricky stuff to think about:

- object names should be safe by default
- it should be possible to force the old scheme for consistency (either a single global flag, or new template fields that keep the old value)
- it should be possible to reference created resources where object rules may not apply (add pod_name, pvc_name, etc. to template namespace)
- we should not lose track of pvcs across upgrade (pvc_name should have been in get_state all along)
- applying appropriate context-dependent rules in blackboxes like in `_expand_all` is not feasible, so per-field escaping is really the only way to go

The `{username}--{servername}` collision problem changes slightly depending on how/when we apply the escape because e.g. `username="üser", servername=""` produces `'ser--73506260'` which can also be produced with `username="ser", servername="73506260"`, so we would need to pick a different delimiter and/or hash joiner to avoid that.

minrk:
I think I'm narrowing it down to a scheme we can safely and comprehensibly switch _to_. I mainly need to hash out the transition process and write out some tests and docs for how that should work. In all cases, current behavior can be preserved unchanged _explicitly_ by replacing `{username}` with `{escaped_username}` in templates, which will continue to be the old-style value with no plans to change. There is a new `{safe_username}` that is escaped with the new scheme, and a new `{user_server}` field that is the join of `{username}--{servername}` and can be used anywhere a unique slug for a given user+server should be used.

In general, we have:

- `username`, `servername`, `user_server` - value determined by `slug_scheme`. Currently _defaults_ to the new `safe`. Always match exactly one of:
- `ecaped_` variants - current scheme with escapism. Not guaranteed to produce valid values, especially for labels. `{escaped_user_server}` is identical to `{escaped_username}--{escaped_servername}` to preserve backward compatibility
- `safe_` variants - new, safe scheme that always produces a safe label. `safe_user_server` is _not_ always identical to `{safe_username}--{safe_servername}`, but it is in the situations where neither needs any escaping, which is most of the time.
- `pod_name`, `pvc_name`, and `namespace` are available to reference in templates (I'm mostly thinking of `extra_` resources that may refer to them).

I've also removed the config to make the labels in #835 optional since I don't think that helped in the end, as covered by @yuvipanda's comments over there.

I feel like I am getting a good handle on resource names (pod_name, pvc_name), but figuring out the right thing to do for work_dir is a good deal more complex. Part of the trouble is that resource names are a safer transition because we can check if a resource at the old name exists and use it. We can also use state persistence to keep things sticky across config changes. We can't check if a filesystem path exists in time to make a choice.

Filesystems also tend to have very different, much more permissive, rules than k8s object names, but they can vary quite a bit! So what should we do for work_dir?

- use new scheme and document how to keep old scheme with template config
- continue to use old scheme by default, because it's much less likely to have caused any problems for paths
- somehow try to identify if old scheme has been used, and use new if not

manics:
If it helps we can potentially force admins to make a deliberate decision when upgrading Z2JH. If we add a new Z2JH config value to the schema and deliberately set it to an invalid value in values.yaml an admin must set the config. It's inconvenient for anyone expecting a seamless upgrade, but the schema validation should happen before anything is deployed so it shouldn't break anything.

yuvipanda:
> If it helps we can potentially force admins to make a deliberate decision when upgrading Z2JH. 

I like this idea!

@minrk I am torn on storage related defaults. I do think they're the *hardest* to migrate, as it's not a simple 'ok, kick everyone\'s servers out and back'. We should pick between these two options you have mentioned:

> use new scheme and document how to keep old scheme with template config
> continue to use old scheme by default, because it's much less likely to have caused any problems for paths

So it's a choice admins can make at upgrade time.

I can't think of any real way to do 
> somehow try to identify if old scheme has been used, and use new if not

without a million edge cases, so I think we shouldn't try.

minrk:
Yeah, I think there are two categories of templates to consider in migration:

- safe to change on server restart (pod name, label values, user namespace) - persisting in state, re-evaluating template in clear_state handles this well.
- unsafe to change on restart (pvc name, working_dir)

I think all of these should really be persisted in Spawner state, but unfortunately only the _unsafe to change_ values are not remembered, so the first time we deploy whatever changes we make here, we can't know what the previous values were.

I also think there are different categories of change for the unsafe values when Hub restarts with different template config:

- deliberate config, _should_ change existing servers
- deliberate config, should affect _new_ launches, but not existing ones
- default behavior change, should affect _new_ launches, but not existing ones

While I can fix it to make pvc_name sticky (the current state of this PR), the hard part becomes allowing _deliberate_ change to these values (possibly relevant to #761, I think, depending on implementation as in #820). I don't know the best way to tell the difference between a change that _should_ orphan the pvc vs a change that should not. Right now, we are in the situation where only the first case is handled, and any change to defaults or template config orphans pvcs unless admins perform some complex manual migrations. In this PR, once a pvc is mounted it is remembered, but there is no way for admins to change that value for existing servers except with direct db operations (or post-load_state such as when loading user_options). The advantage of the status quo is user config is always respected: clear and simple. The disadvantage is there's no easy way for admins to fix problems if they realize the current scheme doesn't work for all of their users.

I think we can actually 'detect' if previous values were used on first run after upgrade, because if the state dictionary in `load_state` is non-empty, we know with some confidence that previous values were used. The one case that wouldn't be handled is if upgrade _and template config change_ happen at the same time. That may be enough to not try, though.

manics:
Yet another option: we could have a `pre-install`/`pre-upgrade` Helm hook that scans all Kubernetes resources which potentially belong to Z2JH. If the default config doesn't require renaming resources they installation/upgrade proceeds so admins don't notice anything, otherwise there hook exits with an error which should stop the deployment, and the admin updates the config.

Whether this is worth doing depends on how likely it is that deployments have resource names that don't match the new scheme.

minrk:
I realize one way to recognize changing templates in _future_ upgrades is to persist not just the resolved strings, but the templates themselves. Then it will be possible to distinguish when the template config changed.

We really have these readily detectable cases:

- load_state called without empty dict: new Spawner, nothing to load, no migration concerns
- load_state called with non-empty dict, pvc_name not present (or other value that identifies current version): upgrade happened since last launch, there may be migration steps to take
- load_state called with keys recognizable from this version: we know any upgrading steps have happened already

Increasingly, I think we should be placing the kubespawner _version_ in a couple of places (resource annotations, spawner state) so we can make decisions with this information more precisely and with more confidence.

minrk:
So we can _choose_ to be able to detect changed templates if we want going forward. The only option we can't detect is upgrade and changed template at the same time, since we can't change what previous versions persist. The only real solution to that is documentation: do not change templates during upgrade. The downside is that it's loaded on a per-spawner basis, so only safe after _every user server has started at least once since the upgrade_, not just restarting the Hub once.

I've been thinking about working_dir, and since the default is None, I think we should not try to handle it. As @manics mentioned, one version of this is a helm-upgrade hook that checks for working_dir config for a template containing `{username}` which is the value that will change, and recommend explicitly using `{escaped_username}` (no change) or `{safe_username}` (new scheme). We could make persistence of working_dir across template changes user-selectable via e.g. `remember_working_dir` config. That way, deployments can _choose_ if changing working_dir template should change existing servers or not. We could do the same for pvc_name and potentially any other resources that should be sticky.

Originally, when I opened this PR, the new scheme was opt-in. As it is now, it is the default. A third option is 'safe-upgrade' scheme, where we use the new scheme on object names/labels specifically, but not in general. The biggest downside of this that I can see is the same template won't resolve to the same string in different contexts (e.g. if the pvc_name template is referenced in extra_containers or something).

There are a lot of choices to make, and not many with clear right answers. I will try to get this PR to a state where at least one version of each choice is made deliberately, and we can better discuss which parts need improving.

minrk:
Forgot to add: one reason I don't think handling working_dir is as important is getting it wrong shouldn't result in data loss - getting the wrong value due to a change will likely result in servers either failing to start or not having access to the right files until the template is updated, but there is no _persistent_ cost to this being wrong for a time.

So it's disruptive (servers may not start or may _appear_ to be missing files if they successfully start), but the fix is clear: replace `{username}` with `{escaped_username}` in working_dir config to stick with old values, and there should be no lasting effect to this having been wrong in the meantime. pvc names are much more costly, because whole new PVCs will be created if the name changes and that makes a mess that is hard to cleanup.

minrk:
I wouldn't call this "done" by any means, but think it's now at least "complete" in that I have a pretty good understanding of what's changed, what works on upgrade and what requires user intervention, it's documented more thoroughly than before, and the new upgrade logic is tested and explained.

So I think this is ready for review!

consideRatio:
## NFS storage considerations

When using NFS storage, one may have a single PVC configured with a `subPath` referencing `"{username}"`.

<details><summary>Click to see example z2jh / kubespawner config</summary>

This is z2jh Helm chart config:

```yaml
jupyterhub:
  singleuser:
    storage:
      type: static
      static:
        pvcName: home-nfs
        subPath: "{username}"
```

Via [a z2jh provided jupyterhub_config.py](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/bbc38ca286ac15f1dbd8c69c5cad676057e7bb10/jupyterhub/files/hub/jupyterhub_config.py#L287-L299), this leads to KubeSpawner config:

```python
c.KubeSpawner.volumes = [
    {
        "name": "home",
        "persistentVolumeClaim": {
            "claimName": "home-nfs",
        },
    },
],
    {
        "mountPath": "/home/jovyan",
        "name": "home",
        "subPath": "{username}",  # this is whats important
    },
]
```

</details>

In such situations, `{username}` have been the same as `{escaped_username}`, but will transition to being `{safe_username}` KubeSpawner v7. This will lead to mounting a new folder for usernames with characters including for example: `.`, `-`, `@`, or capital letters.

A quick but essential config change before upgrading to z2jh v4 / kubespawner v7 would be to do a change like this:

```diff
jupyterhub:
  singleuser:
    storage:
      type: static
      static:
        pvcName: home-nfs
-        "subPath": "{username}"
+        "subPath": "{escaped_username}"
```

For new hubs though, it would make sense to do `"subPath": "{username}".

docs/source/templates.md
```diff
@@ -0,0 +1,157 @@
...
+
+Because there are two escaping schemes for `username`, `servername`, and `user_server`, you can explicitly select one or the other on a per-template-field basis with the prefix `safe_` or `escaped_`:
+
+| field                   | description                                                                                                                                         |
+| ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `{escaped_username}`    | the username passed through the old 'escape' slug scheme                                                                                            |
+| `{escaped_servername}`  | the server name passed through the 'escape' slug scheme                                                                                             |
+| `{escaped_user_server}` | the username and servername together as a single slug, identical to `"{escaped_username}--{escaped_servername}".rstrip("-")` (new in kubespawner 7) |
```
consideRatio, line 47: These `escaped_` template field names are also new in kubespawner 7

consideRatio:
In case one wants to migrate some resource, for example NFS storage folder names, one may need to identify how `{username}` has changed for a set of users before and after. This comment explores some steps one could need to take in such migration.

## Detecting if `{username}` changed between v6 and v7 for specific usernames

```python
import string

from escapism import escape
from kubespawner.slugs import safe_slug

test_username = "some-unescaped-username"

escaped_safe_chars = set(string.ascii_lowercase + string.digits)
escaped_username = escapism.escape(test_username, safe=escaped_safe_chars, escape_char='-').lower()
safe_username = safe_slug(test_username)
if safe_username != escaped_username:
    print(f"safe_username != escaped_username: {safe_username} != {escaped_username}")
```

## Getting all usernames via JupyterHub API

TODO

## Creating a map of old -> new `{username}` expansion

TODO

kubespawner/spawner.py
consideRatio: I think they need to be made available no matter what, because it should always be possible to reference `{safe_username}` even though `{username}` only conditionally becomes `safe_username` for example.

kubespawner/spawner.py
```diff
@@ -1338,11 +1376,23 @@ def _validate_image_pull_secrets(self, proposal):
...
+    handle_legacy_names = Bool(
+        True,
+        config=True,
+        help="""handle legacy names and labels
+        
+        kubespawner 7 changed the scheme for computing names and labels to be more reliably valid.
+        In order to preserve backward compatibility, the old names must be handled in some places.
+
+        You can safely disable this if no PVCs were created or running servers were started
+        before upgrading to kubespawner 7.
```
consideRatio, line 1395: This is only used to handle a transition to safe name expansions for PVCs, not for anything else like the leading help text `handle legacy names and labels` indicates it may.

I think updating this to describe its just about managing the PVC naming makes sense rather than trying to handle anything beyond the PVC naming.

kubespawner/spawner.py
```diff
@@ -183,17 +187,25 @@ def __init__(self, *args, **kwargs):
         # By now, all the traitlets have been set, so we can use them to
         # compute other attributes
 
+        # namespace, pod_name, dns_name are persisted in state
+        # these same assignments should match clear_state
         if self.enable_user_namespaces:
-            self.namespace = self._expand_user_properties(self.user_namespace_template)
+            self.namespace = self._expand_user_properties(
+                self.user_namespace_template, slug_scheme="safe"
+            )
```
consideRatio, line 195: I've been thinking about if/why slug_scheme should be hardcoded to "safe" here, leaving this comment for myself to not forget this open question.

consideRatio: The `user_namespace_template` default is `"{hubnamespace}-{username}"`, so expanding that with `slug_scheme="safe"` in a hardcoded way here ought to hardcode a breaking change that will make PVCs be lost as they are namespaced resources.

I think if we hardcode `slug_scheme="safe"` here, we'd introduce a breaking change that can't be avoided by configuration for users of `enable_user_namespaces` that also use PVCs.

minrk: I believe this is not _intended_ to be a breaking change because namespace is persisted in Spawner state. So this hardcoded scheme only affects _new_ spawners that have never launched before, and ensures namespace is always valid which the escape scheme does not. Otherwise, `namespace` is sticky and remembered for all launches after the first for a given Spawner.

I've updated the comment ot try to clarify the relationship of values set in `get_state` vs `__init__`.

minrk: I think this may be a remnant of 'safe' not being the default. Now that it is, since escape is explicitly opt-in, I don't think we need to do this. I'll remove the hardcoded `safe` here, and leave it only on the labels, where it is actually required.

kubespawner/spawner.py
minrk: Yes, I intended to have this for a catch-all config "check if the upgraded defaults might have detached something without user input," but PVCs were the only thing I found where there was something to be done. Maybe namespace should be handled as well?

kubespawner/spawner.py
```diff
@@ -2910,8 +3105,39 @@ async def _start(self):
             self._last_event = events[-1]["metadata"]["uid"]
 
         if self.storage_pvc_ensure:
-            pvc = self.get_pvc_manifest()
+            if (
+                self.handle_legacy_names
+                and not self._pvc_exists
+                and self._state_kubespawner_version == "unknown"
```
minrk, line 3111: add `and self.remember_pvc_name` because the result of this process won't be remembered otherwise, and note to this effect to help string.

consideRatio:
I chatted with @minrk about this just now, here is some planning made explicit:

1. @minrk does a final pass on this PR, possibly with no changes
2. We go for a kubespawner 7 beta release
3. We bump to kubespawner 7 beta in z2jh to get a dev release out
4. I'll do some test upgrades to verify things seem ok
   - [ ] test the proposed migration in https://github.com/jupyterhub/kubespawner/pull/744#issuecomment-2241287196
   - [ ] test a user with username something.something
   - [ ] test a user with a pvc created already
   - [ ] test a user with a pvc not created already

Notes fot he future:
- z2jh changelog/migration notes
  - see a specific migration step in the end of https://github.com/jupyterhub/kubespawner/pull/744#issuecomment-2241287196
  - script to look at usernames etc considered in https://github.com/jupyterhub/kubespawner/pull/744#issuecomment-2241297361 considered

kubespawner/spawner.py
minrk: namespaces don't need to be handled because they were persisted in 6.x, so I think this is okay as-is.

minrk:
I determined that `handle_legacy_names` should indeed be subordinate to `remember_pvc_name`, since it is itself an extended implementation of remembering the previously used pvc_name. Both are true by default.

minrk:
Naturally, tests have started to fail for unrelated reasons on the last commit. Looking into it.

minrk:
Fixed tests with #846

consideRatio (approved):

consideRatio:
Thank you for amazing effort into this @minrk!!

I'm on vacation for two weeks now, but hope to find time to contribute towards release of beta and testing anyhow

Issue #753, opened 2023-07-17
high k8s etcd/apiserver load due to pod listings


juliantaylor:
### Bug description

In clusters with many pods the kubespawner creates significant load on the etcd/apiserver due to its frequent pod listings when using `enable_user_namespaces` which does listings of all pods in the whole cluster.
full pod listings even with a label selector are very expensive as the pods in the etcd are only indexed by the namespaces, the label index is only performed in the apiserver caches.
When not setting resource_version (equivalent to setting it to "") the apiserver will always retrieve all data from etcd instead of its caches and thus the listing becomes very expensive [0].
Additionally your watches have a default timeout of 10s causing it to relist the pods frequently. In a cluster with 10k pods the listing can take several seconds in the etcd.

To improve the situation I would suggest to set resource_version=0 in the initial list of the _list_and_update which means the request can be fulfilled by the apiserver caches without hitting the etcd (which makes the label selector effective). This also means you do not have the strong consistency guarantee of the etcd.

If this is inappropriate I would suggest to put an additional warning into the `enable_user_namespaces` documentation to note that the watch timeout should be increased from the default 10s in large clusters.

[0] https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions

juliantaylor:
something along the line of this should be more efficient (not yet tested):

```patch
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -208,7 +208,7 @@ class ResourceReflector(LoggingConfigurable):
 
         self.watch_task = None
 
-    async def _list_and_update(self):
+    async def _list_and_update(self, resource_version=None):
         """
         Update current list of resources by doing a full fetch.
 
@@ -218,9 +218,12 @@ class ResourceReflector(LoggingConfigurable):
         kwargs = dict(
             label_selector=self.label_selector,
             field_selector=self.field_selector,
+            resource_version=resource_version,
             _request_timeout=self.request_timeout,
             _preload_content=False,
         )
+        if resource_version:
+            kwargs["resource_version_match"] = "NotOlderThan"
         if not self.omit_namespace:
             kwargs["namespace"] = self.namespace
 
@@ -264,6 +267,8 @@ class ResourceReflector(LoggingConfigurable):
             selectors.append("field selector=%r" % self.field_selector)
         log_selector = ', '.join(selectors)
 
+        resource_version = None
+
         cur_delay = 0.1
 
         if self.omit_namespace:
[...]

PR #755, opened 2023-07-24
improve efficiency of reflector

juliantaylor:
The reflector performs a _list_and_update every watch timeout (default 10s) requesting most recent data from the apiserver via the default resource_version="".

This requires the apiserver to retrieve the latest data from the backing store (etcd) which can be very expensive in large clusters. For example if omit_namespace is set the full cluster pod listings require several seconds retrieve all data from the etcd in clusters with more then 10000 pods. The label selector filtering is only done in the apiserver.

As the _list_and_update is used for watches the listing should set the resource version and resource version match to tell the apiserver that its locally cached data is sufficient.
The initial listing will fetch any data the apiserver currently has to prepopulate the watch, the watch itself will then fetch whatever is remaining from the etcd.

The subsequent relistings after the watch timeouts use their last received resource version in the same way.

For details see https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions

Closes gh-753

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -208,7 +208,7 @@ class ResourceReflector(LoggingConfigurable):
 
         self.watch_task = None
 
-    async def _list_and_update(self):
+    async def _list_and_update(self, resource_version=None):
         """
         Update current list of resources by doing a full fetch.
 
@@ -221,6 +221,9 @@ class ResourceReflector(LoggingConfigurable):
             _request_timeout=self.request_timeout,
             _preload_content=False,
         )
+        if resource_version is not None:
+            kwargs["resource_version"] = resource_version
+            kwargs["resource_version_match"] = "NotOlderThan"
         if not self.omit_namespace:
             kwargs["namespace"] = self.namespace
 
@@ -264,6 +267,11 @@ class ResourceReflector(LoggingConfigurable):
             selectors.append("field selector=%r" % self.field_selector)
         log_selector = ', '.join(selectors)
 
+        # fetch Any (=api-server cached) data from apiserver on initial fetch
+        # see https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions
+        # for more information
+        resource_version = "0"
+
         cur_delay = 0.1
 
         if self.omit_namespace:
@@ -282,7 +290,7 @@ class ResourceReflector(LoggingConfigurable):
             start = time.monotonic()
             w = watch.Watch()
             try:
-                resource_version = await self._list_and_update()
+                resource_version = await self._list_and_update(resource_version)
                 watch_args = {
                     "label_selector": self.label_selector,
                     "field_selector": self.field_selector,
@@ -325,6 +333,7 @@ class ResourceReflector(LoggingConfigurable):
                         else:
                             # This is an atomic operation on the dictionary!
                             self.resources[ref_key] = resource
+                            resource_version = resource["metadata"]["resourceVersion"]
                         if self._stopping:
                             self.log.info("%s watcher stopped: inner", self.kind)
                             break
@@ -346,6 +355,9 @@ class ResourceReflector(LoggingConfigurable):
                 self.log.debug("Cancelled watching %s", self.kind)
                 raise
             except Exception:
+                # ensure we request a valid resource version on retry,
+                # needed on 410 Gone errors
+                resource_version = "0"
                 cur_delay = cur_delay * 2
                 if cur_delay > 30:
                     self.log.exception("Watching resources never recovered, giving up")


kubespawner/reflector.py
```diff
@@ -208,7 +208,7 @@ def __init__(self, *args, **kwargs):
 
         self.watch_task = None
 
-    async def _list_and_update(self):
+    async def _list_and_update(self, resource_version=""):
```
yuvipanda, line 211:
```suggestion
    async def _list_and_update(self, resource_version=None):
```

Reading through https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions, I see that there's a difference between this being unset, and being set to the string "0" but not for the empty string "". 

So I think we should either use `None` here to indicate that it will be *unset* if not specified, or "0" to get the behavior for that. As "" doesn't seem to be defined, I don't think we should default to that.

I'm not super set on `None` or `"0"`, we can go with whatever you think works better here :)

kubespawner/reflector.py
```diff
@@ -218,9 +218,12 @@ async def _list_and_update(self):
         kwargs = dict(
             label_selector=self.label_selector,
             field_selector=self.field_selector,
+            resource_version=resource_version,
             _request_timeout=self.request_timeout,
             _preload_content=False,
         )
+        if resource_version:
```
yuvipanda, line 225:
```suggestion
        if resource_version is not None:
```

kubespawner/reflector.py
```diff
@@ -218,9 +218,12 @@ async def _list_and_update(self):
         kwargs = dict(
             label_selector=self.label_selector,
             field_selector=self.field_selector,
+            resource_version=resource_version,
```
yuvipanda, line 221: Should we set this only if it's not `None`?

kubespawner/reflector.py
```diff
@@ -264,6 +267,9 @@ async def _watch_and_update(self):
             selectors.append("field selector=%r" % self.field_selector)
         log_selector = ', '.join(selectors)
 
+        # fetch Any (=api-server cached) data from apiserver on initial fetch
```
yuvipanda, line 270:
```suggestion
        # fetch Any (=api-server cached) data from apiserver on initial fetch
        # see https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions
        # for more information
```

yuvipanda (changes requested):
Pretty excited for this PR, @juliantaylor! Thank you for taking the time, I appreciate it :)

I've left one review comment about the function default, but otherwise, given the tests pass, I am pretty happy with this. If you can resolve that comment one way or another, unless someone else objects, I'll be happy to merge this in a couple days.

kubespawner/reflector.py
juliantaylor: it is not really clear what unset means, in the code resource version seems to always just be a string and it is compared against an empty string which I interpret as meaning unset.

then another issue is what the python api library does with None vs a string but I assume it is likely equivalent so using None may be clearer

kubespawner/reflector.py
juliantaylor: this is intentional so that it also does not match the empty string which I'm pretty sure also means unset

kubespawner/reflector.py
juliantaylor: As it is confusing what unset means in the api and its clients Ill update the code to use None and explicitly not add it to the kwargs

kubespawner/reflector.py
```diff
@@ -218,9 +218,12 @@ async def _list_and_update(self):
         kwargs = dict(
             label_selector=self.label_selector,
             field_selector=self.field_selector,
+            resource_version=resource_version,
             _request_timeout=self.request_timeout,
```
juliantaylor, line 222: I do think it would be a good idea to change this timeout to a larger value for the watches, even if cached lists are still very expensive. Other k8s components typically relist watches in the range of minutes to never, every 10 seconds is a bit much.
Though probably should be done in a different PR.

kubespawner/reflector.py
yuvipanda: In the page you linked to, I see a column in the table for `resourceVersion not set`. I am slightly confused about what you mean by "it is not really clear what unset means". By "in the code", do you mean in this codebase or in kubernetes?

Another alternative is to just require `resource_version` to always be passed in here, and not let that be optional.

kubespawner/reflector.py
yuvipanda: Yep let's consider that in a different PR

minrk (approved):

kubespawner/reflector.py
juliantaylor: yes I meant the kubernetes code, the resource version there is a string which either has an opaque value (which happens to be etcd's modification index) or is empty

empty means unset

kubespawner/reflector.py
```diff
@@ -346,6 +355,9 @@ async def _watch_and_update(self):
                 self.log.debug("Cancelled watching %s", self.kind)
                 raise
             except Exception:
+                # ensure we request a valid resource version on retry,
+                # needed on 410 Gone errors
+                resource_version = "0"
```
juliantaylor, line 360: added this, should not be needed with the `NotOlderThan` resource_version_match but it probably makes sense to start from the beginning on errors.

yuvipanda:
Thanks, @juliantaylor!

Issue #758, opened 2023-07-25
API Calls to spawn user servers do not handle defaults in profileList correctly


GeorgianaElena:
### Bug description

Following up on https://github.com/jupyterhub/kubespawner/pull/735, there is still an issue related to how defaults are considered when `user_options` is empty and server spawned is triggered through an api call, rather than through the UI.

Unfortunately, it's not easy to spot where the issue is coming from, as the profileList codebase has grown a lot and it's becoming harder to follow. We should take this opportunity, of searching the issue to also do some refactoring and possibly move all the profile handling code into a separate file. 

#### Expected behaviour

When an user spawn is triggered through an api call, with an empty `user_options` dict, a server should be spawned with default profile and option choices.

#### Actual behaviour

An error is thrown from the validatation of the options like in https://github.com/2i2c-org/infrastructure/issues/2146#issuecomment-1649445203.

consideRatio:
Closed by #769

consideRatio:
There is a related bug to this and I'm writing about it here for now.

We have an expectation that users starting a server and then starting another server shouldn't get influenced by what they did before, but I think that may be a broken expectation.

I'm digging into this @GeorgianaElena. I'm planning to provide a "Explanation" section about how profile_list is implemented on top of the Spawner base class functionality as I need it for myself to understand what goes wrong. I think the crux may be in [this code](https://github.com/jupyterhub/jupyterhub/blob/4.0.2/jupyterhub/user.py#L737-L744), but I'm trying to piece together my undertanding on what goes on before addressing it.

GeorgianaElena:
@consideRatio, let's track this bug in https://github.com/jupyterhub/kubespawner/issues/772, as the current issue is related to a totally different bug. What do you think?

PR #768, opened 2023-08-22
Add changelog for 6.1.0

### LINKED ISSUE #767: Release planning for 6.1.0
KubeSpawner would benefit greatly from a release with #742 in it, what I think was a regression from previous major versions of KubeSpawner. Z2JH 3.0.0 - 3.0.3 users are affected by it. The consequences of this bug is that user pods can be left running in k8s while JupyterHub thinks the servers are stopped, and this can incurr significant cloud costs. Users with a stranded server can still re-start it via JupyterHub though.

To make this release, we should also include notes on how users can cleanup such stranded servers intelligently. @minrk has made [a script that inspects k8s resources and the jupyterhub state](https://gist.github.com/minrk/e15653520847746e643a6ca5e48d3949) to help figure out what user servers should be considered for cleanup.

## Todo

- #702, fixed by #773
- #742
- #785
- #787
- #768

## Todo related to `unlisted_choice`

- #758, fixed by #772
- #756, fixed by #778
- #757

## Todo considered outside the scope of this release

- #775

consideRatio:
I propose that we cut a release now with bugfixes, a performance improvement, and the `unlisted_choice` functionality part of `profile_lists` -> `profile_options`.

- Closes #767

If approved by monday next week, I'll cut a release then.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -4,13 +4,62 @@
 
 ## [Unreleased]
 
-#### Breaking changes
+## 6.1
 
-- JupyterHub 4 is now required by KubeSpawner.
-  [#726](https://github.com/jupyterhub/kubespawner/pull/726)
-- Versions of K8s older than 1.24 are no longer supported, KubeSpawner may still
-  work but this is not guaranteed.
-  [#726](https://github.com/jupyterhub/kubespawner/pull/726)
+### [6.1.0] - 2023-09-28
+
+```{warning}
+If you have been using z2jh 3.0 or KubeSpawner 5.0-6.0, you may have _orphaned
+user server pods_ due to a now fixed bug, these are pods that are running but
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
@consideRatio, I think we might want to also fix this https://github.com/jupyterhub/kubespawner/issues/758 as part of a new release. A workaround we've used for this was https://github.com/2i2c-org/infrastructure/pull/2865/files.

consideRatio:
I've not pushed on this as i felt we absolutely need a note in the changelog to help users remediate the consequences of the bug as well - this is the place to make such note.

I plan to test a script by Min tomorrow helping with that, possibly make a smaller tweak, and then link to it from the changelog.

shaneknapp:
@consideRatio thanks for the update!

yuvipanda (approved):
LGTM

PR #769, opened 2023-08-23
[bufix] Allow POST requests without profile_options specified (defaults will be used)

GeorgianaElena:
This is a test that reproduces the bug in https://github.com/jupyterhub/kubespawner/issues/758.

It should help us find the bug by investigating the particularities of this test profile list and also know if we'd fix the bug.


GeorgianaElena:
I believe the issue was that we were validating user's chosen `user_options`  against the profile's `profile_options`even if the user didn't chose anything, hence `user_options` was empty.

consideRatio (approved):

yuvipanda:
I just tested this and it works!

consideRatio:
Nice work identifying this @GeorgianaElena!! This looks good!!

yuvipanda:
@consideRatio the docstring became out of date after this PR, I opened https://github.com/jupyterhub/kubespawner/pull/771 to update it.

PR #774, opened 2023-08-31
Rework of profile_list backend validation for readability and details

consideRatio:
## Summary

I've failed to isolate the changes in this PR in smaller commits, but below are some highlights.

- Revised docstrings
- Some kubespawner internal functions was async for no reason so I made them non-async
- Some kubespawner internal fucntions got their signatures tweaked
- Fixed bug making warnings not be emitted about unregonized options passed via web forms

## Review together with my comments

In the "Files changed" view you can see comments I've made on the changes in various sections. I think that may be the easiest way to review this PRs changes.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3011,69 +3011,97 @@ class KubeSpawner(Spawner):
             user_options (dict): the selected profile in the user_options form,
                 e.g. ``{"profile": "cpus-8"}``
         """
-        profile = formdata.get('profile', [None])[0]
+        profile_slug = formdata.get('profile', [None])[0]
 
-        options = {'profile': profile}
+        # initialize a dictionary to return
+        user_options = {}
 
-        # Load any options if they are set for this profile, and this profile only
-        # In the default template we use, all form options for a particular profile
-        # come through of the format 'profile-option-{profile}-{option-slug}'
+        # if a profile is declared, add a dictionary key for the profile, and
+        # dictionary keys for the formdata related to the profile's
+        # profile_options, as recognized by being named like:
         #
-        # FIXME: Since we only return recognised formdata in user_options, we
-        #        shouldn't need later validation of this, and we need to emit
-        #        warnings here rather than later about unrecognised options.
+        #     profile-option-{profile_slug}--{profile_option_slug}
         #
-        if profile:
-            option_formdata_prefix = f'profile-option-{profile}--'
+        if profile_slug:
+            user_options["profile"] = profile_slug
+            prefix = f'profile-option-{profile_slug}--'
             for k, v in formdata.items():
-                if k.startswith(option_formdata_prefix):
-                    stripped_key = k[len(option_formdata_prefix) :]
-                    options[stripped_key] = v[0]
+                if k.startswith(prefix):
+                    profile_option_slug = k[len(prefix) :]
+                    user_options[profile_option_slug] = v[0]
 
-        return options
+        # warn about any unrecognized form data, which is anything besides
+        # "profile" and "profile-option-" prefixed keys
+        unrecognized_keys = set(formdata)
+        unrecognized_keys = unrecognized_keys.difference({"profile"})
+        unrecognized_keys = [
+            k for k in unrecognized_keys if not k.startswith("profile-option-")
+        ]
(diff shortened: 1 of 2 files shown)


kubespawner/spawner.py
```diff
@@ -3236,44 +3260,27 @@ async def load_user_options(self):
...
         profile_list = self.profile_list
         if callable(profile_list):
             profile_list = await maybe_future(profile_list(self))
         profile_list = self._get_initialized_profile_list(profile_list)
 
-        selected_profile = self.user_options.get('profile', None)
-        selected_profile_user_options = dict(self.user_options)
-        if selected_profile:
-            # Remove the 'profile' key so we are left with only selected profile options
-            del selected_profile_user_options['profile']
```
consideRatio, line 3253: Since `user_options` are filtered to `selected_profile_user_options` via `_options_from_form` anyhow, and its not an issue to retain `profile` there - I opted to remove the complexity of making a copy in order to mutate the user options.

kubespawner/spawner.py
```diff
@@ -3221,10 +3249,6 @@ def _get_initialized_profile_list(self, profile_list: list):
 
         return profile_list
 
-    # partial set of recognised user_option keys
-    # used for warning about ignoring unrecognised options
-    _user_option_keys = {'profile'}
-
```
consideRatio, line 3227: I inlined this to the logic where it ended up used, in `_options_from_form`. I twasn't correctly named, because it was really `_formdata_keys` or similar rather than valid keys for the parsed `user_options`.

kubespawner/spawner.py
```diff
@@ -3011,69 +3011,97 @@ def _options_from_form(self, formdata):
...
+        unrecognized_keys = set(formdata)
+        unrecognized_keys = unrecognized_keys.difference({"profile"})
+        unrecognized_keys = [
+            k for k in unrecognized_keys if not k.startswith("profile-option-")
+        ]
+        if unrecognized_keys:
+            self.log.warning(
+                "Ignoring unrecognized form data in spawn request: %s",
+                ", ".join(map(str, sorted(unrecognized_keys))),
+            )
```
consideRatio, line 3044: This logic existed elsewhere before in a broken state. It was broken before becaue it looked at parsed user_options, but had logic about the formdata that is to be parsed to user_options.

consideRatio (commented):
Since I've failed to break apart this PR in smaller PRs or even several commits, I've made review comments about the various sections.

Note that you should probably read these review comments by visiting the "Files changed" view where you see all lines of code associated with the comments.

GeorgianaElena (approved):
Thanks for leaving inline comments @consideRatio! They were really helpful understanding what's going on and where to go look next.

I did a couple of passes of the changes in this PR, focusing on the changes in behavior, rather than  name changes and docstrings updates, as I don't want to get into subjective  change proposals given the complexity of this PR.

I am now pretty confident, the changes in functionality are beneficial and do not break anything ✨ So, I believe this is safe to merge! Thank you for improving the logic!

Also, while discussing about the implementation of the `unlisted_choice` logic with @yuvipanda, there was a general conclusion that would be beneficial to have all the logic related to profiles into its own module, as the `spawner.py` file has grown a lot and it's became harder to follow. What do you think? Do you think it would be wise to do this as part of this PR, or at least this future reaease?

consideRatio:
Thank you soo much for reviewing @GeorgianaElena!!!

> I did a couple of passes of the changes in this PR, focusing on the changes in behavior, rather than name changes and docstrings updates, as I don't want to get into subjective change proposals given the complexity of this PR.

I appreciate that - you are most welcome to iterate on the subjective parts still though!

> there was a general conclusion that would be beneficial to have all the logic related to profiles into its own module, as the spawner.py file has grown a lot and it's became harder to follow. What do you think? Do you think it would be wise to do this as part of this PR, or at least this future reaease?

I agree - I'd love to see a clearly scoped set of logic extracted from spawner.py, and I think there is good chance we clearly scope profile_list related logic somewhere else.

I think extracting code from spawner.py into another file is perfect for a dedicated PR and shouldn't go into this PR. This PR was messy and would just sully the other changed.

If you want we can hold the 6.1.0 release for such work, but I don't mind the overhead of an additional minor/patch release!

Issue #757, opened 2023-07-24
Allow configuring text shown in profile_option dropdown for selecting unlisted_choice


yuvipanda:
Follow-up to https://github.com/jupyterhub/kubespawner/pull/735#discussion_r1254129852, we currently have the text "Other..." hardcoded in the dropdown box of choices if `unlisted_choice` is enabled. This should be configurable - for example, if the choice is for an image, you might want to say "Enter image manually" or something like that

Issue #756, opened 2023-07-24
Don't show dropdown in profile_options if only unlisted_choice is enabled, no choices are present


yuvipanda:
Follow-up to https://github.com/jupyterhub/kubespawner/pull/735#discussion_r1254129852, if a profile_option is configured to have `unlisted_choice.enabled` to `True`, but `choices` is not set, we currenlty show a dropdown with just 'Other' as an option. This is kinda ugly. Instead, it should just not show the dropdown at all.

PR #776, opened 2023-09-02
Unlisted choice improvements

batpad:
Towards #756 and #757 -

 - Allows for an `other_text` option inside `unlisted_choice` which allows user to specify a custom text in the select drop-down
 - Hide select drop-down and label if no choices are defined
 - Handles no choices in backend processing

I might have underestimated the complexity in hiding the select dropdown if there are no `choices` present, and that bit seems to have gotten a bit ugly. Will leave some comments in the code to see if there's things we can improve.

@yuvipanda @consideRatio @GeorgianaElena - firstly, thank you so much for fixing up issues with the `unlisted_choices` work, and many apologies for the delay in getting to these fixes. I am not super happy with the implementation in this PR, and I'll leave some more notes with the code. I did a bit of testing though, and it *seems* to me like it all works as expected.

If this approach seems ok to you all, I think what might be pending:

 - Create an example profile option to not have any `choices` but have an `unlisted_choice` to make testing easier
 - Create a test with no `choices` to make sure the backend handling is covered with tests.

### DIFF
--- a/jupyterhub_config.py
+++ b/jupyterhub_config.py
@@ -49,6 +49,7 @@ c.KubeSpawner.profile_list = [
                 'unlisted_choice': {
                     'enabled': True,
                     'display_name': 'Image Location',
+                    'other_text': 'Enter Image manually',
                     'validation_regex': '^pangeo/.*$',
                     'validation_message': 'Must be a pangeo image, matching ^pangeo/.*$',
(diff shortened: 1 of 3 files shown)


batpad:
Oops, my bad - I thought I was up-to-date with kubespawner:main but apparently not. Just going to close with PR and open a new one cleaned up. Apologies for the noise.

PR #785, opened 2023-09-14
Support lists and dicts as values in `kubespawner_override`

yuvipanda:
Reported in https://github.com/2i2c-org/infrastructure/pull/3124#issuecomment-1716254905
for a list.

### DIFF
--- a/kubespawner/utils.py
+++ b/kubespawner/utils.py
@@ -3,6 +3,7 @@ Misc. general utility functions, not tied to KubeSpawner directly
 """
 import copy
 import hashlib
+from typing import Union
 
 
 def generate_hashed_slug(slug, limit=63, hash_length=6):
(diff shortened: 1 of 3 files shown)


kubespawner/utils.py
```diff
@@ -222,3 +223,45 @@ def recursive_update(target, new):
...
+    placeholders for passed in kwargs, no formatting is performed.
+
+    recursive_format("{v}", v=5) -> Returns "5"
+    recrusive_format("{a}") -> Returns "{a}" rather than erroring, as is
+    the behavior of "format"
+    """
+    if isinstance(format_object, str):
+        return format_object.format_map(IgnoreMissing(kwargs))
+    elif isinstance(format_object, list):
+        return [recursive_format(i, **kwargs) for i in format_object]
```
minrk, line 258: Not really important, since this is probably not getting called a million times, but it would be a bit more efficient to create the IgnoreMissing once and pass it around like `recursive_format(obj, format_dict)`, rather than create a fresh kwargs dict (calling a function with `**kwargs` creates a new dict to be passed) and IgnoreMissing for every value.

kubespawner/utils.py
```diff
@@ -222,3 +223,45 @@ def recursive_update(target, new):
...
+    elif isinstance(format_object, dict):
+        return {
+            recursive_format(k, **kwargs): recursive_format(v, **kwargs)
+            for k, v in format_object.items()
+        }
+    else:
+        raise ValueError(
```
minrk, line 265: Is it reasonable for recursive_format to be called on a dict with non-string scalars, e.g. numbers? In that case, presumably they should be returned as-is rather than raise. I think this raise means we're still requiring all override values to be strings.

minrk (commented):
Looks great! Is it a draft because you have more places where you think recursive_format should be called?

kubespawner/utils.py
yuvipanda: @minrk ah good catch! The values can indeed be ints. I added a test, and relaxed this to not throw but instead return the object unformatted.

kubespawner/utils.py
yuvipanda: @minrk I want the call sites to still be able to pass in values as **kwargs, rather than have to import and pass in IgnoreMissing. Think it's ok to leave this as is?

yuvipanda:
@minrk this was marked as draft mostly still by accident! Ready to go now.

I agree it would be useful to call this from other places too! Probably not in this PR though.

kubespawner/utils.py
minrk: I do, it's not a big deal. I was thinking the function would expect a dict (like `format_map`), but we'd cast to IgnoreMissing if it wasn't cast already, so the cast would happen at most once, rather than N times, but still callers wouldn't need to know anything about IgnoreMissing.

kubespawner/utils.py
```diff
@@ -222,3 +222,42 @@ def recursive_update(target, new):
 
         else:
             target[k] = v
+
+
+class IgnoreMissing(dict):
+    """
+    Dictionary subclass for use with format_map
+
+    Renders missing values as "{key}", so format strings with
```
consideRatio, line 231:
```suggestion
    Returns missing dictionary keys' values as "{key}", so format strings with
```

consideRatio (commented):
I pushed a docs commit and made some suggestions, with them this LGTM and can be considered approved. If they LGTY feel free to go for a merge @yuvipanda!

PR #791, opened 2023-10-02
fix: unneeded escaping of capital letters in k8s labels

byteMuncher0x:
The current version of the Kubespawner escapes all capital letters from labels before they are asigned to K8s resources. This is unnecessary as per the [K8s standards](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -122,9 +122,9 @@ class KubeSpawner(Spawner):
     # KubeSpawner objects.
     reflectors = {}
 
-    # Characters as defined by safe for DNS
+    # Characters defined as safe for DNS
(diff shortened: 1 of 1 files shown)


yuvipanda:
Further discussion in https://github.com/jupyterhub/kubespawner/issues/737

yuvipanda:
Thank you very much for your PR, @byteMuncher0x. Unfortunately this turns out to be more complex and possibly a breaking change. @minrk has picked up this work again in https://github.com/jupyterhub/kubespawner/pull/744, and progress is being made. I invite you to participate in review and suggestions there! But I'm going to close this PR, as it unfortunately will cause a breaking change as is.

PR #793, opened 2023-10-04
Dynamically retrieve podCIDR for routing in CONTRIBUTING

yuvipanda:
The CIDR from which pod IPs are allocated can change depending on platform and driver used for `minikube`. On my machine, it's 10.244.0.0/24. This changes the instructions to determine the podCIDR dynamically by talking to kubernetes, rather than hardcoding it in.

I personally prefer just repeating the command here rather than using an environment variable, as it means the single line can be just copy pasted.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,14 +81,14 @@ rules on your computer. We will now walk you through the steps to get going:
 
     ```bash
     # Linux
-    sudo ip route add 172.17.0.0/16 via $(minikube ip)
+    sudo ip route add $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}") via $(minikube ip)
     # later on you can undo this with
-    sudo ip route del 172.17.0.0/16
+    sudo ip route del $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}")
 
     # MACOS
-    sudo route -n add -net 172.17.0.0/16 $(minikube ip)
+    sudo route -n add -net $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}") $(minikube ip)
     # later on you can undo this with
-    sudo route delete -net 172.17.0.0
+    sudo route delete -net $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}")
     ```
 
     ### Troubleshooting


yuvipanda:
It would be useful for someone with a different configuration than mine to test this command to make sure it provides appropriate output for you. I'm on Mac OS.

CONTRIBUTING.md
```diff
@@ -81,14 +81,14 @@ rules on your computer. We will now walk you through the steps to get going:
...
     # later on you can undo this with
-    sudo ip route del 172.17.0.0/16
+    sudo ip route del $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}")
 
     # MACOS
-    sudo route -n add -net 172.17.0.0/16 $(minikube ip)
+    sudo route -n add -net $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}") $(minikube ip)
     # later on you can undo this with
-    sudo route delete -net 172.17.0.0
+    sudo route delete -net $(kubectl get node minikube -o jsonpath="{.spec.podCIDR}")
```
minrk, line 91: FWIW, I've never run any of these `route` commands on mac and everything has worked fine. I'm not sure what problem it solves, though? This line does get the right podCIDR.

minrk (approved):

CONTRIBUTING.md
yuvipanda: @minrk without these, the hub can't talk to the spawned pods. But I suspect that it's now only needed when the driver is qemu not docker.

minrk: Yeah, I don't think I've ever done this, either when I used minikube in a VM in the olden days, or now with docker. I don't think I ever ran minikube with --driver=qemu, though.

Should it just be removed, or left in a "you may also need this if..." section? I don't know the situation on linux.

yuvipanda: Hmm, I definitely ran into errors with jupyterhub trying to reach out to the spawned pods without this. I didn't specify `--driver=qemu` - if you don't have docker desktop or another docker thing running, `--driver=qemu` is the default. 

I agree it could be moved into a subsection. I can dig up the failure mode when this isn't done.

minrk (approved):
In any case, this PR is already a strict improvement, if you want to go ahead and merge. Adding detail/conditions can be its own PR if it will take some figuring out.

yuvipanda:
Thanks @minrk :)

consideRatio:
Wieee thanks for taking the time to work this @yuvipanda and @minrk!!! :heart: :tada:

I added a documentation label for this PR, a small detail that reduces the workload when its time to make a changelog!

PR #797, opened 2023-10-19
make sure futures can be cancelled

### LINKED ISSUE #796: Object has no attribute 'cancel' - a secondary error when handling a timeout error
From [this forum post](https://discourse.jupyter.org/t/jupyterhub-helm-chart-3-1-0-k8s-1-27-6-user-pods-not-spawning-got-timeouterror/21956) by @yambottle, I saw that we have a bug that shows during another erroring situation.

```python-traceback
[E 2023-10-18 19:00:12.626 JupyterHub user:884] Unhandled error starting test's server: 'coroutine' object has no attribute 'cancel'
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/user.py", line 798, in spawn
        url = await gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/kubespawner/spawner.py", line 2669, in _start
        future.cancel()
        ^^^^^^^^^^^^^
    AttributeError: 'coroutine' object has no attribute 'cancel'
```

The related code is here:

https://github.com/jupyterhub/kubespawner/blob/6d9d9a36df3077b6ba22e4a94107ffe96fc2b18c/kubespawner/spawner.py#L2660-L2670

minrk:
In order to cancel tasks, we need to have a handle on the Future, not just the Coroutine object.

The code we have implements TaskGroup-style cancellation of pending tasks, since gather doesn't cancel unfinished tasks unless gather _itself_ is cancelled, but TaskGroups are new in Python 3.11.

See the note in the docs for [asyncio.gather](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather):

> TaskGroup provides stronger safety guarantees than gather for scheduling a nesting of subtasks: if a task (or a subtask, a task scheduled by a task) raises an exception, TaskGroup will, while gather will not, cancel the remaining scheduled tasks).

i.e. TaskGroup has the behavior "if anything here fails, _stop everything still running_ and then propagate the error", while gather has the behavior "if anything here fails, propagate the error, but leave any other tasks running for the user to clean up or not." And since we don't yet require Python 3.11, we use gather and then implement the cancellation ourselves.

closes #796

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2467,11 +2467,18 @@ class KubeSpawner(Spawner):
             reflector = cls.reflectors.pop(key)
             tasks.append(reflector.stop())
 
+        # make sure all tasks are Futures so we can cancel them later
+        # in case of error
+        futures = [asyncio.ensure_future(task) for task in tasks]
         try:
-            await asyncio.gather(*tasks)
+            await asyncio.gather(*futures)
         except Exception:
-            for task in tasks:
-                task.cancel()
+            # cancel any unfinished tasks before re-raising
+            # because gather doesn't cancel unfinished tasks.
+            # TaskGroup would do this cancel for us, but requires Python 3.11
+            for future in futures:
+                if not future.done():
+                    future.cancel()
             raise
 
     def start(self):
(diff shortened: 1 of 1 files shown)


minrk:
It may well make sense to actually change the wait to `return_when=ALL_COMPLETED` so there's nothing to cancel, especially in the stop_all. But that's not the most important, really, since it's a cleanup utility purely for use in tests and always means a serious problem in the test suite if there's any error anyway.

Pro waiting for everything instead of cancelling:

- don't need manual cancel
- all tasks _tried_ to complete

Con:

- slightly more complex to re-raise errors
- delays reporting of errors in what's already necessarily a serious failure (can't start or stop reflectors!), after which nothing really makes sense anyway

minrk:
The third option is not cancelling at all. But then we're in the situation of a task still running with nothing waiting for it (e.g. stop could still be outstanding when the next start is called, etc.). Again, this wouldn't be the biggest deal because there are likely serious kubernetes communication problems whenever this comes up, but leaking tasks nobody's waiting for is generally something to avoid, which I think is why folks added trio-style TaskGroups so folks can write some code that does "start some async things; but no matter what happens, when I exit this context, none of those tasks are running".

consideRatio (approved):
@minrk :heart: :tada: thank you for resolving this, helping me learn with great PR comments, and providing inline notes to help us update this in the future to use TaskGroup!!

PR #807, opened 2023-11-10
Omit ImagePullPolicy by default

yuvipanda:
Our documentation says that if ':latest' is used as a tag in an image, ImagePullPolicy will be automatically set by Kubernetes to 'Always'. This cites https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting.

However, that's only true if the imagePullPolicy is *omitted*! But we actually *do* set it, to 'IfNotPresent', so this behavior is actually not what we get. This came to light when a user noticed that on nodes with images already present, ':latest' was not taking effect. Looking at the pod definition, we see:

```yaml
image: quay.io/henrykmodzelewski/2i2c-eosc211:latest
imagePullPolicy: IfNotPresent
```

This PR just omits imagePullPolicy by default, so we *actually* get the behavior we said we were getting.

### DIFF
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -140,7 +140,7 @@ def make_pod(
     image_pull_policy:
         Image pull policy - one of None, 'Always', 'IfNotPresent' or 'Never'. Decides
         when kubernetes will check for a newer version of image and pull it when
-        running a pod. If set to None, it will omitted from the spec.
+        running a pod. If set to None, it will be omitted from the spec.
 
     image_pull_secrets:
         Image pull secrets - a list of references to Kubernetes Secret resources
--- a/kubespawner/objects.py
+++ b/kubespawner/objects.py
@@ -80,7 +80,7 @@ def make_pod(
     cmd,
     port,
     image,
-    image_pull_policy,
+    image_pull_policy=None,
     image_pull_secrets=None,
     node_selector=None,
     uid=None,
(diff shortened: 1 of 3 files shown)


yuvipanda:
Because `allow_none` is not currently true, it's actually not possible to get the default kubernetes behavior right now!

kubespawner/objects.py
```diff
@@ -138,9 +138,9 @@ def make_pod(
         arguments
 
     image_pull_policy:
-        Image pull policy - one of 'Always', 'IfNotPresent' or 'Never'. Decides
+        Image pull policy - one of None, 'Always', 'IfNotPresent' or 'Never'. Decides
         when kubernetes will check for a newer version of image and pull it when
-        running a pod.
+        running a pod. If set to None, it will omitted from the spec.
```
GeorgianaElena, line 143:
```suggestion
        running a pod. If set to None, it will be omitted from the spec.
```

kubespawner/spawner.py
```diff
@@ -721,21 +721,16 @@ def _deprecated_changed(self, change):
...
-        is `:latest`. For more information on image pull policy,
-        refer to `the Kubernetes documentation <https://kubernetes.io/docs/concepts/containers/images/>`__.
-
-
-        This configuration is primarily used in development if you are
-        actively changing the `image_spec` and would like to pull the image
-        whenever a user container is spawned.
+        Defaults to `None`, which means it is omitted. This leads to it behaving
+        like 'Always' when a tag is absent or 'latest', and 'IfNotPresent' when
+        the tag is specified to be something else.
```
GeorgianaElena, line 733:
```suggestion
        the tag is specified to be something else, per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting.
```

GeorgianaElena (approved):

minrk (approved):

yuvipanda:
Thanks, @GeorgianaElena!

yuvipanda:
Merging as there was enough approvals!

PR #808, opened 2023-11-19
update_k8s_model: log override at info not warn

manics:
warning implies the admin should takes some action in response, but this is intentionally configured by the admin.


consideRatio (approved):
Seems reasonable to me, lets go for it!

PR #820, opened 2024-01-16
Fix #761 expand pvc_name outside of init

### LINKED ISSUE #761: Some _template options don't work when defined in profile_list profiles
https://github.com/jupyterhub/kubespawner/issues/528 touches on this but the issue is more general than described there.

### Bug description

Some of the `*_template` options are expanded in `KubeSpawner.__init__()` and never again. Overriding them in profile_list profiles therefore doesn't work - you can override the template, but since the template has already been expanded, overriding doesn't produce the desired effect.

See e.g. [pvc_name](https://github.com/jupyterhub/kubespawner/blob/cd17869ed4fe0eb9e65dd3bc7b994687a2a72b8e/kubespawner/spawner.py#L194), which is set in __init__ and then used in [get_pod_manifest()](https://github.com/jupyterhub/kubespawner/blob/cd17869ed4fe0eb9e65dd3bc7b994687a2a72b8e/kubespawner/spawner.py#L2091) as-is. Instead, I think pvc_name_template should be expanded in get_pod_manifest().

#### Expected behaviour

Expansion of templates should be deferred wherever possible so that their overridden values can be respected.

#### Actual behaviour

Some templates are expanded during __init__ and never again, so overriding them has no effect.

### How to reproduce

Deploy z2jh with 
```
  singleuser:
    profileList:
      - display_name: Override test
        default: true
        kubespawner_override:
          service_account: default
          volumes:
            - name: user-home
              persistentVolumeClaim:
                claimName: override-claim-{userid}--{servername}
          volume_mounts:
            - name: user-home
              mountPath: /home/jovyan
          pvc_name_template: override-claim-{userid}--{servername}
```

And try to launch your server. It will fail, because the volume definitions are expanded after overrides are applied, but the PVC name is not.

Ph0tonic:
Partial fix for #761 by moving the expansion of `pvc_name` and `secret_name` outside of `__init__`.

\cc @yuvipanda

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -192,7 +192,6 @@ class KubeSpawner(Spawner):
         )
         self.secret_name = self._expand_user_properties(self.secret_name_template)
 
-        self.pvc_name = self._expand_user_properties(self.pvc_name_template)
         if self.working_dir:
             self.working_dir = self._expand_user_properties(self.working_dir)
         if self.port == 0:
@@ -2085,7 +2084,7 @@ class KubeSpawner(Spawner):
             annotations=annotations,
         )
 
-    def get_pvc_manifest(self):
+    def get_pvc_manifest(self, pvc_name):
         """
         Make a pvc manifest that will spawn current user's pvc.
         """
@@ -2099,7 +2098,7 @@ class KubeSpawner(Spawner):
         storage_selector = self._expand_all(self.storage_selector)
 
         return make_pvc(
-            name=self.pvc_name,
+            name=pvc_name,
             storage_class=self.storage_class,
             access_modes=self.storage_access_modes,
             selector=storage_selector,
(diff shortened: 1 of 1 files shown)


manics:
I've added the breaking label since this changes method signatures that may be used by subclasses.

kubespawner/spawner.py
```diff
@@ -2662,6 +2660,8 @@ async def _make_create_resource_request(self, kind, manifest):
 
     async def _start(self):
         """Start the user's pod"""
+        pvc_name = self._expand_user_properties(self.pvc_name_template)
```
minrk, line 2663: I've been looking at this in #744, and `pvc_name` should actually be one of the things persisted in Spawner state, because the pvc name should probably be preserved to avoid data loss even if it. Resolving the template here would ensure the new template is used and the old pvc is orphaned. But that's _intended_ when overrides are used as in #761.

Should we:

1. not try to track existing pvcs (status quo, this PR), or
2. try not to lose data (what I'm trying to do in #744)

If we should try not to lose data, how do we distinguish between a changed pvc for a profile (should change) vs config (should _not_ change if prior pvc already created)?

Issue #651, opened 2022-10-14
Improved user experience for slow starting pods by improving messaging


jmunroe:
### Proposed change

I think the user experience from kubespawner's `progress` could be improved.  Currently, the language used is fairly specific to kubernetes (refering to things like 'tainted' nodes that don't necessarily have meaning for end users.

When a users sees a "warning" from kubernetes they are not able to understand that as something to be concerned about or whether that is normal behaviour.  User behaviour appears to be, upon seeing warnings or waiting a "long time" (which in a world of instant on might be anything longer than 1-2 mins!), causing a server to shut down and restarted. This bounces the user to the end of the queue which is probably not the intended effect.

Is there a way of augmenting the process output when waiting for server to spawn?  Ideally, giving some user feed back about how long this is "supposed" to take (maybe drawn from metrics of previous server spawns?)

Also, I think it is wise to advise the user that a new node needs to be provisioned (and usually takes xx of time) and alerting that image is being downloaded (and usually takes xxx of time). I think would provide valuable feedback to an end user. Bonus feature: progress bar (25% when node is being provisioned, 50% when node is provisioned, 75% when the container image is downloaded). It doesn't need to be precise; users just need some guidance on how many more steps need to occur before their servers appears. (And guidance that what they are experiencing is normal, expected behaviour).

Finally,  I think it should shared with users that there is also a "time-out" of so many seconds so that is the maximum amount of time you should every need to be waiting! if it is taking longer than that, something is definitely wrong.

### Who would use this feature?
Improve user experience by given more useful information about the start up process which will potentially reduce anxiety from users.  

### (Optional): Suggest a solution
(Based on a conversation with @yuvipanda and inspired from https://github.com/berkeley-dsep-infra/datahub/issues/3845)

balajialg:
Thanks, @jmunroe for validating and amplifying @ryanlovett's request upstream! Both the progress bar to track the pod startup and the countdown to timeout ideas look promising. It will definitely reduce the number of error messages we get from students who are anxious during assignment deadlines at Berkeley.

shaneknapp:
ohai!  i'm shane, and taking over yuvi's role at UCB.

i've put this PR together, which addresses *some* of the suggestions here:
https://github.com/berkeley-dsep-infra/datahub/pull/3963

right now, after 60sec it inject a message every 5sec imploring the user to be patient...  this is hopefully something i'd like to eventually get upstream.  :)

PR #825, opened 2024-03-20
add slow_spawn_message progress for slow spawns

shaneknapp:
this PR  adds a configurable message to present to users after an amount of time has passed, imploring patience.

~~adding a delay was useful in debugging some grafana metrics (https://github.com/berkeley-dsep-infra/datahub/issues/4237, https://github.com/berkeley-dsep-infra/datahub/pull/4241).~~

the 'patience message' was added as users would believe that hubs were down/not working/etc when it took longer than a couple of minutes to launch their notebooks, particularly when a new GCP node was spinning up during peak usage times (https://github.com/berkeley-dsep-infra/datahub/issues/3845, https://github.com/berkeley-dsep-infra/datahub/pull/3963).

this has been incredibly useful for us at UC Berkeley and has saved both us and our users many hours of unnecessary support and debugging requests due to slow launches. 

ref:  https://github.com/jupyterhub/kubespawner/issues/651

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -12,6 +12,7 @@ import os
 import re
 import string
 import sys
+import textwrap
 import warnings
 from functools import partial
 from typing import Optional, Tuple, Type
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2729,9 +2729,7 @@ class KubeSpawner(Spawner):
             ):
                 # don't spam the user, so only update the timer message every few seconds
                 if timer % self.slow_spawn_message_frequency == 0:
-                    patience_message = textwrap.dedent(
-                        self.slow_spawn_message
-                    )
+                    patience_message = textwrap.dedent(self.slow_spawn_message)
                     patience_message = patience_message.format(seconds=timer)
 
                     yield {
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2740,7 +2740,10 @@ class KubeSpawner(Spawner):
 
             # if the timer is greater than self.server_spawn_launch_timer_threshold
             # display a message to the user with an incrementing count in seconds
-            if timer >= self.server_spawn_launcher_timer_threshold and self.server_spawn_launch_timer_enabled:
+            if (
+                timer >= self.server_spawn_launcher_timer_threshold
+                and self.server_spawn_launch_timer_enabled
+            ):
                 # don't spam the user, so only update the timer message every few seconds
                 if timer % self.server_spawn_launch_timer_threshold == 0:
                     patience_message = textwrap.dedent(
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -12,6 +12,7 @@ import os
 import re
 import string
 import sys
+import time
 import textwrap
 import warnings
 from functools import partial
@@ -2703,7 +2704,7 @@ class KubeSpawner(Spawner):
         next_event = 0
 
         # count in seconds to time single user server spawn duration
-        timer = 0
+        start_time = time.perf_counter()
 
         break_while_loop = False
         while True:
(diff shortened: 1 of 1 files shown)


shaneknapp:
@yuvipanda @consideRatio

manics:
I'm not a fan of adding properties just for testing, especially since we already have a lot of complicated configuration properties, and it's trivial to add a delay by subclassing the spawner in-line in the Z2JH config.

A custom user message about the delay sounds useful though. Is there any way to put this in the parent class so it's available to all spawners, or is it inherently tied to KubeSpawner?

shaneknapp:
sorry, this completely and utterly fell off of my radar!

> I'm not a fan of adding properties just for testing, especially since we already have a lot of complicated configuration properties, and it's trivial to add a delay by subclassing the spawner in-line in the Z2JH config.
> 
fair enough.  :)

> A custom user message about the delay sounds useful though. Is there any way to put this in the parent class so it's available to all spawners, or is it inherently tied to KubeSpawner?

ah, i didn't think of that!  good point -- it's absolutely not tied to the kubespawner, and i think it should be trivial to port this to the parent class.

shaneknapp:
alright, so @ryanlovett and i chatted about this for a bit today and we were a little curious on how exactly proceed.

currently, the only two spawners that override the default `jupyerterhub.spawner.Spawner.progress` function are kubespawner and batchspawner (https://github.com/jupyterhub/batchspawner/blob/main/batchspawner/batchspawner.py#L492).

i'd be curious/concerned about adding in a `while` loop w/basic logic would break things, as the `process` function in both kube and batch spawners use different hooks to indicate that the `start()` function completed (`start_future` and `state_isrunning` respectively).

do you have any thoughts?  perhaps @yuvipanda or @minrk would like to take a look as well?

edit:  not sure why the doc tests are failing...

```
Converting `source_suffix = ['.rst', '.md']` to `source_suffix = {'.rst': 'restructuredtext', '.md': 'restructuredtext'}`.
ERROR: Invalid value `None` in intersphinx_mapping['https://docs.python.org/']. Expected a two-element tuple or list.
```

kubespawner/spawner.py
```diff
@@ -1688,6 +1689,42 @@ def _validate_image_pull_secrets(self, proposal):
...
+        Sets a delay of N seconds between updates to the message buffer, so
+        that we don't spam the user with too many messages.
+        """,
+    )
+
+    server_spawn_launch_timer_message = Unicode(
+        "Server launch is taking longer than expected. Please be patient! Current time spent waiting: XXX seconds.",
+        config=True,
+        help="""
+        The injected timing message to display to the user. The string XXX
```
manics: Can we make this a python template string instead, e.g. `blah {seconds} blah`? Then expand it using `server_spawn_launch_timer_message.format(seconds=seconds)`

kubespawner/spawner.py
```diff
@@ -1895,6 +1896,34 @@ def _validate_image_pull_secrets(self, proposal):
...
+        Sets a delay of N seconds between updates to the message buffer, so
+        that we don't spam the user with too many messages.
+        """,
+    )
+
+    server_spawn_launch_timer_message = Unicode(
+        "Server launch is taking longer than expected. Please be patient! Current time spent waiting: {seconds} seconds.",
+        config=True,
+        help="""
+        The injected timing message to display to the user. The variable {seconds}
```
manics:
```suggestion
        The injected timing message to display to the user. The variable `{seconds}`
```

Backticks should make this appear as `code` instead of normal text

kubespawner/spawner.py
```diff
@@ -1895,6 +1896,34 @@ def _validate_image_pull_secrets(self, proposal):
...
+        will be replaced by the number of seconds the spawn has taken in
+        self.progress().
```
manics:
```suggestion
        will be replaced by the number of seconds the spawn has taken in
        `self.progress()`.
```

manics: Or should we omit mention  of `self.progress()` because it's an internal detail that an admin shouldn't care about?

minrk: Yeah, I think that makes sense.

kubespawner/spawner.py
```diff
@@ -2689,6 +2721,21 @@ async def progress(self):
...
 
+            # if the timer is greater than self.server_spawn_launch_timer_threshold
+            # display a message to the user with an incrementing count in seconds
+            elapsed = time.perf_counter() - start_time
+            if (
+                elapsed >= self.slow_spawn_message_threshold
+                and self.slow_spawn_message_threshold > 0
+            ):
+                # don't spam the user, so only update the timer message every few seconds
+                if elapsed % self.slow_spawn_message_frequency == 0:
```
minrk: elapsed won't be an integer anymore, so I think you'll need:

```python
if elapsed >= last_message_time + self.slow_spawn_frequency:
    show_message
    last_message_time = elapsed
```

kubespawner/spawner.py
```diff
@@ -2673,6 +2702,10 @@ async def progress(self):
         progress = 0
         next_event = 0
 
+        # timers for deciding when to show slow spawn patience message
+        start_time = time.perf_counter()
+        last_message_time = float()
```
minrk: One last thing:

```suggestion
        last_message_time = start_time
```

then I think we're set!

kubespawner/spawner.py
```diff
@@ -2673,6 +2702,10 @@ async def progress(self):
         progress = 0
         next_event = 0
 
+        # timers for deciding when to show slow spawn patience message
+        start_time = time.perf_counter()
+        last_message_time = start_time
```
minrk: Sorry, I was wrong, the units for this are relative to start_time:


```suggestion
        last_message_time = 0
```

minrk (approved):

PR #837, opened 2024-05-03
Improve exception handling for listing Kubernetes resources

josefhandl:
When the Kubernetes API fails to list Kubernetes resources in the reflector (for example too strict permissions), no relevant error message appears in the logs, and the program fails. I'm adding a check to see if the call is successful. If not, an error with the API is printed, and an exception is thrown.

This is what the previous logs look like (`enable_user_namespaces = True`):
```
[I 2024-05-02 13:20:03.726 JupyterHub spawner:188] Using user namespace: user-jhu
[E 2024-05-02 13:20:03.739 JupyterHub reflector:412] Initial list of pods failed
      File "/opt/kubespawner/kubespawner/reflector.py", line 410, in start
        await self._list_and_update()
      File "/opt/kubespawner/kubespawner/reflector.py", line 249, in _list_and_update
        for p in initial_resources["items"]
    KeyError: 'items'
    
[E 2024-05-02 13:20:03.739 JupyterHub spawner:2411] Reflector with key ('pods', None) failed to start.
      File "/opt/kubespawner/kubespawner/spawner.py", line 2409, in catch_reflector_start
        await func
      File "/opt/kubespawner/kubespawner/reflector.py", line 410, in start
        await self._list_and_update()
      File "/opt/kubespawner/kubespawner/reflector.py", line 249, in _list_and_update
        for p in initial_resources["items"]
    KeyError: 'items'
```

...and this now:
```
[I 2024-05-02 13:20:26.046 JupyterHub spawner:188] Using user namespace: user-jhu
[E 2024-05-02 13:20:26.059 JupyterHub reflector:237] Error when calling Kubernetes API. Status: 403 Forbidden. Message: pods is forbidden: User "system:serviceaccount:default:jupyterhub-spawner-sa" cannot list resource "pods" in API group "" in the namespace "user-jhu"
[E 2024-05-02 13:20:26.059 JupyterHub reflector:412] Initial list of pods failed
      File "/opt/kubespawner/kubespawner/reflector.py", line 410, in start
        await self._list_and_update()
      File "/opt/kubespawner/kubespawner/reflector.py", line 240, in _list_and_update
        raise client.ApiException(
    kubernetes_asyncio.client.exceptions.ApiException: (403)
    Reason: Forbidden
    
    
[E 2024-05-02 13:20:26.059 JupyterHub spawner:2411] Reflector with key ('pods', None) failed to start.
      File "/opt/kubespawner/kubespawner/spawner.py", line 2409, in catch_reflector_start
        await func
      File "/opt/kubespawner/kubespawner/reflector.py", line 410, in start
        await self._list_and_update()
      File "/opt/kubespawner/kubespawner/reflector.py", line 240, in _list_and_update
        raise client.ApiException(
    kubernetes_asyncio.client.exceptions.ApiException: (403)
    Reason: Forbidden
```

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -5,7 +5,7 @@ import json
 import time
 from functools import partial
 
-from kubernetes_asyncio import watch, client
+from kubernetes_asyncio import client, watch
 from traitlets import Any, Bool, Dict, Int, Unicode
 from traitlets.config import LoggingConfigurable
 from urllib3.exceptions import ReadTimeoutError
@@ -231,12 +231,13 @@ class ResourceReflector(LoggingConfigurable):
 
         initial_resources_raw = await list_method(**kwargs)
         if not initial_resources_raw.ok:
-            self.log.error(f'Error when calling Kubernetes API.'
+            self.log.error(
+                f'Error when calling Kubernetes API.'
                 f' Status: {initial_resources_raw.status} {initial_resources_raw.reason}.'
-                f' Message: {(await initial_resources_raw.json())["message"]}')
+                f' Message: {(await initial_resources_raw.json())["message"]}'
+            )
             raise client.ApiException(
-                status=initial_resources_raw.status,
-                reason=initial_resources_raw.reason
+                status=initial_resources_raw.status, reason=initial_resources_raw.reason
             )
 
         # This is an atomic operation on the dictionary!
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -234,12 +234,14 @@ class ResourceReflector(LoggingConfigurable):
             if not initial_resources_raw.ok:
                 raise client.ApiException(
                     status=initial_resources_raw.status,
-                    reason=initial_resources_raw.reason
+                    reason=initial_resources_raw.reason,
                 )
         except client.ApiException:
-            self.log.exception(f'An error occurred when calling Kubernetes API.'
+            self.log.exception(
+                f'An error occurred when calling Kubernetes API.'
                 f' Status: {initial_resources_raw.status} {initial_resources_raw.reason}.'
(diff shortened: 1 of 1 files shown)


kubespawner/reflector.py
```diff
@@ -228,7 +228,18 @@ async def _list_and_update(self, resource_version=None):
             kwargs["namespace"] = self.namespace
 
         list_method = getattr(self.api, self.list_method_name)
+
         initial_resources_raw = await list_method(**kwargs)
+        if not initial_resources_raw.ok:
+            self.log.error(
```
yuvipanda, line 234: Can you use `self.log.exception` here instead, and pass it the exception you're creating?

josefhandl: Yes, but as far as I know, it's not possible without a try-catch. I've pushed a fix

consideRatio:
@yuvipanda I tried catching up with differences of log.error and log.exception, but I've not come to a conclusion if 0f2555d was what you wanted - could you review this again?

kubespawner/reflector.py
consideRatio: Based on [python docs](https://docs.python.org/3/library/logging.html#logging.Logger.exception):

> This method should only be called from an exception handler.

So I think you were forced to do that for this to make sense. But, doesn't this mean we get stack trace details from raising the error a few lines above, and that in turn doesn't help us get informed about the original error anyhow associated with getting a response where `response.ok == False`?

consideRatio: With that in mind, was the original thing better or not? I'm not sure - I'll defer to @yuvipanda to make a decision and then we go for it.

yuvipanda:
I think either works, and given that the ultimate goal is to figure out better error handling, I'm just going to merge this.

Thanks for your patience @josefhandl and thanks for the ping, @consideRatio

PR #841, opened 2024-06-12
update images used in tests

### LINKED ISSUE #833: Are we testing against JupyterHub 1.3?
It seems we have fixture that creates a jupyterhub pod with the version hardcoded to jupyterhub 1.3.

We should make our tests test against the version of jupyterhub we expect it to be tested against.

https://github.com/jupyterhub/kubespawner/blob/127647ce9f25c75fb2acbea43e1c39fe11fdce7a/tests/conftest.py#L414-L423

https://github.com/jupyterhub/kubespawner/blob/127647ce9f25c75fb2acbea43e1c39fe11fdce7a/.github/workflows/test.yaml#L40-L61

minrk:
they were very out of date

for whatever reason, I couldn't run the tests locally with the old images, but they run fine with these

closes #833

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -420,7 +420,7 @@ async def create_hub_pod(kube_client, kube_ns, pod_name="hub", ssl=False):
             volumes=volumes,
             containers=[
                 {
-                    "image": "jupyterhub/jupyterhub:1.3",
+                    "image": "quay.io/jupyterhub/jupyterhub:5.0",
                     "name": "hub",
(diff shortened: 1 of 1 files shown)


consideRatio:
Nice, closes #833 perhaps?

consideRatio (approved):

minrk:
@consideRatio good point! We don't have upper bound on the hub and `:latest` is the default tag in KubeSpawner.image, so latest is probably the most consistent choice with versions coming from elsewhere.

PR #843, opened 2024-07-15
Allow volumes and volume mounts to be defined as Dicts

sunu:
I've made the following changes:

- Modified the trait type for volumes and volume mounts to `Union(trait_types=[List(), Dict()])` so that they can be defined as dictionaries (for easy overriding) or as lists (for backwards compatibility)
- Updated the help messages explaining both list and dictionary input options and clarifying how dictionary inputs are processed (sorted lexicographically by keys)
- Added tests for defining volumes and volume mounts as dictionaries and test that the output is sorted lexicographically and tests for defining volumes and volume mounts as lists to make sure these changes are backwards compatible

Defining these values as dictionaries enables easy overriding. refs https://github.com/jupyterhub/jupyterhub/pull/4822 and https://github.com/NASA-IMPACT/veda-jupyterhub/issues/44

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2061,7 +2061,9 @@ class KubeSpawner(Spawner):
             pod_security_context=psc,
             env=self.get_env(),  # Expansion is handled by get_env
             volumes=self._expand_all(self._sorted_dict_values(self.volumes)),
-            volume_mounts=self._expand_all(self._sorted_dict_values(self.volume_mounts)),
+            volume_mounts=self._expand_all(
+                self._sorted_dict_values(self.volume_mounts)
+            ),
             working_dir=self.working_dir,
             labels=labels,
             annotations=annotations,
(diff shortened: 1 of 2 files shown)


sunu:
@yuvipanda Could you please take a look at these changes and let me know if they look good? If this looks ok, I'll proceed to make similar updates for other traitlets that need to support dictionaries. Thanks!

kubespawner/spawner.py
```diff
@@ -1869,6 +1891,12 @@ def _expand_all(self, src):
         else:
             return src
 
+    def _sorted_dict_values(self, src):
```
yuvipanda, line 1894: Can you add a docstring here?

yuvipanda (approved):
One minor ask for a docstring, but otherwise this approach looks great to me.

Thank you for working on this, @sunu!

kubespawner/spawner.py
sunu: Done!

sunu:
Thanks for the review @yuvipanda! I have added the missing docstring. I'll open a follow up PR to make similar updates for other traitlets that need to support dictionaries.

yuvipanda:
Thanks @sunu!

PR #853, opened 2024-09-22
Refresh sphinx config file and resolve failure in sphinx 8

consideRatio:
We had docs build failures due to a legacy line in this config file. I went ahead and refreshed it to look more like other jupyterhub project's sphinx config file, including adding rediraffe and intersphinx setup to make it easier to setup redirects and reference JupyterHub's docs.


consideRatio:
Looks right, going for a merge.

PR #856, opened 2024-10-02
fix some safe slug patterns

minrk:
- escape `-` in label pattern. Lacking this meant `\.-_` was the _range_ which included `@`, causing emails to not get escaped
- `is_valid_object_name` now produces a valid object name for [all cases](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), adding restrictions:
  - start with letter (numbers not allowed)
  - don't allow '.'
  - length 63
- `_extract_safe_name` now satisfies the "starts with a letter" rule for RFC1035 label names

improved test coverage for several affected cases


consideRatio (approved):
Thank you, seems ok to merge - is it ready to merge?

minrk:
Yeah, I think so. Can do another quick beta after merging

consideRatio:
Thank you for working this so thoroughly Min!!

consideRatio:
I realized just after merge, is docs updates relevant also?

minrk:
> I realized just after merge, is docs updates relevant also?

No, the docs don't go into the details of exactly what the patterns are. This makes the docs accurate, where some escapes were not applied when they should be.

Issue #858, opened 2024-10-08
Should we always await asyncio.ensure_future()?


manics:
### Bug description

There are somer calls to `asyncio.ensure_future(...)` that aren't awaited:
https://github.com/jupyterhub/kubespawner/blob/621be0e68d9dea38ac899c6054f966dd68d1d028/kubespawner/spawner.py#L2803-L2805
https://github.com/jupyterhub/kubespawner/blob/621be0e68d9dea38ac899c6054f966dd68d1d028/kubespawner/spawner.py#L3252-L3257
https://github.com/jupyterhub/kubespawner/blob/621be0e68d9dea38ac899c6054f966dd68d1d028/kubespawner/spawner.py#L3379-L3382

The [docs for `asyncio.ensure_future`](https://docs.python.org/3.12/library/asyncio-future.html#asyncio.ensure_future) say the resulting Task will be scheduled, but also says
> Save a reference to the result of this function, to avoid a task disappearing mid-execution.

[`asyncio.create_task()` has a longer explanation](https://docs.python.org/3.12/library/asyncio-task.html#asyncio.create_task)
> Important. Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. 

I think we should either await those, save a reference to them, or add a comment to the code explaining why it isn't necessary.

manics:
There's also some calls in proxy.py
https://github.com/jupyterhub/kubespawner/blob/7ba57a8c915b190b24d514d91427620c08eb1353/kubespawner/proxy.py#L358-L360

consideRatio:
@manics thanks for investigating this thoroughly -- I agree on the decision to await things now that I've seen the motivation in this issue!

PR #859, opened 2024-10-09
Reflectors watch: close after stop

manics:
I'm trying to track down the
```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7fcfe6bbc690>
```
warning in https://github.com/lsc-sde/lsc-sde/issues/93#issuecomment-2399981874

I don't know if this is the cause, but based on the context manager code in
https://github.com/tomplus/kubernetes_asyncio/blob/v21.7.1/kubernetes_asyncio/watch/watch.py#L212-L216
we should call `close()` anyway.

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -389,6 +389,7 @@ class ResourceReflector(LoggingConfigurable):
                 self.log.debug("%s watcher timeout", self.kind)
             finally:
                 w.stop()
+                await w.close()
                 if self._stopping:
                     self.log.info("%s watcher stopped: outer", self.kind)
                     break


consideRatio (approved):
I think this is correct and suitable to do, but I struggle to become confident about it.

Do you want to merge or ask for another set of eyes?

consideRatio:
By awaiting close, we'll have a delay of finalizing the `_watch_and_update` function call.

The `_watch_and_update` is called via `self.watch_task = asyncio.create_task(self._watch_and_update())`, and then `self.watch_task` is checked to be `.done()` or not at several locations.

The thing I'm not getting confident about, is the consequences ofnow start waiting for that to finalize. I figure at worst we get a performance hit, but I think we'll end up with something more robust. Due to that, I'm leaning towards we go for a merge.

manics:
Thanks for reviewing, let's merge then!

I want to look into https://github.com/jupyterhub/kubespawner/issues/858 , maybe we can have a beta-3 after that

consideRatio:
Thank you for working this @manics!!

PR #864, opened 2024-10-14
CI: Cleanup namespaces in parallel

manics:
Currently if deletion of one namespace fails during cleanup the other namespace won't be deleted. Using asyncio.gather means we can delete in parallel but still receive the first Exception

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -236,6 +236,21 @@ async def watch_kubernetes(kube_client, kube_ns):
         raise exc
 
 
+async def _delete_namespace(client, namespace):
+    await client.delete_namespace(namespace, body={}, grace_period_seconds=0)
+    for _ in range(20):  # Usually finishes a good deal faster
+        try:
+            await client.read_namespace(namespace)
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -236,6 +236,22 @@ async def watch_kubernetes(kube_client, kube_ns):
...
+            await client.read_namespace(namespace)
+        except ApiException as e:
+            if e.status == 404:
+                return
+            else:
+                raise
+        else:
+            print("waiting for %s to delete" % namespace)
+            await asyncio.sleep(1)
+    raise Exception(f"Namespace {namespace} not deleted after 20 s")
```
manics, line 252: This line is new. Previously this function would wait up to 20s but not do anything if the namespace hadn't been deleted. Is that the behaviour we desire?

consideRatio (approved):

tests/conftest.py
consideRatio: What you did now is better!

PR #865, opened 2024-10-18
Add escape_slug as a utility function alongside safe_slug

consideRatio:
By providing this, it becomes easier to test and compare slug changes.

Related to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3532.


consideRatio:
Added the `new` label, happy to make it `maintenance` or similar as well.

PR #869, opened 2024-10-24
Reset cur_delay after listing resources

### LINKED ISSUE #868: Intermittent Kubernetes API server errors
### Bug description

Hello, and thanks everyone for great work on JupyterHub :rocket:

We are noticing an issue with the reflector when the Kubernetes API server gives 504 responses. The hub operates smoothly, but once in a while the following scenario happens:

On this [line](https://github.com/jupyterhub/kubespawner/blob/269f51a694928ba6ea96b25bffa1544737104041/kubespawner/reflector.py#L231) there is an error in the API server, so `initial_resources_raw` become:

```
[D 2024-10-23 11:18:38.042 JupyterHub reflector:233] <ClientResponse(https://10.0.0.1:443/api/v1/namespaces/NAMESPACE/pods?fieldSelector=&labelSelector=component%3Dsingleuser-server&resourceVersion=97691474&resourceVersionMatch=NotOlderThan) [504 Gateway Timeout]>
  <CIMultiDictProxy('Audit-Id': 'c9a5f3c5-c636-4f6f-86ad-80b1d8149d61', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Retry-After': '1', 'X-Kubernetes-Pf-Flowschema-Uid': 'ec263431-93d6-4a38-afa2-1cd62c7d0380', 'X-Kubernetes-Pf-Prioritylevel-Uid': '2663c056-d0ae-405b-84c4-aa8fc7327b62', 'Date': 'Wed, 23 Oct 2024 11:18:38 GMT', 'Content-Length': '297')>
  ```

Then [here](https://github.com/jupyterhub/kubespawner/blob/269f51a694928ba6ea96b25bffa1544737104041/kubespawner/reflector.py#L233) `initial_resources` are:
```
[D 2024-10-23 11:18:38.042 JupyterHub reflector:238] {'kind': 'Status', 'apiVersion': 'v1', 'metadata': {}, 'status': 'Failure', 'message': 'Timeout: Too large resource version: 97691474, current: 97690604', 'reason': 'Timeout', 'details': {'causes': [{'reason': 'ResourceVersionTooLarge', 'message': 'Too large resource version'}], 'retryAfterSeconds': 1}, 'code': 504}
```

Causing an error [here](https://github.com/jupyterhub/kubespawner/blob/269f51a694928ba6ea96b25bffa1544737104041/kubespawner/reflector.py#L234):
```
[E 2024-10-23 11:18:38.042 JupyterHub reflector:382] 'items'
      File "/usr/local/lib/python3.8/site-packages/kubespawner/reflector.py", line 300, in _watch_and_update
        resource_version = await self._list_and_update(resource_version)
      File "/usr/local/lib/python3.8/site-packages/kubespawner/reflector.py", line 242, in _list_and_update
        for p in initial_resources["items"]
    KeyError: 'items'
```

The exception is handled [here](https://github.com/jupyterhub/kubespawner/blob/8706bf5f115b0f9ef5116c94570f7180ad749bdc/kubespawner/reflector.py#L372-L386) and it's finally causing the Hub pod to break and log `Watching resources never recovered, giving up`

Then the Hub pod has to be restarted, and it only runs until the same scenario recurs.

After the error happens, and the loop retries [here](https://github.com/jupyterhub/kubespawner/blob/8706bf5f115b0f9ef5116c94570f7180ad749bdc/kubespawner/reflector.py#L308), pods are successfully listed.

However, unless an event happens [here](https://github.com/jupyterhub/kubespawner/blob/269f51a694928ba6ea96b25bffa1544737104041/kubespawner/reflector.py#L314-L324), `cur_delay` is never reset.
It will keep increasing with every new error from the API server, eventually accumulating enough to cause the Hub to break.

This may be an edge case scenario, but has been happening to us multiple times.

The fix was to reset `cur_delay` to 0.1 after the command on [this line](https://github.com/jupyterhub/kubespawner/blob/8706bf5f115b0f9ef5116c94570f7180ad749bdc/kubespawner/reflector.py#L308).
```
try:
    resource_version = await self._list_and_update(resource_version)
    cur_delay = 0.1
...
```

Should I create a PR? Is there a reason why this fix won't work?

### How to reproduce

We weren't able to reproduce on demand, this only happened when the Kubernetes API server gave 504 responses.

#### Expected behaviour

Even if the Kubernetes API server fails once in a while, and there are no events for pods, the reflector should continue running.

#### Actual behaviour

If there are Kubernetes API server errors and no pod events, the reflector stops and it's not possible for users to create new servers.

### Your personal set up

KubeSpawner 6.2.0
JupyterHub 5.0.0

### Additional

In the main branch, there is a [try-except](https://github.com/jupyterhub/kubespawner/blob/452039e5f0da755957ab60758ce9d791ca8d8bb0/kubespawner/reflector.py#L232-L245) for the Kubernetes API server errors, but `cur_delay` still is not reset unless there are events.

d-gol:
Reset `cur_delay` when resources are listed, so that it doesn't keep increasing if there are no events.
Detailed info: https://github.com/jupyterhub/kubespawner/issues/868

Edit: fixes #868

### DIFF
--- a/kubespawner/reflector.py
+++ b/kubespawner/reflector.py
@@ -306,6 +306,7 @@ class ResourceReflector(LoggingConfigurable):
             w = watch.Watch()
             try:
                 resource_version = await self._list_and_update(resource_version)
+                cur_delay = 0.1
                 watch_args = {
                     "label_selector": self.label_selector,
                     "field_selector": self.field_selector,


manics (approved):
Thanks!

PR #873, opened 2024-11-15
Delete named PVCs if they contain `{user_server}`

manics:
`pod_name_template` was updated in 7.0.0, but the logic to detect whether a named server PVC is safe to delete wasn't.

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3807,7 +3807,7 @@ class KubeSpawner(Spawner):
             self.log.info(f"Not deleting pvc for {log_name}: {self.pvc_name}")
             return
 
-        if self.name and '{servername}' not in self.pvc_name_template:
+        if self.name and '{user_server}' not in self.pvc_name_template:
             # named server has the same PVC as the default server
             # don't delete the default server's PVC!
             self.log.info(


consideRatio (approved):
Thank you Simon!!

kubespawner/spawner.py
```diff
@@ -3807,7 +3807,7 @@ async def delete_forever(self):
             self.log.info(f"Not deleting pvc for {log_name}: {self.pvc_name}")
             return
 
-        if self.name and '{servername}' not in self.pvc_name_template:
+        if self.name and '{user_server}' not in self.pvc_name_template:
```
consideRatio, line 3810: I'm not confident about what makes sense here, it seems like we want to ensure that we don't delete a named servers PVC if its shared. Due to that, should we maybe instead check for either servername and user_server?

Is servername still accepted, or was it removed entirely?

manics: The check is inverted so we're only deleting if `pvc_name_template` contains `{user_server}`.

We could also delete the PVC if the template contains `{servername}` as in KubeSpawner 6 but strictly speaking this isn't completely safe, since if (for some unknown reason) an admin wanted all users to share hte same volume based on servername we'd incorrectly delete the PVC.

There's also the problem of future maintenance- the more conditions we add the more likely it is to break in future.

consideRatio: Good thinking, thank you for being so thorough @manics!!

Issue #839, opened 2024-05-07
Bring into line with the Kubernetes operator design pattern


qcaas-nhs-sjt:
### Proposed change
At present, the kubespawner is self contained making the hub actually being performed by a single service that is extended by individual organisations to cover off a variety of different solutions. While this does make it easier in some ways, it does not make a closed and easily extendable solution that can meet the needs of organisations without a lot of fudging, and it also adds in a number of issues which could be exploited should the service ever be breached. For one the same service that the user interacts with can effectively make changes to pod definitions and therefore run whatever workloads they want. This could then also potentially give the user the ability to create other breaches which would undermine the solution. 

### Alternative options
We can of course continue to operate on the current solution, however I feel that this is risky as people add more and more functionality to the existing hub it will be much harder to split the solution out as the solution continues to grow. 

The proposed solution gives us a microservice implementation and we can extend by either:
* extending the operator
* creating additional operators

It also could potentially allow organisations to build their own microservices on other code bases and languages if needed.

### Who would use this feature?
* **Developers** and **System Administrators** would benefit as it is much easier to develop microservices based on a template and add your own logic than it is to build custom python files in the current manner. Kubernetes developers are also usually familiar with the operator model. This also has benefits when debugging as they would be able to track the assets through the process.
* **Extension Builders** would benefit as it would be easier to extend this solution by using additional microservices and controllers in some instances. Say for example we wanted to build a multi-cluster implementation we could add in some additional services to handle this securely.
* **IT Security officers** would benefit as they could see appropriate isolation of services and validation can easily be implemented that can enforce that the operator will only provision images and workloads that organisations want to be provisioned, rather than anything.

### Suggest a solution
I propose that we break the hub down in kubernetes into multiple services and utilise the kubernetes operator design pattern and implement a more refined security policy for these services. As we use the solution in secure data environments and trusted research environments it is essential that this is secure and there are other benefits such as scalability and extensibility that could be advantageous as a result of the new model.

To avoid this being a breaking change in existing applications, we could split the functionality of the current kubespawner into multiple modules that could be imported into the core kubespawner project. We could then have a feature flag which would then allow the hub to use the legacy framework or the new one. Those wanting to use the old framework could continue to do so, but these would remain offline until the organisation is ready to migrate to the new one. This would also allow us to provide bug fixes and so on to both pathways at the same time.

```mermaid
sequenceDiagram
    participant Hub
    participant API as Kubernetes Operator
    participant Operator
    
[...]

PR #885, opened 2025-03-25
Added pod template support for the single user

sanjog152:


manics:
Hi! Please can you provide some background to this change? My initial thoughts are this is outside the scope of KubeSpawner since it pretty much overrides everything, and would be more appropriate in a separate package.

See for example https://github.com/jupyterhub/kubespawner/issues/839

PR #887, opened 2025-04-07
Unconditionally put servername in annotations

yuvipanda:
- Matches behavior of what we do with servername in labels
- Labels are escaped while annotations are not, making annotations suitable for filtering in metrics and dashboards

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2245,6 +2245,7 @@ class KubeSpawner(Spawner):
         # Annotations don't need to be escaped
         annotations = {
             'hub.jupyter.org/username': self.user.name,
+            'hub.jupyter.org/servername': self.name,
             "hub.jupyter.org/kubespawner-version": __version__,
             "hub.jupyter.org/jupyterhub-version": jupyterhub.__version__,
         }
@@ -2377,8 +2378,6 @@ class KubeSpawner(Spawner):
         annotations = self._build_common_annotations(
             self._expand_all(self.extra_annotations)
         )
-        # Unconditionally put servername on the pods
-        annotations['hub.jupyter.org/servername'] = self.name
 
         return make_pod(
             name=self.pod_name,
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2243,11 +2243,12 @@ class KubeSpawner(Spawner):
 
     def _build_common_annotations(self, extra_annotations):
         # Annotations don't need to be escaped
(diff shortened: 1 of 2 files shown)


minrk (approved):

kubespawner/spawner.py
```diff
@@ -2285,6 +2285,8 @@ async def get_pod_manifest(self):
         annotations = self._build_common_annotations(
             self._expand_all(self.extra_annotations)
         )
+        # Unconditionally put servername on the pods
+        annotations['hub.jupyter.org/servername'] = self.name
```
consideRatio, line 2289: It looks like the unconditional assignment following right after calling _build_common_annotations should only be relevant if self.extra_annotations were declaring another `hub.jupyter.org/servername`, which would have a blank string value or similar.

```python
        annotations = self._build_common_annotations(
            self._expand_all(self.extra_annotations)
        )
        # Unconditionally put servername on the pods
        annotations['hub.jupyter.org/servername'] = self.name
```

I wonder if its really a kubespawner issue, or z2jh / end-user issue based on this.

kubespawner/spawner.py
```diff
@@ -2166,11 +2166,11 @@ def _build_pod_labels(self, extra_labels):
...
-        annotations = {'hub.jupyter.org/username': self.user.name}
-        if self.name:
-            annotations['hub.jupyter.org/servername'] = self.name
-        annotations["hub.jupyter.org/kubespawner-version"] = __version__
-        annotations["hub.jupyter.org/jupyterhub-version"] = jupyterhub.__version__
+        annotations = {
+            'hub.jupyter.org/username': self.user.name,
+            "hub.jupyter.org/kubespawner-version": __version__,
+            "hub.jupyter.org/jupyterhub-version": jupyterhub.__version__,
+        }
```
consideRatio, line 2173: With this change we wouldn't get the servername annotation on PVCs etc

minrk: While addressing the issue of PVCs is potentially relevant, I don't think it was the goal of this PR. I think it was only consistency with ensuring servername is always set, rather than only set when the name is non-default.

I've updated the PR so the only change is what it says on the title: servername annotation is unconditional, rather than changing which objects the servername annotation is applied to.

@manics mentioned possible safety issues with this PR before, which was the case when it was _removing_ this annotation for named servers from resources other than pods. But as it is now, the only change is setting the value consistently even when `self.name` is an empty string. No other changes.

yuvipanda:
Ty for picking this up, @minrk!

the test failures seem real and probably mean the tests need updating

PR #900, opened 2025-10-07
Add KubeSpawner.add_custom_event

### LINKED ISSUE #899: System for displaying internal spawn messages outside of K8s events
### Proposed change

Include a more streamlined way to add additional events manually in functions like hooks or other overloaded functions. Some of these functions might have code running that fails/succeeds in ways that might be relevant for the user and there currently is no easy way to emit these messages.

### Alternative options

The current alternative would be manually creating K8s events from the code, so they get picked up again by the events property and then yielded by the progress property.

### Who would use this feature?

Any setup that modifies the pods or has additional steps before spawning could benefit from giving additional information to the user in a streamlined way. As an example, in our setup, the image repository (harbor) scans user-built (binderhub) images for vulnerabilities and only then gives them up for pulling. This feature would make it simpler to inform the user of what's going on, or what went wrong (if the image were to be flagged as vulnerable).

### (Optional): Suggest a solution

Add a way of including messages in the format of what is yielded in ´progress´ that you can emit easily from within the spawner code using a spawner method (e.g. `emit_message` or similar).

I can create a PR for this myself, I just wanted to see if there is any interest in this feature first.

rtmiz:
Add: add_custom_event method, custom_event_queue trait, 
Update: events property to handle custom events

Adjust KubeSpawner class so you can easily include custom messages for the user during the spawning process, for instance like so (When using hooks inside of a jupyterhub config):
```
import time
from datetime import datetime, timezone

def modify_pod_hook(spawner, pod):
    spawner.add_custom_event(
        eventTime=str(datetime.now(timezone.utc)),
        lastTimestamp
        message="Test event. This is a test event inside of modify_pod_hook.",
        type="Custom",
        involvedObject: {"name": spawner.pod_name},
        metadata: {"uid": str(time.time())}
    )
    return pod

```

Closes #899

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -317,6 +317,14 @@ class KubeSpawner(Spawner):
         """,
     )
 
+    _sent_custom_events = List(
+        [],
+        config=False,
+        help="""
+        List of (index, event) tuples of custom events that have been sent to the user.
+        """,
+    )
+
     enable_user_namespaces = Bool(
         False,
         config=True,
@@ -2669,7 +2677,9 @@ class KubeSpawner(Spawner):
             return []
 
         events = []
+
         for event in self.event_reflector.events:
+
             if event["involvedObject"]["name"] != self.pod_name:
                 # only consider events for my pod name
                 continue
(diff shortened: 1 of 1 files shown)


minrk:
Sorry for leaving this for so long. My only question is whether every field should be required for custom events, or if some should have default values filled in, so e.g. maybe only eventTime and message are required (e.g. default involvedObject is self.pod)?

rtmiz:
@minrk easiest solution would probably be to fall back to defaults like the current time for timestamps, {"name": self.pod_name} for involvedObject (name is the only relevant field, as it is checked in the events property).

I've added these in the latest commit, so now you can call a add_custom_event with only the message.

minrk:
Thanks! Ignoring test failures in with dev JupyterHub, which I'll take care of in its own PR.

PR #903, opened 2025-11-25
Load port from named notebook-port containerPort

### LINKED ISSUE #827: JupyterHub Singleuser Port Routing Issue After Restart
### Bug description

When customizing the singleuser port (e.g., setting it to 9999), the routing works correctly, directing traffic to port 9999. However, after a JupyterHub restart, the routing becomes erroneous and redirects to port 8888 instead.

### How to reproduce

1. use a custom singleuser port (e.g., 9999) in jupyterhub.
2. Start Jupyter notebook and verify that the routing works as expected.
3. Restart JupyterHub.
4. Observe that the routing now incorrectly directs to port 8888 instead of the custom port (9999).

#### Expected behaviour

After a JupyterHub restart, the routing should continue to direct traffic to the custom singleuser port specified during configuration.

#### Actual behaviour

After a JupyterHub restart, the routing that was previously working correctly (directing traffic to the custom singleuser port, e.g., 9999) becomes erroneous. Instead of continuing to route to the specified custom port, it now incorrectly redirects to port 8888.

### Your personal set up

- Deployment Method: zero-to-jupyterhub-k8s  
- Chart name: jupyterhub  
- Chart version: 3.2.1  
- Chart config: singleuser.profileList[0].port=9999

### Others
1. **Observation**:
   - After starting a JupyterHub singleuser pod, the database (`servers` table) correctly records a port value of 9999.
   - However, upon restarting JupyterHub, this port value is changed to 8888 in the database.
   - Simultaneously, the `configurable-http-proxy` receives a request to modify the routing to port 8888.
   - Eventually, JupyterHub becomes unable to connect to the singleuser pod via port 8888, leading to the cleaning of the singleuser pod.

2. **Hypothesis**:
   - The bug seems to occur during JupyterHub's initial startup, where it incorrectly resets all singleuser routes to port 8888 without referencing the port settings stored in the database.

minrk:
avoids changes to self.port (e.g. dynamic or changed config) being interpreted as a changed URL for a running pod

closes #893 

closes #827

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -2201,6 +2201,19 @@ class KubeSpawner(Spawner):
         """,
     )
 
+    def _get_pod_port(self, pod):
+        """
+        Return the port of the server in the pod.
+
+        That port must be called 'notebook-port'.
+        """
+        for container in pod["spec"]["containers"]:
+            for port in container["ports"]:
+                if port.get("name") == "notebook-port":
+                    return port["containerPort"]
+        pod_name = pod["metadata"]["name"]
+        raise KeyError(f"No port 'notebook-port' in pod {pod_name}")
+
     def _get_pod_url(self, pod):
         """Return the pod url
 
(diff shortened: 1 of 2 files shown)


manics (commented):
Did you consider getting the port from the podspec since it's named?
https://github.com/jupyterhub/kubespawner/blob/5dcfbeb3a5f7b36cd0d5fa54a0ab153ea83b121c/kubespawner/objects.py#L538

Or do you still think an annotation is best?

minrk:
> Did you consider getting the port from the podspec since it's named?

I did not, good call!

manics:
```
FAILED tests/test_spawner.py::test_ipv6_addr - KeyError: 'spec'
```
Most likely because this test uses a mock
https://github.com/minrk/kubespawner/blob/2db388cd82194bef10bd337c9722ffd908673abf/tests/test_spawner.py#L1929-L1931

minrk:
test mock updated and passes

manics (approved):

PR #906, opened 2025-12-04
Fix input field for unlisted choice to be only display when selected

### LINKED ISSUE #892: Input field for unlisted choice is display even if not selected
### Bug description
When configuring a profile option containing an "unlisted choice" the input field to enter the "unlisted choice" is always displayed, even if a predefined choice is selected. 

### How to reproduce

configure profile_options like this:
```yaml
profile_options:
  image:
    display_name: "Image"
    choices:
      0:
        display_name: "Image 1"
        kubespawner_override:
          image: registry.com/image1
      1:
        display_name: "Image 2"
        kubespawner_override:
          image: registry.com/image2
    unlisted_choice:
      enabled: true
      display_name: "Custom image"
      display_name_in_choices: "Enter custom image"
      kubespawner_override:
        image: '{value}'
```

#### Expected behaviour
The input field for the custom image should only be displayed when selecting the option "Enter custom image".

#### Actual behaviour
The input field for the custom image is always displayed.

Image

---

Image

### Your personal set up

Deployment:
- [Zero 2 Jupyterhub Helm Chart](https://z2jh.jupyter.org/en/stable/jupyterhub/installation.html#install-jupyterhub): Version 4.2.0

Versions:
- jupyterhub: 5.3.0
- jupyterhub-kubespawner: 7.0.0

### Bug investigation
The bug is not yet present in jupyterhub 4.1.6 combined with jupyterhub-kubespawner 6.2.0. 
When compared the CSS class `.hidden` which is used to hide the input field of the unlisted choice (check [here](https://github.com/jupyterhub/kubespawner/blob/main/kubespawner/templates/form.html#L35)) is different in both jupyterhub versions.

In JupyterHub 4.1.6:

```css
.hidden {
    display: none !important;
}
```

In Jupyterhub 5.3.0

```css
.hidden {
    display: none;
}
```

fpfuetsch:
Fixes #892 using the proposed code changes from @iamvikaskumar, which were successfully tested with JupyterHub 5.3.0.

### DIFF
--- a/kubespawner/templates/style.css
+++ b/kubespawner/templates/style.css
@@ -27,3 +27,9 @@
   margin-right: 8px;
   min-width: 96px;
 }
+
+/* Make .hidden win over the .option flex rule */
+#kubespawner-profiles-list .profile .option.hidden,
+#kubespawner-profiles-list .js-other-input-container.hidden {
+  display: none;
+}


minrk:
Nice, thanks!

PR #910, opened 2026-03-03
Add hook for formatting Kubernetes Event messages

agoose77:
## TL;DR

> [!Note] 
> No LLMs were used in the authoring of this PR.

2i2c is currently working on [a user-story](https://github.com/2i2c-org/infrastructure/issues/7718) to improve the kubespawner progress messages, as part of [an initiative](https://github.com/2i2c-org/infrastructure/issues/7160) to improve the spawn-progress page. 

This PR does several things:

1. Adds a custom `decorate_progress_message` that overrides the pretty-printing of event messages.
2. Adds a `kubespawner.events` module for richer built-in formatting of log messages.
3. Adds a `kubespawner.events.RuleEventFormatter` and other types for defining event formatting rules.

See the before and after:

**Before**
old

**After**
new

## Goal
The goal is modest: to improve the human readability of spawn messages, and to allow further customisation.

## Example Decoration Hook
**Basic hook**
```python
def decorate_progress_message(spawner, event, text):
    return { 
        "message": f"custom-message-{text}",
        "html_message": f"<span>{text}</span>"
    }
c.KubeSpawner.decorate_progress_message = decorate_progress_message
```

**Use the rules to define custom renderers**

```python
c.RuleEventFormatter.rules = {
    "01-container-image-events": {
        "match": {
            "reportingComponent": r"kubelet",
            "fieldPath": r"spec\.(?P<container>initContainers|containers)\{([^}]+)\}",
            "reason": r"(?P<action>Pulling|Pulled)",
        },
        "template": "{action} {image} image for the {container} container",
    },
    "02-container-lifecycle-events": {
        "match": {
            "reportingComponent": r"kubelet",
            "fieldPath": r"spec\.(?P<container>initContainers|containers)\{([^}]+)\}",
            "reason": r"(?P<action>Started|Killing|Created|Stopped)",
        },
        "template": "{action} the {container} container",
    },
    "03-pod-resource-events": {
        "match": {
            "reportingComponent": r"kubelet",
            "reason": r"OutOf(?P<resource>memory|cpu|ephemeral-storage|pods)",
        },
        "template": "The node selected to run your server ran out of {resource}",
    },
    "04-scheduler-node-found-events": {
        "match": {
            "reportingComponent": r".*-user-scheduler",
            "reason": r"Scheduled",
            "message": r".*?assigned \S+ to (?P<node>\S+)",
        },
        "template": "A node ({node}) has been found to run your server",
    },
    "05-scheduler-no-nodes-events": {
        "match": {
            "reportingComponent": r".*-user-scheduler",
            "reason": r"FailedScheduling",
        },
        "template": "No existing nodes are currently able to run your server",
    },
    "06-cluster-autoscaler-events": {
        "match": {
            "reportingComponent": r"cluster-autoscaler",
            "reason": r"TriggeredScaleUp",
        },
        "template": "Launching new nodes by scaling up the cluster",
    },
    "07-node-affinity-events": {
        "match": {
            "reportingComponent": r"kubelet",
            "message": r"Predicate NodeAffinity failed.*",
            "reason": r"NodeAffinity",
        },
        "template": "It was not possible to find or launch any nodes to run your server. This is likely due to a configuration problem with the infrastructure or the JupyterHub",
    },
    "08-gke-scheduler-node-found-events": {
        "match": {
            "reportingComponent": r"gke\.io/optimize-utilization-scheduler",
            "reason": r"Scheduled",
            "message": r".*?assigned \S+ to (?P<node>\S+)",
        },
        "template": "A node ({node}) has been found to run your server",
    },
    "09-gke-scheduler-no-nodes-events": {
        "match": {
            "reportingComponent": r"gke\.io/optimize-utilization-scheduler",
            "reason": r"FailedScheduling",
        },
        "template": "No existing nodes are currently able to run your server",
    },
    "10-taint-eviction-events": {
        "match": {
            "reportingComponent": r"taint-eviction-controller",
            "reason": r"gke\.io/optimize-utilization-scheduler",
            "message": r"Cancelling deletion of Pod.*",
        },
        "template": "Cancelling deletion of your server. This normally happens when a scale-up has just taken place.",
    },
}
```

### UI

- Timestamps are formatted to regular isoformat-like `%Y-%m-%DTHH:MM:SSZ` to keep fixed width
- Timestamps and message types are pretty formatted as button-pills
- Messages are simplified where possible

### Constraints
- I targeted Python 3.7 given `pyproject.toml`, meaning no `match`, `:=`, or `removeprefix`.

## Questions
- Is Kubespawner making too many assumptions if we bake-in the expectation of Bootstrap?
- Could we consider adding a start-time timestamp so that times can simply be given as "minutes since spawn" rather than UTC times?


kubespawner/spawner.py
```diff
@@ -2765,15 +2785,17 @@ async def progress(self):
                     # 30 50 63 72 78 82 84 86 87 88 88 89
                     progress += (90 - progress) / 3
 
+                    if self.format_event_hook is None:
+                        message_bundle = format_reflected_event(event)
+                    else:
+                        message_bundle = await maybe_future(
```
agoose77: Should this have some kind of error handling and graceful fallback?

kubespawner/messages.py
```diff
@@ -0,0 +1,242 @@
...
+    events = json.load(args.file)
+    formatters = [] if args.no_pretty else None
+
+    bundles = [format_reflected_event(event, formatters) for event in events]
+
+    if args.json:
+        json.dump(bundles, sys.stdout)
+    else:
+        for bundle in bundles:
+            print(bundle["message"])
```
agoose77: Useful for debugging and defining the regression tests

manics:
> This does effectively vendor some cluster-provider specifics (like the GCP scheduler). I think that's OK? But if we are vehemently against that, we can just pull those parts out.

Whatever we decide we need to be consistent in future. If we add GCP specific code we need to accept code for other clouds, including from third parties who use platforms that we can't test ourselves.

kubespawner/spawner.py
```diff
@@ -1924,6 +1925,25 @@ def _validate_image_pull_secrets(self, proposal):
...
+        Expects a callable that takes two parameters:
+
+           1. The spawner object that is doing the spawning
+           2. The Event object that is to be formatted
+
+        This can be a coroutine if necessary. When set to none, the default formatter is used.
+        The hook function should return a dictionary containing a required key `message`,
+        and an optional key `html_message`, pertaining to the plain-text and rich-representation,
+        of the formatted event.
+        """,
```
jnywong, line 1977:
```suggestion
        help="""
        Callable to build an event message from a reflected Event object.

        Expects a callable that takes two parameters:

           1. The spawner object that is doing the spawning
           2. The Event object to be formatted

        This can be a coroutine if necessary. When set to none, the default formatter is used.
        The hook function should return a dictionary containing a required key `message` for reformatting plain-text and an optional key `html_message` for the rich-representation.

        For example, add a basic format event hook::

           def format_hook(spawner, event):
              return {
                  "message": f"custom-message-{event['message']}",
                  "html_message": f"<span>{event['message']}</span>",
              }
           c.KubeSpawner.format_event_hook = format_hook

        or extend the default formatter with::

          from kubespawner.messages import format_reflected_event, DEFAULT_EVENT_FORMATTERS
          
          def my_formatter(event):
              # ... logic
              return "I'm a custom event formatter"
          
          def format_hook(spawner, event):
              return format_reflected_event(event, [my_formatter, *DEFAULT_EVENT_FORMATTERS])
              
          c.KubeSpawner.format_event_hook = format_hook
        """,
```

kubespawner/spawner.py
jnywong: Yes I think there's no harm in swapping these two conditions around to use the default formatter as the fallback. I think if something goes wrong here then I wouldn't want that to disrupt this loop.

For example, I set `format_event_hook` to something ridiculous like `return 10/0` and this did not yield any `message_bundle`s at all.

tests/test_spawner.py
```diff
@@ -657,7 +657,55 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
             json.dump(progress, devnull)
-    assert 'Started container' in '\n'.join(messages)
+    assert 'Started the notebook container' in '\n'.join(messages)
+
+    await start_future
+    # stop the pod
+    await spawner.stop()
+
+
+async def test_spawn_progress_formatter_hook(
```
jnywong: As far as I can tell, this tests that hook works, great!

However, I think the bulk of this PR is actually introducing a default formatter in `messages.py` and I note that the tests here do not make use of the included `sample-events`.

Can you add a test that recreates the formatted messages from the raw events from the sample data?

kubespawner/messages.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # Match the container paths and parse the container name
+    match = re.match(CONTAINER_FIELD_PATH_PAT, field_path)
+    if not match:
+        return
+
+    container_field, container = match.groups()
+
+    # Switch on the reason for the event
+    if event["reason"] in ("Started", "Killing", "Created", "Stopped"):
```
jnywong: I'm looking at the [list of container event reasons](https://github.com/kubernetes/kubernetes/blob/3799fc9942a4df017e2c4f30524816fcf89045e4/pkg/kubelet/events/event.go#L19) and see we have partial coverage here – that's okay! Just noting for my own understanding that from line 241 we fallback to the original `event["message"]` if we don't match any conditions.

kubespawner/messages.py
```diff
@@ -0,0 +1,275 @@
...
+        return
+
+    if event["reason"] != "NodeAffinity":
+        return
+
+    predicate_match = re.match(NODE_AFFINITY_FAILED_PAT, event["message"])
+    if predicate_match is None:
+        return
+
+    return "It was not possible to find or launch any nodes to run your server. This is likely due to a configuration problem with the infrastructure or the JuyterHub"
```
jnywong:
```suggestion
    return "It was not possible to find or launch any nodes to run your server. This is likely due to a configuration problem with the infrastructure or the JupyterHub"
```

kubespawner/messages.py
```diff
@@ -0,0 +1,275 @@
...
+        return
+
+    if event["reason"] != "TaintManagerEviction":
+        return
+
+    predicate_match = re.match(CANCELLING_DELETION_MESSAGE_PAT, event["message"])
+    if predicate_match is None:
+        return
+
+    return "Cancelling deletion of your server. This normally happens when a scale-up has just taken place."
```
jnywong: Small grammar suggestion: let's punctuate with full-stops consistently across all message returned.

jnywong (commented):
I enjoy this feature ❤️ I like that the `format_event_hook` was easy to configure for basic formatting. It took me a while to understand what was going on with the default formatting, but I think there will alway be room for improvements there. 

In general, my main comment is to complete the test suite to regenerate how the `sample-events` were created through the `message.py`, since I think that represents the bulk of the work in this PR. The default formatting may undergo further development, but like you say I think we want to add in some basic regression testing and update that if needed in future.

---

In answer to your questions:

>Is Kubespawner making too many assumptions if we bake-in the expectation of Bootstrap?

We know Bootstrap ships with JupyterHub, so I think this is a safe assumption for now. I don't *think* we need to worry about this for BinderHub?

> Could we consider adding a start-time timestamp so that times can simply be given as "minutes since spawn" rather than UTC times?

I think this is a nice-to-have. Most users hopefully shouldn't have to dwell on the spawn progress screen, but if they do, then they will likely screenshot their spawn failure to send to an admin. Keeping the timestamps consistent with server side logs with raw k8s events should be useful for sysadmins for troubleshooting.

> Should I rework the built-in formatter to generate HTML at every stage — would it be useful to have e.g. image names/tags, and node names in button-tags?

At this stage, I would prefer not to have the default formatter be too flashy.

> Is there motivation to move the default message format into its own configurable, rather than requiring users to create their own hook?

Yes, out of all of these questions I think this would be one to focus on. Most people will probably want to configure an extension to the default formatter.

jnywong:
> > This does effectively vendor some cluster-provider specifics (like the GCP scheduler). I think that's OK? But if we are vehemently against that, we can just pull those parts out.
> 
> Whatever we decide we need to be consistent in future. If we add GCP specific code we need to accept code for other clouds, including from third parties who use platforms that we can't test ourselves.

I am okay with that – I don't think we need to assume full responsibility for testing code on platforms we don't have access to, but we should ensure contributors who would like this functionality to include full test suites for that. I think the scope of changes in this PR are pretty cosmetic, so there doesn't seem to be huge scope for someone to introduce anything too crazy on a third party platform.

There is a small question about how to structure this as the corpus of messages to reformat scales, but I think we can cross that bridge when we get to it?

kubespawner/messages.py
```diff
@@ -0,0 +1,275 @@
...
+def format_reflected_event(event: dict, formatters: list = None) -> str:
+    """
+    Format a Kubernetes Event object into a human-readable message.
+
+    :ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#event-v1-core
+    """
+    if formatters is None:
+        formatters = DEFAULT_EVENT_FORMATTERS
+
+    # Format each message into a bundle
```
yuvipanda: I would like to make this a little more declarative. We can learn from kubernetes here!

For each event, we have to determine:

1. Do we have a matching formatter?
2. If we do, then pass the event to it, and we are done
3. If we don't, then just pass in the default

How do we determine `matching`? The current implementation basically doesn't do any matching, and passes all events to all formatters. And it becomes the responsibility of each formatter to implement checking logic, and determine if it wants to do something. This is repetitive and makes adding new formatters harder.

Instead, can we think about what kinda config sturcture would help us write more helpful matchers? For example, you could rewrite the `container_events_formatter` in a declarative way like this:

```
event_formatters = [
    {
        "match": {
            "reportingComponent": "kubelet",
            "fieldPath": r"^spec\.(initContainers|containers)\{(?P<container>[^}]+)\}$",
            "reason": ["Started", "Killing", "Created", "Stopped"]
        },
        "formatter": lambda event, matches: f"{event['reason']} the {matches['fieldPath']['container']} container"
    },
    {
        "match": {
            "reportingComponent": "kubelet",
            "fieldPath": r"^spec\.(initContainers|containers)\{(?P<container>[^}]+)\}$",
            "reason": ["Pulling", "Pulled"],
            "message": r'^.*image "(?P<image>[^"]+)".*$'
        },
        "formatter": lambda event, matches: f"{event['reason']} image {matches['message']['image']} for the {matches['fieldPath']['container']} container"
    }
]
```

We normalize `reportingComponent` to contain the reporting component. We write a simple matcher that supports regular strings (as regexes), and lists of regexes, and a formatter is a function (could eventually be a [template string](https://docs.python.org/3/library/string.templatelib.html) also). The `event` param is the event, and the `matches` is a dict of regex matches for each item in `match`. We can rely on named capture groups to make our formatters easier. 

This is much easier to extend, reason about and write. It also allows admins to add more by offering an `extra_event_formatters` or something like that - so they can *add* to this without having to replace what we have. And we can easily add more for multiple other sources as we run into them over time. Makes testing much easier too - as we can simply test our matchers with custom events, and then go on with our lives.

This is a bit of a fundamental change here, hope that's ok!

yuvipanda (dismissed):
Thank you for working on this, @agoose77! I left a comment about changing the implementation to be a lot more declarative than it is now, which should hopefully make both maintenance and extension much easier.

tests/test_spawner.py
agoose77: The tests do this already! The test_messages.py performs this check.

kubespawner/messages.py
agoose77: Yes, this is only partial as I wanted to back it with things I could easily generate. We could be more exhaustive if there is scope.

kubespawner/messages.py
agoose77: My kind of pedantry!

kubespawner/messages.py
agoose77: @yuvipanda I LOVE this API suggestion. It naturally avoids the need to e.g. pass in custom formatters via a hook.

What's interesting is that looking over my PR with fresh eyes, this kind of design feels fairly self-evident. When I started the work over in the JS side, it felt a lot like a parse-IR-render loop.

I am inclined to support both template strings and callables -- the former feels like the sane default (and in 3.14, t-strings can be accepted too!)

agoose77:
@yuvipanda I've reworked the PR to leave the functional hook that completely bypasses event formatting and remove the default formatter callable.

I think the test failure is just a flaky test?

I've then added a rule system with no defaults. This was a two-fold decision:
1. It keeps kubespawner leaner (no cloud specific functionality)
2. It resolve the problem of making this overrideable.[^why]

As such, I intend to put the specific implementation of these rules in the z2hj chart instead. I'm happy to revert that decision, but I am not off-the-top-of-my-head aware of a nice way to do it other than defining the default value as a dict and implementing the `extra` functionality in kubespawner itself, where users could clobber the default names and set them to None (plus None handling logic to remove these).

[^why]: My knowledge of traitlets is that one can't refer to the default value of a trait. This means that to allow users to override this, e.g. z2jh would literally have to import the HasTraits-derived parent class and extract the default to compose them.

tests/test_spawner.py
```diff
@@ -657,7 +657,160 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
+    corpus = "\n".join(messages)
+    # K8s changed the format of this message: https://github.com/kubernetes/kubernetes/pull/134043
+    assert "Started container" in corpus or "Container started" in corpus
+
+    await start_future
+    # stop the pod
+    await spawner.stop()
+
+
+@pytest.mark.parametrize("rules_as_dict", [True, False])
```
agoose77: TODO: test validation logic

yuvipanda:
I haven't looked at the implementation yet, but thank you for reworking it! I think the rules system should be here, not in z2jh. You can make it extensible by having a default list in here, and then just allowing `extra_event_formatter_rules` or similar that are appended. This is essential functionality that will benefit everyone using kubespawner, and I'm not concerned about a few extra rules that are cloud provider specific, especially as they also mostly come from the open source [autoscaler](https://github.com/kubernetes/autoscaler/) project. We aren't doing anything specific to any cloud provider here, but supporting things from the autoscaler which has plugins for cloud specific functionality.

agoose77:
OK, I didn't originally think it would be easy to add such a hook _here_ because of the challenges posed to the z2jh chart. However, with a fresh pair of eyes I no longer think that's the case.

I replaced the parse-instead-of-validate validator because trailets doesn't invoke the validator on the default. Now that we're setting a default, that's a problem, and it has other trade-offs (mutating the trait value).

I reworked the escape-hatch hook to be more of a "modify or replace" instead of "replace" style hook, which should improve UX.

kubespawner/spawner.py
```diff
@@ -1924,6 +1926,297 @@ def _validate_image_pull_secrets(self, proposal):
...
+            "reportingComponent": event.get("reportingComponent")
+            or event.get("source", {}).get("component")
+            or "",
+            "message": event.get("message") or "",
+            "reason": event.get("reason") or "",
+        }
+
+        # Populate cache of compiled event rules
+        if self._compiled_event_formatter_rules is None:
+            self._compiled_event_formatter_rules = self._compile_event_formatter_rules()
```
agoose77: Is there a better pattern for this?

I have a lot of experience with traitlets, but I may have missed an alternative.

I effectively want a computed property that is invalidated by the trait changes, _and_ is computed on the default value as well as subsequent values.

kubespawner/messages.py
```diff
@@ -0,0 +1,80 @@
+import datetime
+
+
+def parse_micro_timestamp(time: str) -> datetime.datetime:
```
agoose77: This module is now pretty small. I'm inclined to pull more out of kubespawner into here, but that does also harm extensibility. Any guidance on the kubespawner project's preference?

yuvipanda: I think it's useful to pull out the entire matching mechanism out of spawner into here, both for easier testing and reasoning.

agoose77:
@yuvipanda rather than moving just the business logic of event formatting to the new module created in this PR, I opted to move it into a new module under a new Configurable. This helps better isolate the responsibility, makes testing easier, and also improves extensibility.

tests/test_spawner.py
```diff
@@ -648,16 +652,65 @@ async def test_spawn_progress(kube_ns, kube_client, config, hub_pod, hub):
...
-        assert 'progress' in progress
-        assert isinstance(progress['progress'], int)
-        assert 'message' in progress
-        assert isinstance(progress['message'], str)
-        messages.append(progress['message'])
+        assert "progress" in progress
+        assert isinstance(progress["progress"], int)
+        assert "message" in progress
+        assert isinstance(progress["message"], str)
+        messages.append(progress["message"])
```
agoose77, line 658: My LSP changes these. It passes pre-commit, so I assume OK to leave?

kubespawner/events.py
```diff
@@ -0,0 +1,411 @@
...
+        We normalise missing or "" to ""
+        """
+        return {
+            "fieldPath": event["involvedObject"].get("fieldPath") or "",
+            "reportingComponent": event.get("reportingComponent")
+            or event.get("source", {}).get("component")
+            or "",
+            "message": event.get("message") or "",
+            "reason": event.get("reason") or "",
+            "type": event["type"],
```
agoose77, line 320: Added the `type` field, as it's harmless and useful!

kubespawner/events.py
```diff
@@ -0,0 +1,413 @@
...
+    rules = Union(
+        trait_types=[
+            List(),
+            Dict(),
+        ],
+        config=True,
+        help="""
+        List or dictionary of event formatter rules.
+
+        A rule consists of two required fields:
```
jnywong:
```suggestion
        A "rule" is an object that consists of two required fields:
```

Took me a beat on first read without looking ahead to realise when you wrote

> If provided as a list, each item should be an aforementioned "rule" object.

that you meant something like `rule = {"match": object, "template": string}`.

kubespawner/events.py
```diff
@@ -0,0 +1,413 @@
+import datetime
```
jnywong, line 5:
```suggestion
"""
Class for formatting Kubernetes event messages.
"""
import datetime
```

kubespawner/events.py
```diff
@@ -0,0 +1,413 @@
...
+            # of where rule came from
+            trait_path_template = "{}[{!r}]"
+
+            # Merge (in order) the given rulesets
+            merged_rules = {}
+            for i, rules_name in enumerate(("rules", "extra_rules")):
+                rules = getattr(self, rules_name)
+
+                if isinstance(rules, list):
+                    # NOTE: the rules index i must not exceed 9, as it is
```
jnywong, line 280: Is there an upper limit on the number of rules we can have because of this?

kubespawner/events.py
```diff
@@ -0,0 +1,413 @@
...
+            Dict(),
+        ],
+        config=True,
+        help="""
+        List or dictionary of event formatter rules.
+
+        A rule consists of two required fields:
+
+        - `match` — an object containing regular expression patterns (strings or compiled regular expressions) that match similarly named `Event` fields. Any named capture groups will be made available to the `template`.
+        - `template` — a string whose `.format` method will be invoked with any named capture group results. Missing named capture groups are provided as empty strings.
```
jnywong:
```suggestion
        - `template` — a string whose `.format` method is used to substitute placeholders `{}` with named capture group results. Missing named capture groups are provided as empty strings.
```

kubespawner/events.py
```diff
@@ -0,0 +1,413 @@
...
+                },
+                "template": "No existing nodes are currently able to run your server",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"taint-eviction-controller",
+                    "reason": r"TaintManagerEviction",
+                    "message": r"Cancelling deletion of Pod.*",
+                },
+                "template": "Cancelling deletion of your server. This normally happens when a scale-up has just taken place.",
```
jnywong:
```suggestion
                "template": "Canceling deletion of your server. This normally happens when a scale-up has just taken place",
```

I removed the full stop(period) for consistency with the other templates 🖊️ 

Additionally, as it pains me to say, we should stick to American English rather than British 🫖 🇬🇧

kubespawner/events.py
```diff
@@ -0,0 +1,413 @@
...
+                    "rule['match'] missing required key 'reportingComponent'"
+                )
+
+            # Check types of fields
+            allowed_match_fields = (
+                "reportingComponent",
+                "fieldPath",
+                "reason",
+                "message",
+                "type",
```
jnywong, line 219: Probably worth documenting that these are the only allowed kubernetes reporting components that we allow matches against.

tests/test_events.py
jnywong, line 1: Can we get some negative tests in here too?

For example, an exception is raised in https://github.com/jupyterhub/kubespawner/pull/910/changes#diff-77b886bba948cea4b7ea785aca33d3a1efa7a79a745cadf80b0d342d1640842eR330 that we can unit test.

jnywong (commented):
Nice, a significant improvement over the last time I reviewed this. It makes sense to make all of the matching rulesets configurable rather than essentially hard coding them into the library itself as before. Good job!

I have minor style suggestions and comments, but I think this is gtg once addressed.

kubespawner/events.py
agoose77: Thanks!

kubespawner/events.py
agoose77: Actually this is half-right. The existing text talks about a callable under the guise of a string template, whereas both are supported. I did a lot of reworking of this, so it must have dropped off my radar. Let me rework it.

kubespawner/events.py
agoose77: TIL cancelling is Brit!

kubespawner/events.py
agoose77: Nice catch.

kubespawner/events.py
agoose77: No! This is just the rule index into the sequence `(rules, extra_rules)`. This is an implementation detail that I wanted to surface because I really don't see anyone else adding 8 additional rule sources, but it's worth making it clear that I've assumed so!

jnywong:
I tested the scenario where I triggered a `Error: ImagePullBackOff` by entering an invalid image name into the kubespawner form and I think the default event rules introduced by this PR makes this more inscrutable than before

| Before | After |
| -- | -- |
| Screenshot 2026-04-13 at 14 53 41  |  Screenshot 2026-04-13 at 14 55 31 |

Could you take a close look at this? I am not sure why [this rule](https://github.com/jupyterhub/kubespawner/pull/910/changes#diff-77b886bba948cea4b7ea785aca33d3a1efa7a79a745cadf80b0d342d1640842eR159) would match the `Error: ErrImagePull` event.

agoose77:
@jnywong nice catch!

This PR adds a default formatter that was not being returned properly. I've lifted it earlier into the transforms to avoid needing to manually handle the exception case.

I've added rules for these errors:
image

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+                "template": "Waiting to try pulling images the {container} container",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"Failed",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r"Error: ErrImagePull",
+                },
+                "template": "Attempts to pull images for the {container} container failed",
```
jnywong:
```suggestion
                "template": "Attempts to pull an image for the container {container} failed",
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+                "template": "Attempts to pull {image} ({tag}) for the {container} container failed. Does this image exist?",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"Failed",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r"Error: ImagePullBackOff",
+                },
+                "template": "Waiting to try pulling images the {container} container",
```
jnywong:
```suggestion
                "template": "Waiting to try pulling an image into the container {container}",
```

A pod may have multiple containers, but each container is created from only one image.

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+    def _default_rules(self):
+        return [
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "reason": r"(?P<action>Pulling|Pulled)",
+                    "message": r'.*image\s*"(?P<image>[^"]+)\:(?P<tag>[^"]+)"',
+                },
+                "template": "{action} {image} image ({tag}) for the {container} container",
```
jnywong:
```suggestion
                "template": "{action} image {image}:{tag} for the {container} container",
```

I suggest keeping the image name and tag presented in the same format that it was entered into the form input.

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "reason": r"(?P<action>Started|Killing|Created|Stopped)",
+                },
+                "template": "{action} the {container} container",
```
jnywong, line 104:
```suggestion
                "template": "{action} the container {container}",
```

Reordering to _verb - noun - identifier_ reads better to me.

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+                "template": "Canceling deletion of your server. This normally happens when a scale-up has just taken place",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"BackOff",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r'Back-off pulling image "(?P<image>[^"]+)\:(?P<tag>[^"]+)"',
+                },
+                "template": "Waiting to try pulling {image} ({tag}) for the {container} container after the last attempt failed",
```
jnywong:
```suggestion
                "template": "Waiting to try pulling {image}:{tag} for the container {container} after the last attempt failed",
```

kubespawner/spawner.py
```diff
@@ -1924,6 +1934,48 @@ def _validate_image_pull_secrets(self, proposal):
...
+
+           1. The spawner object that is doing the spawning
+           2. The Event object that is to be formatted
+           3. The rendered event message string
+
+        This can be a coroutine if necessary. When set to none, the default formatter is used.
+        The hook function should return a dictionary containing a required key `message`,
+        and an optional key `html_message`, pertaining to the plain-text and rich-representation,
+        of the formatted event.
+        """,
```
jnywong, line 1977:
```suggestion
    decorate_progress_message = Callable(
        _decorate_progress_message,
        allow_none=True,
        config=True,
        help="""
        Callable to decorate a rendered event message from a reflected Event object.

        Expects a callable that returns a dictionary containing a required key `message`,
        and an optional key `html_message`, each with callable values that render plain-text and 
        rich-representation of the formatted event, respectively. The callable takes three parameters:

           1. The spawner object that is doing the spawning
           2. The event object to be formatted
           3. The rendered event message string

        This can be a coroutine if necessary. When set to none, the default formatter is used.
        
        Example
        -------
        .. code-block:: python

           def my_plain_decorator(message, event):
               event_type = event["type"]
               if event_type == "Normal":
                   return f"INFO: {message}"
               elif event_type == "Warning":
                   return f"WARNING: {message}"
               else:
                   return f"{message}"
   
           def my_decorator(spawner, event, message):
               return {"message": my_plain_decorator(message, event)}
           
           c.KubeSpawner.decorate_progress_message = my_decorator
        """,
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"Failed",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r'Failed to pull image "(?P<image>[^"]+)\:(?P<tag>[^"]+)"',
+                },
+                "template": "Attempts to pull {image} ({tag}) for the {container} container failed. Does this image exist?",
```
jnywong:
```suggestion
                "template": "Attempts to pull {image}:{tag} for the container {container} failed
```

I think I would avoid being too suggestive here by default, there could be other reasons for this failure.

jnywong (commented):
I think this looks good. There are a few more styling suggestions that i have made. @manics what do you think?

kubespawner/events.py
agoose77: Nice catch, I was thinking about pod-level logs.

kubespawner/events.py
agoose77: Yep, see above!

kubespawner/events.py
agoose77: If so, let's change the other messages too.

kubespawner/events.py
agoose77: I am less sure:
"Killing the notebook container"
vs
"Killing the container notebook"

kubespawner/spawner.py
agoose77: I hadn't realised we're OK to put inline examples — great idea!

kubespawner/events.py
agoose77: I'm leaning towards keeping it — anecdotally most of the times I see failed image pulls it is because the image doesn't actually exist rather than auth/network errors. Won't strongly push that.

jnywong:
@agoose77 the checks are probably failing because of the suggested tweaks i made to the default ruleset.

kubespawner/events.py
jnywong: How about

Killing the container "notebook"

?

I think my suggested ordering applies more generally than this and we can enclose identifiers/named matches in double-quotes/single-quotes/brackets/em-dashes/whatever you like.

agoose77:
> @agoose77 the checks are probably failing because of the suggested tweaks i made to the default ruleset.

Yep, I still need to update the regression tests.

kubespawner/events.py
agoose77: Verb - noun - identifier feels backwards in my mind, on the basis that when we talk about containers, we might say "HAL, please restart the container. Certainly. Which one? The NGINX container, please". I suspect this follows from the way that container identifiers are used both as unique IDs and as categorical descriptors, where the latter is more like an adjective than a noun.

In any case, given that I can see arguments for both, I'll propose that we accept your preference on the basis so that we can wrap this up! :shipit:

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+                "template": "Canceling deletion of your server. This normally happens when a scale-up has just taken place",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"BackOff",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r'Back-off pulling image "(?P<image>[^"]+)\:(?P<tag>[^"]+)"',
+                },
+                "template": "Waiting to try pulling {image}:{tag} for the {container} container after the last attempt failed",
```
agoose77:
```suggestion
                "template": 'Waiting to try pulling {image}:{tag} for the container "{container}" after the last attempt failed',
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+                },
+                "template": "{action} image {image}:{tag} for the {container} container",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "reason": r"(?P<action>Started|Killing|Created|Stopped)",
+                },
+                "template": "{action} the {container} container",
```
agoose77:
```suggestion
                "template": '{action} the container "{container}"'
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+                "template": "Waiting to try pulling {image}:{tag} for the {container} container after the last attempt failed",
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"Failed",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r'Failed to pull image "(?P<image>[^"]+)\:(?P<tag>[^"]+)"',
+                },
+                "template": "Attempts to pull {image}:{tag} for the {container} container failed",
```
agoose77:
```suggestion
                "template": 'Attempts to pull {image}:{tag} for the container "{container}" failed',
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"Failed",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r"Error: ImagePullBackOff",
+                },
+                "template": "Waiting to try pulling an image into the container {container}",
```
agoose77:
```suggestion
                "template": 'Waiting to try pulling an image into the container "{container}"',
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+            },
+            {
+                "match": {
+                    "reportingComponent": r"kubelet",
+                    "reason": r"Failed",
+                    "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
+                    "message": r"Error: ErrImagePull",
+                },
+                "template": "Attempts to pull an image for the container {container} failed",
```
agoose77:
```suggestion
                "template": 'Attempts to pull an image for the container "{container}" failed',
```

agoose77:
Let's hold this until I can clean up some style-only changes that have accidentally been introduced.

kubespawner/spawner.py
jnywong, line 45:
```suggestion
from traitlets import (
    Bool,
    Dict,
    Enum,
    Instance,
    Integer,
    List,
    Unicode,
    Union,
    default,
    observe,
    validate,
)
from traitlets import Type as TypeTrait
```

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+
+        - `match` — an object containing regular expression patterns (strings or compiled regular expressions) that match similarly named `Event` fields. Any named capture groups will be made available to the `template`. Supported fields are `reportingComponent`, `fieldPath`, `reason`, `message`, and `type`.
+        - `template` — a string, or a callable that will be used to build a formatted event message. If a string, the `.format` method will be invoked with any named capture group results as keyword arguments. If a callable, the same named capture groups will be directly passed as keyword arguments. Missing named capture groups are provided as empty strings.
+
+        If provided as a list, each item should be an aforementioned "rule" object.
+        If provided as a dictionary, the keys can be any descriptive name and the values should be the aforementioned "rule" objects.
+        The items will be sorted lexicographically by the dictionary keys, and the sorted values will be used to build the list of rules.
+
+        :ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#event-v1-core
```
jnywong, line 95:
```suggestion
        .. admonition:: Example
           :collapsible: closed

           Here is an example of two rules that are defined by default in the `RuleEventFormatter`.

           .. code-block:: python

              c.RuleEventFormatter.rules = 
                [
                    {
                        "match": {
                            "reportingComponent": r"kubelet",
                            "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
                            "reason": r"(?P<action>Pulling|Pulled)",
                            "message": r'.*image\s*"(?P<image>[^"]+)\:(?P<tag>[^"]+)"',
                        },
                        "template": "{action} image {image}:{tag} for the {container} container",
                    },
                    {
                        "match": {
                            "reportingComponent": r"kubelet",
                            "fieldPath": r"spec\.(initContainers|containers)\{(?P<container>[^}]+)\}",
                            "reason": r"(?P<action>Started|Killing|Created|Stopped)",
                        },
                        "template": '{action} the container "{container}"',
                    }
                ]
                
        When set to none, a default ruleset is used.

        :ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#event-v1-core
```

After not looking at this for a while, I think having an example ruleset here in the docs helps with quickly getting my head around how i should format them.

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+        icon = " ℹ️ "
+    else:
+        icon = " "
+
+    if event["lastTimestamp"]:
+        moment = parse_timestamp(event["lastTimestamp"])
+    else:
+        moment = parse_micro_timestamp(event["eventTime"])
+
+    # Trim the time to the nearest section, assume UTC
```
jnywong:
```suggestion
    # Trim the time to the nearest second, assume UTC
```

I think this is a typo

kubespawner/events.py
```diff
@@ -0,0 +1,455 @@
...
+    event_type = event["type"]
+
+    if event_type == "Warning":
+        icon = ' <span class="badge bg-warning-subtle text-warning-emphasis rounded-pill">Warning</span> '
+    elif event_type == "Normal":
+        icon = ' <span class="badge bg-info-subtle text-info-emphasis rounded-pill">Info</span> '
+    else:
+        icon = " "
+
+    # Trim the time to the nearest section, assume UTC
```
jnywong:
```suggestion
    # Trim the time to the nearest second, assume UTC
```

kubespawner/spawner.py
```diff
@@ -1972,6 +1982,65 @@ def _validate_image_pull_secrets(self, proposal):
...
+                   return f"INFO: {message}"
+               elif event_type == "Warning":
+                   return f"WARNING: {message}"
+               else:
+                   return f"{message}"
+   
+           def my_decorator(spawner, event, message):
+               return {"message": my_plain_decorator(message, event)}
+           
+           c.KubeSpawner.decorate_progress_message = my_decorator
```
jnywong, line 2030:
```suggestion
        .. code-block:: python

           def my_plain_decorator(message, event):
               event_type = event["type"]
               if event_type == "Normal":
                   return f"INFO: {message}"
               elif event_type == "Warning":
                   return f"WARNING: {message}"
               else:
                   return f"{message}"

            def my_html_decorator(message, event):
                return f"{event['lastTimestamp']}<span class='badge bg-info-subtle text-info-emphasis rounded-pill'>Info</span>{message}"

            def my_decorator(spawner, event, message):
                return {
                    "message": my_plain_decorator(message, event),
                    "html_message": my_html_decorator(message, event)
                }
           
           c.KubeSpawner.decorate_progress_message = my_decorator
```

Added an example of how to add a html decorator.

kubespawner/spawner.py
```diff
@@ -1972,6 +1982,65 @@ def _validate_image_pull_secrets(self, proposal):
...
+
+        Expects a callable that returns a dictionary containing a required key `message`,
+        and an optional key `html_message`, each with callable values that render plain-text and 
+        rich-representation of the formatted event, respectively. The callable takes three parameters:
+
+           1. The spawner object that is doing the spawning
+           2. The event object to be formatted
+           3. The rendered event message string
+
+        This can be a coroutine if necessary. When set to none, the default formatter is used.
```
jnywong:
```suggestion
        This can be a coroutine if necessary. When set to none, the default decorator is used.
```

kubespawner/events.py
jnywong:
```suggestion
            "type": event.get("type") or "",
```

For consistency, in case the key doesn't exist

jnywong (commented):
Leaving this review here before I go away next week.

One thing I wanted to check but didn't have time is test coverage of events emitted by the Scheduler, as [mentioned in this comment](https://github.com/2i2c-org/infrastructure/issues/7160#issue-3658227854)

I think we can make an rc and deploy this on a community hub, e.g. Earthscope, to see how this holds up in production next. What do you think?

kubespawner/events.py
agoose77: Brains are weirddddd :fish:

kubespawner/spawner.py
agoose77: Ooh my pet peeve! Thanks for catching this one.

rtmiz:
Very cool feature!

I've worked on a spawn message related PR previously and we do use this in most of our deployments so I just wanted to mention it here as it might be relevant, #900 supports adding custom messages that do not correspond to Kubernetes events, which is quite neat for informing users about things happening in hooks that are being ran.

minrk (approved):
Thanks for all the work on this! I've had a read through and it looks sensible to me, and I think this will really improve what users see on the progress page.

Thanks folks for reviewing!

PR #918, opened 2026-05-16
pre-select saved user options on the spawn form

### LINKED ISSUE #917: Named servers should remember saved user options
Currently if we save a named server, the server doesn't remember the specified user options (like the profile and image used etc). Jupyterhub stores this information in user_options, so the data is available if Kubespawner makes use of it.

### Proposed change
We should pass the `user_options` data into the profile selection form. There we could preselect the profile, image etc based on `user_options` if available. This will make sure that when a user restarts a named server, the options they choose when creating the server originally are preselected for them.

### Who would use this feature?
Anyone who uses the named server feature. And this would also help us support this feature in [jupyterhub-fancy-profiles](https://github.com/2i2c-org/jupyterhub-fancy-profiles) and on [NASA VEDA Jupyterhub instances ](https://github.com/NASA-IMPACT/veda-jupyterhub/issues/64) by extension.

sunu:
Preselects user options like profile, image etc if available. This should make sure that the saved options are pre-selected when starting a previously saved named server.

Fixes https://github.com/jupyterhub/kubespawner/issues/917. And would be useful to fix the same downstream issue(https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/42) in jupyterhub-fancy-profiles

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3487,7 +3487,9 @@ class KubeSpawner(Spawner):
             profile_form_template = env.from_string(self.profile_form_template)
         else:
             profile_form_template = env.get_template("form.html")
-        return profile_form_template.render(profile_list=profile_list)
+        return profile_form_template.render(
+            profile_list=profile_list, user_options=self.user_options or {}
+        )
 
     async def _render_options_form_dynamically(self, current_spawner):
         """
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3511,6 +3511,20 @@ class KubeSpawner(Spawner):
         """
         profile_list = self._get_initialized_profile_list(profile_list)
 
+        # user_options may be stale (e.g. profile_list changed since the last spawn)
+        # or set unvalidated via the REST API, so validate them first and fall back
+        # to defaults if invalid.
+        user_options = self.user_options or {}
+        if user_options:
+            try:
+                self._validate_user_options(profile_list)
+            except ValueError as e:
+                self.log.warning(
+                    f"Not pre-selecting saved user_options on the spawn form "
+                    f"because they failed validation: {e}"
+                )
+                user_options = {}
+
         loader = ChoiceLoader(
             [
                 FileSystemLoader(self.additional_profile_form_template_paths),
@@ -3536,7 +3550,7 @@ class KubeSpawner(Spawner):
         else:
             profile_form_template = env.get_template("form.html")
         return profile_form_template.render(
-            profile_list=profile_list, user_options=self.user_options or {}
+            profile_list=profile_list, user_options=self.user_options
         )
 
     async def _render_options_form_dynamically(self, current_spawner):
--- a/kubespawner/templates/form.html
+++ b/kubespawner/templates/form.html
@@ -9,7 +9,9 @@
                name="profile"
                id="profile-item-{{ profile.slug }}"
                value="{{ profile.slug }}"
-               {% if profile.default %}checked{% endif %} />
+               {% if user_options.profile == profile.slug %}checked
+               {% elif not user_options.profile and profile.default %}checked
+               {% endif %} />
       </div>
       <div>
         <h3>{{ profile.display_name }}</h3>
@@ -53,6 +55,7 @@
   {%- endfor %}
 </div>
 <script>
+  const userOptions = {{ user_options|tojson }};
   $('.js-profile-option-select, .js-profile-option-label').click(function() {
     // we need this bit of JS to select the profile when a <select> inside is clicked.
     $(this).parents('.js-profile-label')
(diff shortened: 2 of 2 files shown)


sunu:
All tests pass locally. Not sure if the failing one is flaky test that needs a retry 🤔

jnywong:
The CI was fixed in https://github.com/jupyterhub/kubespawner/pull/923 so you can rebase this and try again

agoose77:
@sunu LGTM. Have you tested this yourself?

sunu:
@agoose77 yes, here's a before and after comparison without the changes in this PR and with the changes.

**Before:** Launching a named server doesn't load the previously used profile and image automatically.

https://github.com/user-attachments/assets/3e07a903-11aa-4742-98b7-911909d175cd

---

**After:** Launching a named server does load the previously used profile and image automatically.

https://github.com/user-attachments/assets/8b6bfcd4-218d-4d7b-aa29-8de4cd061019

kubespawner/spawner.py
```diff
@@ -3535,7 +3535,9 @@ def _render_options_form(self, profile_list):
             profile_form_template = env.from_string(self.profile_form_template)
         else:
             profile_form_template = env.get_template("form.html")
-        return profile_form_template.render(profile_list=profile_list)
+        return profile_form_template.render(
```
jnywong, line 3552:
```suggestion
        if self.user_options:
            self._validate_user_options(profile_list)
        return profile_form_template.render(
```

Given that the javascript heavily depends on the name format `profile-option-{slug}--{optionKey}` to be preserved, I suggest running the validation step here before this form is rendered since looking at the function description of that validator, it's possible to set `user_options` through the REST API without any input validation.

kubespawner/templates/form.html
```diff
@@ -89,6 +90,22 @@ <h3>{{ profile.display_name }}</h3>
...
   // on page load
   $(document).ready(() => {
+    // Pre-select options from previously saved user_options
+    const savedProfile = userOptions && userOptions.profile;
+    if (savedProfile) {
+      $(`label[for="profile-item-${savedProfile}"] .js-profile-option-select`).each(function() {
+        // name format: profile-option-{slug}--{optionKey}
+        const optionKey = $(this).attr('name').split('--').pop();
+        const unlistedValue = userOptions[optionKey + '--unlisted-choice'];
+        const savedValue = userOptions[optionKey];
```
jnywong, line 102: This could be brittle towards future code changes since there is string manipulation that depends on the name format `profile-option-{slug}--{optionKey}` being preserved. That being said, I am okay with this for now.

kubespawner/templates/form.html
```diff
@@ -89,6 +90,22 @@ <h3>{{ profile.display_name }}</h3>
   // wrapping in a document.ready to make clear this is executed once
   // on page load
   $(document).ready(() => {
+    // Pre-select options from previously saved user_options
+    const savedProfile = userOptions && userOptions.profile;
+    if (savedProfile) {
+      $(`label[for="profile-item-${savedProfile}"] .js-profile-option-select`).each(function() {
```
jnywong, line 98: I see an error in the browser console: `Incorrect use of <label for=FORM_ELEMENT>` 

I think the `select` and `input` tags need the `id` attribute defined: `<select id="profile-option-{{ profile.slug }}--{{ k }}"` so that screen-readers know which id is attached to which label. See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label

jnywong (commented):
I have a couple of minor comments that need addressing, but other than that I think this is G2G

kubespawner/templates/form.html
sunu: oh nice catch! Just for context, this seems like an existing issue and not something introduced by this PR. Happy to add a fix here though since it's quite a low hanging fruit.

kubespawner/templates/form.html
sunu: fair point. This is actually an existing pattern in the template. the same `profile-option-{slug}--{optionKey}` name format is already there, so this change isn't introducing this pattern so much as following what's already there. but agreed may be it's worth coming back to it later.

kubespawner/spawner.py
sunu: that's a great catch @jnywong, and you're right that these can arrive unvalidated. I dug into the Jupyterhub side to be sure and yes, options set via the REST API are passed straight through and persisted to the DB before any validation runs.

I'll add a validation step before rendering. But rather than let it raise an exception directly, I think it's better to let it it fall back to profile defaults when validation fails (with a warning log message). I think safely degrading to defaults is friendlier than crashing the spawn page on stale/invalid saved options.

sunu:
Thanks so much for the review @jnywong! I've pushed a few commits to address the feedback.

jnywong (approved):
Great, thanks for responding so swiftly!

PR #845, opened 2024-07-26
Allow more pod properties to be defined as dictionaries

sunu:
As a continuation of https://github.com/jupyterhub/kubespawner/pull/843, allow more pod properties to be defined as dictionaries for easy overriding.

Adds dictionary support for `init_containers`, `extra_containers`, `tolerations`, `node_affinity_preferred`, `node_affinity_required`, `pod_affinity_preferred`, `pod_affinity_required`, `pod_anti_affinity_preferred` and `pod_anti_affinity_required`

PR #919, opened 2026-05-25
Allow profile_list to be assigned a dictionary

GeorgianaElena:
This is a continuation of #843 and #845 and adds dictionary support for `profile_lists`.

Didn't want to make this a breaking change and update the name of the traitlet, though in its current form it assumes lists and might be confusing.

How do folks feel about renaming it?

Ref: https://github.com/2i2c-org/infrastructure/issues/6941

### DIFF
--- a/kubespawner/spawner.py
+++ b/kubespawner/spawner.py
@@ -3573,7 +3573,9 @@ class KubeSpawner(Spawner):
             return self._render_options_form_dynamically
         else:
             # Return the rendered string, as it does not change
-            return self._render_options_form(self._sorted_dict_values(self.profile_list))
+            return self._render_options_form(
+                self._sorted_dict_values(self.profile_list)
+            )
 
     @default('options_from_form')
     def _options_from_form_default(self):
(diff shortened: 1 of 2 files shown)


yuvipanda:
The design of this feels perfectly fine to me, and I don't think we need to rename it. The profiles are still *presented* as a list, so this is fine.

The test failures look real though.

GeorgianaElena:
Thanks @yuvipanda! I think I fixed the tests related to the PR. The others seem like versions incompatibilities somewhere?

yuvipanda:
Ty for fixing the tests, @GeorgianaElena! I'm fixing the upstream tests in https://github.com/jupyterhub/kubespawner/pull/923

yuvipanda:
Merging this here, since the tests are broken on main too. Thanks, @GeorgianaElena!

jnywong:
~~Yeah I think some tests are not passing because of changes in this PR didn't catch, e.g. https://github.com/jupyterhub/kubespawner/actions/runs/27967014719/job/82763344623?pr=910~~

No wait nvm this PR is unrelated to that.

PR #923, opened 2026-06-19
Fix Failing Tests on latest k3s on Main

yuvipanda:
- Fix failing test due to changed structure of event description on latest k8s
- Seperate out 'test on latest k3s' and 'test on latest jupyterhub main' so it's easier to  
   tell what fails. Remove an earlier redundant lower version python test combo so we don't
  explode our CI time
- Increase the timeout for deleting namespaces, as it seems to take >20s for all finalizers to
  settle on k8s / k3s 1.32+

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -59,27 +59,7 @@ jobs:
 
           # Test with latest python and newer kubernetes to see where it fails
           - python: "3.12"
-            k3s: v1.30
-
-          # Test with latest python and newer kubernetes to see where it fails
-          - python: "3.12"
-            k3s: v1.31
-
-          # Test with latest python and newer kubernetes to see where it fails
-          - python: "3.12"
             k3s: v1.32
-
-          # Test with latest python and newer kubernetes to see where it fails
-          - python: "3.12"
-            k3s: v1.33
-
-          # Test with latest python and newer kubernetes to see where it fails
-          - python: "3.12"
-            k3s: v1.34
-
-          # Test with latest python and newer kubernetes to see where it fails
-          - python: "3.12"
-            k3s: v1.35
 
           # Test with specific python, kubernetes and latest jupyterhub
           - python: "3.11"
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -50,14 +50,16 @@ jobs:
...
           # Test with modern python and k8s versions
-          - python: "3.11"
-            k3s: v1.27
           - python: "3.12"
             k3s: v1.28
 
-          # Test with latest python and JupyterHub in main branch
-          - python: "3.X"
+          # Test with latest python and latest kubernetes
+          - python: "3.12"
```
consideRatio: It would be good to retain testing on latest python in the test suite somewhere, maybe we are - quickly looked to approve from mobile

consideRatio (approved):

jnywong:
omg i was just looking at this ty @yuvipanda

yuvipanda:
The test failure here is kinda 'real' in that our tests have a problem with latest k3s and finalizers. I've added some debug statements (that we should remove before merging), but not quite able to get it fixed.

jnywong:
Ty for taking a closer look at this -- i picked up https://github.com/jupyterhub/kubespawner/pull/910 again and definitely want to be sure the CI is working as expected before progressing that PR

yuvipanda:
I just temporarily started testing all the versions, and notice that our tests start failing at version v1.32

image

Here's the 1.32 changelog: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.32.md

yuvipanda:
There are no pods, no pvs, no PVCs, yet the namespace is stuck in 'terminating' :( That's the cause of this problem

yuvipanda:
@jnywong I fixed it! You should be able to merge and rebase that PR on top of main to get more confidence in CI!

yuvipanda:
Turns out that after k8s 1.32, namespace deletion takes more than 20s

jnywong:
Awesome, thanks @yuvipanda !

jnywong (approved):

yuvipanda:
Merging to unbreak master :)

PR #941, opened 2026-09-19
Decrease frequency of 'please be patient' messages

yuvipanda:
Seeing them every 5s does not unfortunately help me be more patient and has 'are we there yet?' vibes

image


GeorgianaElena (approved):
I've been feeling the same 😅

minrk (approved):
Fair enough. How noisy it feels definitely depends on if you expand the log or just look at the last-line that gets displayed by default below the progress bar.

